PR #25, opened 2020-11-11
Add dynamic progress favicon

### LINKED ISSUE #24: Feature-Request: Animated progress bar Favicon
Let me start by saying that Fluidd is a great! Thank you so much for making this cool project!

One small thing that I'm missing compared to Mainsail is the animated Favicon in the browser tab that displays print progress.

Would be great to have such a feature in Fluidd too (not necessarily exactly this animated disk icon that Mainsail has. It could be something like flooding color of the (nice) Fluidd icon).

figgyc:
I did some work on the animated favicon feature since i saw a few requests for it.
I adapted the code from the [equivalent mainsail feature](https://github.com/meteyou/mainsail/blob/master/src/App.vue#L201), I did make some design changes though, not sure what you might think. Here's what it looks like:
image

I changed the start location of the arc to match the one in the main interface and I also added a logo which was a bit awkward to implement but it works.
Should fix #24.


src/mixins/utils.ts
```diff
@@ -23,6 +24,50 @@ export default class UtilsMixin extends Vue {
     }
   }
 
+  get pageIcon () {
```
cadriel, line 27: I think this might be better off in the Globals.ts - that way it's not mixed in to every page that may not require it.

cadriel (changes requested):
Other than that one comment, I think this looks good.

I have a large change coming in soon tho - and would prefer this come in after that - so will defer until then.

Otherwise, good work! :)

src/mixins/utils.ts
figgyc: You're probably right there, the only reason I put it in utils.ts is because that's where the title code is aswell. I'll move the icon code when I get some time/after you push that change. One one hand would it be a good idea for the title code to be in globals aswell? I suppose it is a lot simpler though so maybe it's fine where it is. Thanks for taking a look.

cadriel: Yah - honestly i need to pull a lot of stuff from the util mixin. :)

I had another thought too, can you update so that the color of the ring inherits the primary color theme? This way once I implement more themes, the favicon will replicate the ring color.

You can prolly grab this somewhere from `Vue.$vuetify..` or `this.$vuetify` depending where you're calling from.

cadriel:
Oh, also watch out for the force push to develop - you may need to reset your develop and pick your commits.

cadriel:
This branch is out of date, and needs fixing up + some requested changes. Do you want to go ahead and make these, otherwise I can put them on my list.

figgyc:
Hi, yeah you probably want to get the code moving fixes done on your end. I was struggling a bit because looking at globals it doesn't seem to me like the best place to put it to be honest since it should probably be in a Vue mixin to interact with headful, so maybe make a new mixin? Not too familiar with Vue myself though. Thanks for the help

cadriel:
Ok! I've got this ready for the next release in a new branch. I've made some adjustments, removed the icon - added the gray circle etc. I feel like the icon was a bit overwhelming when including the progress circle.

Closing this PR!

Thanks a bunch for this - and sorry it took so long! :)

PR #29, opened 2020-11-18
Parse macro names to be pretty

th3fallen:


cadriel:
I'm going to close this for now - as per discord discussion i think we need to better handle this so i'll look at implementing a bit later.

PR #62, opened 2021-01-01
feat: Added a dockerfile

shinuza:
Added a dockerfile that uses the `nginx` base image and downloads the latest release from github. I tried to build from source but I have non-reproductible issues in the container. Essentially, you could publish an image after a new release using this Dockerfile.
I also have `buildah` script if that's something you'd like to add.

A few things:

- We could run the `dist`command inside the container and remove the clutter with a multi stage build.
- We could also download the archive locally and then mount it inside the container to achieve the same result.
- I'll probably add a way to generate the config file at run time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM nginx
+
+RUN apt update -qq && apt install -y unzip
+WORKDIR /usr/share/nginx/html
+RUN curl -L https://github.com/cadriel/fluidd/releases/latest/download/fluidd.zip -o fluidd.zip
+RUN unzip -o fluidd.zip


cadriel:
This looks good. A couple of notes;

I think I'd prefer we keep the current impl of grabbing the release ZIP, rather than building inside the docker. The advantage I guess is that when people re-deploy the docker, they'd get the latest release. I think we need to figure out how updating Fluidd would best work with Docker. Happy to take recommendations here if this is an area of expertise for you!

Take a look at the `package.json` scripts and you'll see some further hints as to how releases work.

1. The fluiddpi release has specific nginx config, not all of which would be needed for a dockerfile in this case. But things like enabling gzip etc might be nice. Where is this config defined for this dockerfile?
2. When deploying Fluidd outside of the normal Pi host, there's a `config.json` file that can be configured to tell Fluidd to not try to self update. See the npm `host.deploy` script to see what's happening here. There's a preconfigured `config.json` file we could use. This might be required initially.

I'd like to see this as a thing people can deploy in their own network, that's easily update-able (by either redeploying, or self updating).

Thanks for your time!

shinuza:
So there are two ways at looking at things:

1. People who build the image locally
2. Those who are only going to download the image

The former implies, that they clone the repo or at least copy the Dockerfile locally to build it. The latter means they will only run `docker run -d -p 80:80 cadriel/fluidd`, which also means that we to configure github actions or other means to have the image available publicly ( in Github's registry or Docker's ).

Usually, people expect the solution 2. to work, and 1. to be there just for documentation/development purposes.

Another note is that the image tags should follow that of the releases, which means we need to tag, e.g for the current image `fluidd:v1.4.1` and `fluidd:latest`. We usually do not do self updates in the docker images.

I think I can try to experiment had have something deploying/tagging the image automatically in the github registry.

---

Regarding your questions:

1. The nginx configuration is the default one from the Nginx official docker image, but it's fairly easy either edit it in place or replace it with one of our choice.
2. There's a mean to pass configuration variable when starting the container, which means we can even specify a list of `endpoints` and `blacklist` at run time. Which might be needed if someone wants run multiple containers on the same cluster with different configurations for each.

dimalo:
I've just created a docker-compose setup and pushed to dockerhub...

https://github.com/dimalo/klipper-web-control-docker

As for updates in docker normally images get rebuild and then a newer is pulled. There are tools like watchtower to enable auto updates of containers...

I'll set up an alternate compose file...

I made this it's own repo as it deploys klipper, fluidd and mainsail so it's not meant for fluidd only.

Adding gzip is possible though not really necessary as the static dist is cached and loads really fast.

cadriel:
I think what has been suggested above all sounds good.

The blacklist / endpoints config has fluidd loading via JSON, but I can adjust to look at the `envuronment` or whatever we need that makes sense. Let me know if you need any changes here.

Github actions are running for fluiddpi, pretty sure you can use some of that if you're not already familiar with them. I have been meaning to look into this myself, but - features. :)

Thanks all!

dimalo:
> The blacklist / endpoints config has fluidd loading via JSON, but I can adjust to look at the `envuronment` or whatever we need that makes sense. Let me know if you need any changes here.

At least with the nginx image I figured out they have such a thing build in. I'm creating a template for upstreams.conf and fill it with ENV variables. [see the template](https://github.com/dimalo/klipper-web-control-docker/tree/main/frontend)
putting the template into /etc/nginx/templates makes the container render env variables at startup and move the output to /etc/nginx/conf.d

> Github actions are running for fluiddpi, pretty sure you can use some of that if you're not already familiar with them. I have been meaning to look into this myself, but - features. :)

[got you covered already](https://github.com/dimalo/klipper-web-control-docker/blob/main/.github/workflows/fluidd_multiarch.yml)... pretty easy to adapt

@shinuza sorry to jump in like that, please let me know what you think of my solution, I see it as same direction as your's but pushed further a bit ;)

cadriel:
Note that a lot of the proxy configuration isn't strictly necessary. Fluidd can connect directly to moonraker via port 7125.

The only proxy that's probably relevant is the mjpgstreamer path. So it may be worth considering simplifying the nginx config a bit for this case.

Now thinking about it - moonraker etc would always be external anyway - so its only really about hosting fluidd.. so yeah, should be good.

dimalo:
I guess it's nice and convenient atm because fluidd comes up and instantly is connected to moonraker without any further user action. At least I've tested with new private tab without any cache and it just connects.
But having more printers might be problematic. In the end you might be right but then I guess all proxies have to go, when I have two printers I want to have a cam for each right?

cadriel:
On startup, fluidd pings a bunch of endpoints to try to figure out known places to try to connect.

The host `config.json` file defines these endpoints - we also ping the browser URL. Once it finds one that works, it adds that as your first default printer - and won't ping again for future loads.

So - having it ping the 7125 ports on common hosts should be just as effective as with it pinging the proxy endpoints too.

But, if you're deploying on a host via docker - then likely your printers aren't going to have commonly known host names anyway - which will invoke the add printer dialog.

I think maybe exposing that `config.json` (or via the `environment`) and otherwise letting users define those endpoints might be nice for some.

The `config.json` has the following props so far;

```
{
  "endpoints": [...],
  "blacklist": [...],
  "skipClientUpdates": false
}
```

`endpoints` being the list of ping-able endpoints to try, `blacklist` being specifically a list of hosts NOT to ping (i.e., when hosted on fluidd.xyz for example we don't want it pinging itself) and `skipClientUpdates` prevents fluidd from allowing you to update the client - which we're going to need for this docker deploy.

dimalo:
I've seen in store/version that you try to update not only the client. if skipClientUpdates is false, am I right klipper and moonraker are also not able to update?
In the docker case, only the system and the client are not updateable, though showing an indication if one is available would be nice.
updating klipper and moonraker is possible as their container is shipped with git available for the klippy user (which has control to pull klipper/moonraker)
if the client and system update commands fail gracefully, even having updates ON by default works with the docker setup.

Having the "Enter Host" form is perfectly fine I think as it explicitly sets an endpoint and with that excludes it from any blacklist by user intent.

If I have 4 klipper instances running on a pi I'll need to map each on one of 4 different ports, with those and the pi's IP I can set them up.

If I'm a 200 printer power klipper-ninja I would probably just need to upload a config.json instead of entering each one by one.
So perhaps it's as easy as "Enter Host or Upload/Drop config.json"

Pretty smart that autodiscovery, means the upstream proxies really can be removed. As I have to browse to the IP of the klipper host running klipper & the nginx container, it just picks up the right IP to autodiscover.

Is the webcam URL configurable for each instance? If so, no need for any reverse proxies actually...

cadriel:
> I've seen in store/version that you try to update not only the client. if skipClientUpdates is false, am I right klipper and moonraker are also not able to update?

No. Moonraker and Klipper can update fine in this scenario. It's only the client that can't self update, because the API (moonraker) is not running on the same host as the client. You can also update the system fine too.

> In the docker case, only the system and the client are not updateable, though showing an indication if one is available would be nice.

See above.

> updating klipper and moonraker is possible as their container is shipped with git available for the klippy user (which has control to pull klipper/moonraker)
> if the client and system update commands fail gracefully, even having updates ON by default works with the docker setup.

So, the docker wouldn't be shipped with klipper or moonraker. The expectation is that the fluidd UI would connect to a moonraker instance (being a printer.. ) somewhere. But yeah, other updates should be good.

> Having the "Enter Host" form is perfectly fine I think as it explicitly sets an endpoint and with that excludes it from any blacklist by user intent.
> 
> If I have 4 klipper instances running on a pi I'll need to map each on one of 4 different ports, with those and the pi's IP I can set them up.
> 
> If I'm a 200 printer power klipper-ninja I would probably just need to upload a config.json instead of entering each one by one.
> So perhaps it's as easy as "Enter Host or Upload/Drop config.json"

Yeah - but, err - maybe not. Fluidd pings all of the endpoints asynchronously, and picks whatever one returns first. So it might not be valuable in that example.

> Pretty smart that autodiscovery, means the upstream proxies really can be removed. As I have to browse to the IP of the klipper host running klipper & the nginx container, it just picks up the right IP to autodiscover.

Its not truely an autodiscover, but from the users perspective its nice yeah.

> Is the webcam URL configurable for each instance? If so, no need for any reverse proxies actually...

It IS.. but there's code in place that assumes this proxy exists at the moment. I.e., when users connect via app.fluidd.xyz, then I try to guess the webcam url by way of the instance they're connecting to. So we'd still need that reverse proxy in place. Honestly, I think its a nice to have anyway - mjpgstreamer exposes itself on port 8080 I think by default, which is a bit ugly for most users.

cadriel:
The other PR was merged, so closing this one now.

Thanks all for the input!

PR #63, opened 2021-01-02
feat: Allow multiple files upload from upload button.

shinuza:
This is disabled for "upload & print" button.

### DIFF
--- a/src/components/cards/FileSystemCard.vue
+++ b/src/components/cards/FileSystemCard.vue
@@ -189,33 +189,42 @@ export default class FileSystemCard extends Mixins(UtilsMixin) {
       })
   }
 
-  async upload (file: File, root: string, path: string, andPrint: boolean) {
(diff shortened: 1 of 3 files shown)


cadriel:
Does this respect the same functionality that the multi-file drag and drop upload achieves? I.e., you see the progress - and each file is uploaded sequentially, not async?

I'll test myself too but otherwise if that all works then it looks good! Nice work.

shinuza:
Sure, I just came across that comment in FileSystemCard.vue, will do.

cadriel:
Just tested - it uploads async at the moment, and that's potentially a problem. Need to update to make this synchronous, the same as the drag upload.

shinuza:
Just updated it, it works as intended. But I'd love your input on architecture/coding style.

cadriel:
Great work! Thanks!

PR #71, opened 2021-01-05
add Dockerfile & Github workflow

dimalo:
- workflow untested (will fix as it runs...)
- add Dockerfile for deployment in nginx-alpine container
- build for x64, ARM32v7, ARM64
- use multistage build for small image size (13mb)
- configure gzip & jpegstreamer proxy
- allow passing jpegstreamer ip & port through env variables

You'll need to provide dockerhub user & password/token through github secrets for the workflow

### DIFF
--- a/.github/workflows/multiarch_dockerhub.yml
+++ b/.github/workflows/multiarch_dockerhub.yml
@@ -1,0 +1,66 @@
+name: Build and Push Multiarch Docker image for Dockerhub
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Prepare
+        id: prep
+        run: |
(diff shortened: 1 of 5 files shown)


dimalo:
confirming CI build is working on push to master. resulting image is {DOCKER_USERNAME}/fluidd:latest
I pushed a tag to my fork without any trigger so there might be work left.
But "latest" will be built nevertheless.

provided "dimalo" as DOCKER_USERNAME in repo secrets
provided access token as DOCKER_PASSWORD in repo secrets

removing ci test commit...

cadriel:
ok, this looks good. merged.

I did notice that the configuration is mostly under the docs folder, might want to move that elsewhere at some point.

cadriel:
I notice the action is doing weird things in github too - i should be able to force it to run, or at least view it right? Does it take github a while to pick it up?

cadriel:
nvm - master needed to be fast forwarded, showing up now :)

dimalo:
Didn’t check tags. Guess it needs to match version strings with prepended „v“
Yeah put it where it suits you. Don’t forget to match paths in the workflow

PR #72, opened 2021-01-05
fix: Filament reporting under 1m

### LINKED ISSUE #70: Filament below 1m is reported as millimeters
Hi, 

i noticed a unit mistake in the filament length. If a filament value is below 1m, the unit is wrongly displaxed as millimeters. 

The status in the screenshot shows a short printjob. Correct is: 0.44m / 0.77m.

Fluidd shows 0.44mm / 0.77mm which is wrong.

The same happens on the first 99cm of a printjob. Fluidd says 0.01mm up to 0.99mm, then 1.00m.

shinuza:
44 cm was reported as 0.44mm, it's now reported as 440mm
Fixes #70

### DIFF
--- a/src/plugins/filters.ts
+++ b/src/plugins/filters.ts
@@ -78,11 +78,12 @@ const Filters = {
   },
 
   /**
-   * Formats a number representing mm to m if > 1000
+   * Formats a number representing mm to human readable distance.
    */
   getReadableLengthString (lengthInMm: number) {
-    if (lengthInMm > 1000) return (lengthInMm / 1000).toFixed(2) + ' m'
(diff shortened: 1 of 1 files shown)


cadriel:
I think we need another denomination in here.. this change means that it'd start reporting things like 800mm, which isn't as useful.

Maybe have it report cm down to 10, then in mm?

shinuza:
> 
> 
> I think we need another denomination in here.. this change means that it'd start reporting things like 800mm, which isn't as useful.
> 
> Maybe have it report cm down to 10, then in mm?

Yep, as I mentioned on discord I had another version prepared with cm, wanted to confirm with you.

cadriel:
Looks good!

PR #89, opened 2021-01-10
chore: Unit tests - WIP

shinuza:
This is a work in progress

- [x] Filters
- [ ] Low level components
- [ ] ...

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,5 +25,8 @@ jobs:
       - name: Install deps
         run: npm install
 
+      - name: Run tests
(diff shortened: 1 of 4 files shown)


cadriel:
added build requirements - maybe rebase this, and then update the build.yml to include running the tests too.

shinuza:
Done

cadriel:
ok, i've merged what you had - and fixed the date check. This was caused because dayjs applies the browsers offset for you, so there was confusion there..

I'm not positive it's 100% correct yet, but i'll tweak a bit later if needed. :)

PR #90, opened 2021-01-11
Move z offset to the side

th3fallen:
Split offset to side


cadriel:
Closing as is already in develop

PR #101, opened 2021-01-17
Hide Z Offset when not printing

th3fallen:
Hide z offset panel when printer is idled


th3fallen:
Derp it appears this is intentional

PR #134, opened 2021-02-05
Add i18n, Chinese language support

QABQ:
completed:
1.Add i18n plugin
2.Add Chinese/English language pack
3.English language by default

to do list:
1.You need to add a language switch button at a suitable location, and realize language switch by controlling "localStorage lang"

img:
image
image
image
image
image


QABQ:
@cadriel  hello ?

Issue #99, opened 2021-01-17
Change dashboard title order


ChipCE:
Hi,
Sorry for my skeptic but I think we should change the dashboard title order. When printing the page title will be something like <code>[%12] | Dashboard | PrinterName</code> but due to the length limit of the tab in browser it can only display  <code>[%12] | Dashboard | Pri</code>. There will be no problem with just one printer, but for more than one it's hard to tell what tab is what printer dashboard by the tab name. 

![title](https://user-images.githubusercontent.com/22391265/104830680-463a6b00-58c4-11eb-87be-ebdd1304f054.png)

cadriel:
These are now swapped, look for it in the next release! :)

PR #141, opened 2021-02-08
Re-adjust page title order

ChipCE:
Follow up of my issues #99 about the page title order. Due to my unclear explanation the page title order has been changed to instanceName > pageName > progress. It should be progress > instanceName > pageName.

Signed-off-by: John Smith <login721@gmail.com>


cadriel:
What's the reasoning for moving the progress first?

After some thought i was felt that the icon presents a good visual representation of the progress, thereby reducing the importance of having the progress represented first.

ChipCE:
The icon is small, and its hard to tell the progress just by the icon, and the percent will be hided if you have some tabs open in the same browser windows. Human eyes (or at least me) cannot tell precise number from the icon only. For example this icon

cadriel:
Others have indicated that knowing WHAT printer is more important than knowing the percent, given the icon is there for a quick visual reference.

I.e., if you have more than one printer configured and in your tab list, its easier to see quickly how far through a print is - and for what printer.

I think for now, we might leave this as is. If more users chime in and suggest re-adjusting, then I'll re-look at this - but otherwise for now I'll close this PR.

Thanks for the input! :)

PR #149, opened 2021-02-14
feature i18n - Add  multi lang support

shinuza:
- [x] Working setup
- [x] Marked every strings
- [x] Locale override via configuration
- [x] French translation
- [x] Translate dialogs

For later: 

- Automatic string extraction
- Locale detection


cadriel:
Ok, I've taken a look - and have some comments;

1. It looks like you define the list of locales in the Globals file, however - the locales themselves come via external json files. I think that perhaps the locale list should prolly be defined in the config.json too because of this.  That way, users can test new locales without necessarily having to re-compile Fluidd.

2. I notice a lot of documentation (incl the vue i18n docs) suggest loading the locales via require.context - and I wondered why you took the approach you've taken. Also, how does this affect things if say your default fails to load.

3. I think if you can set a locale in the DB, then it might be worth having a reset here - otherwise it won't be obviousl how to return to a state where your browser selects the locale for you.

4. Finally - can you explain the tooling around this?
    a) We probably need to add some docs surrounding how one can contribute to translations and;
    b) It'd be nice to know what dev tools are useful here.

We can add more docs in another PR - but if you could detail some of the tooling here so I can dig further that'd be helpful.

Thanks!

shinuza:
> 
> 
> Ok, I've taken a look - and have some comments;
> 
>     1. It looks like you define the list of locales in the Globals file, however - the locales themselves come via external json files. I think that perhaps the locale list should prolly be defined in the config.json too because of this.  That way, users can test new locales without necessarily having to re-compile Fluidd.
> 

Yeah sure, that works.

>     2. I notice a lot of documentation (incl the vue i18n docs) suggest loading the locales via require.context - and I wondered why you took the approach you've taken. Also, how does this affect things if say your default fails to load.
> 

I tried the simplest approach that worked, but I willing to change that. Since the keys are in English, if something wrong happens with the files, the UI will stay in english.

>     3. I think if you can set a locale in the DB, then it might be worth having a reset here - otherwise it won't be obviousl how to return to a state where your browser selects the locale for you.
> 

This feature is not included in this PR. I will work on it once this is merged.

>     4. Finally - can you explain the tooling around this?
>        a) We probably need to add some docs surrounding how one can contribute to translations and;
>        b) It'd be nice to know what dev tools are useful here.
> 

As for the dev tools, I don't think anything exists for i18n. I used [i18n](https://github.com/lokalise/i18n-ally/) to quickly translate the strings, but the 💩 part was to mark the strings and extract them.
As I mentioned on Discord, there's a tool that extracts the strings and can create an new locale file with empty translations, this will probably be more engaging for people who want to introduce a new locale from scratch.

The problem is that there's this so-called feature that detects periods in the string and asserts it a namespace.

If you want to group your translations under `$t('ui.toolhead')` it will expect the translation file to be `{"ui": { "toolhead": "Toolhead" }}`

So the for the following string

```js
$t('Fluidd is cool...')
```

will yield

```
{ "Fluidd is cool": {"": {"": {}}}
```

I'm writing a PR to add an option to deactivate this, but I'm not sure it will get merged...

> 
> We can add more docs in another PR - but if you could detail some of the tooling here so I can dig further that'd be helpful.
> 
> Thanks!

Also:

- I'm planning to add guidelines on how to contribute to the translations.
- I still need to figure out how to access the Vuei18n instance outside of Vue context ( e.g when I can't do `this.$t('mystring')` )

src/components/cards/KlippyCard.vue
```diff
@@ -7,9 +7,9 @@
     <v-card-text>
       <v-row>
         <v-col cols="12" sm="auto">
-          <btn v-if="!klippyConnected" block color="primary" @click="serviceRestartKlipper" class="me-2 mb-2">Restart Klipper</btn>
-          <btn v-if="klippyConnected" block color="primary" @click="restartKlippy" class="me-2 mb-2">Restart Klipper</btn>
-          <btn v-if="klippyConnected" block color="primary" @click="firmwareRestartKlippy" class="me-2 mb-2">Firmware Restart</btn>
+          <btn v-if="!klippyConnected" block color="primary" @click="serviceRestartKlipper" class="me-2 mb-2">{{ $t('Restart Klipper') }}</btn>
+          <btn v-if="klippyConnected" block color="primary" @click="restartKlippy" class="me-2 mb-2">{{ $t('Restart Klipper') }}</btn>
+          <btn v-if="klippyConnected" block color="primary" @click="firmwareRestartKlippy" class="me-2 mb-2">{{ $t('Restart Klipper') }}</btn>
```
cadriel, line 12: It looks like all three here are translated to the same string, however - the last item is a bit different. The english translation should read 'Firmware Restart' - which is to say, it restarts the mcu's and reloads klippers configuration - versus 'restart klipper' which just reloads the configuration.

cadriel (changes requested):
It appears that the setting page is now missing the language drop down, probably because of all the refactoring done here.

Otherwise, I think this is ready for merge. Re: my other comments, we can work them out after this initial PR I think. I'd like to get these in so at least others can start working on other translations while we finish working out other requirements.

tldr;

- Fix that one translation
- Add language dropdown

shinuza:
> It appears that the setting page is now missing the language drop down, probably because of all the refactoring done here.
> 
> Otherwise, I think this is ready for merge. Re: my other comments, we can work them out after this initial PR I think. I'd like to get these in so at least others can start working on other translations while we finish working out other requirements.
> 
> tldr;
> 
>     * Fix that one translation
> 
>     * Add language dropdown

Actually there were other translations that were missing, I added them and also:

- Added printer status as mentioned in #182 
- Added some missing translations in the file system headers
- Added the select with hardcoded values for now.

I'm changing the locales loading mechanism. And it should be good to go.

PR #182, opened 2021-03-19
i18n with ts and auto register

bomdia:
better than french one... (i'm joking)
howeaver all work and the only strange thing is a strange bug on jobs column,
by default it guess the language to use from within available
the language chooser display a world for automatic and the flag if choosed directly
the language file is a logical tree to better understand the context of the translation
i've already translated to italian.
i'm free to explain better what i've do but the code seems pretty straightforward for me.


bomdia:
i thinks the bug with the columns had to do with the computed property, and i've added a v-for in a place (i don't remember where but the commit say all) to make it easier to translate

bomdia:
@cadriel what u think about this?

cadriel:
You'll likely need to rebase this.

What's the difference between this and the existing PR?

bomdia:
There are flag, the locale auto register himself, the locale file is grouped in a object tree and not a flat file with sentence as key that make easier to translate  and the degree are out of the sentence for a future choice  for switching between them, andare i mapped some state that come from store and that the script that the other  can't see but is to translate for a complete localization

cadriel:
Ok, thank you for your contribution here.

I've reviewed both PR's, and have researched a little more into how I think it should work in Fluidd, and I think the final solution is going to be a combination of what you've done here, and the other PR.

I'll be merging @shinuza 's PR - and then making some adjustments to how I handle the keys. Once i've done that - I'll let you know, so we can get the IT locale in place too.

bomdia:
@cadriel ok as you prefer, i want to give you some impression of the other pr, i think that the choose for the language has to be in a place accessible from all page and that can be the bar or the drawer but not in the ui settings page, another thing is the ability to auto register (just place a single file in the right folder) the new locale for other contributor translate to less thing to do and for you to less thing to review, another **very important thing** that i don't see in the other pr is the translation of the state of the printer that is provided from socket (idle, printing etc..) i've mapped them too, is simple you concatenate the idle in lowercase in the key provided so in translation map you have { idle: 'Inattivo', ...} and the thing remain reactive and translated,
and last thing i know isn't a very important thing but i think that having the flag of the country can help quickly locate the language we want, so when you have decided i can translate to italian as you suggested, i understand english and i'm native to italy

bomdia:
if you want i can try to rebase my work but i never rebased so maybe i can pick new change and use meld to merge and then create a new pr or you can give me some tip on this

shinuza:
> @cadriel ok as you prefer, i want to give you some impression of the other pr, i think that the choose for the language has to be in a place accessible from all page and that can be the bar or the drawer but not in the ui settings page

I disagree, you change language once, not every day.

> another thing is the ability to auto register (just place a single file in the right folder) the new locale for other contributor translate to less thing to do and for you to less thing to review

This will be added in the PR.

> another **very important thing** that i don't see in the other pr is the translation of the state of the printer that is provided from socket (idle, printing etc..) i've mapped them too, is simple you concatenate the idle in lowercase in the key provided so in translation map you have { idle: 'Inattivo', ...} and the thing remain reactive and translated,

The problem with dynamic keys is that they are non-discoverable.

bomdia:
@shinuza in fact you have to read the code to discover it, i have readed and parsed the code to translate it, you choseed the simple and poor way of making use of a script without put the needed effort, compare my pr With your with a software like meld and merge my work on it :)

shinuza:
> @shinuza in fact you have to read the code to discover it, i have readed and parsed the code to translate it, you choseed the simple and poor way of making use of a script without put the needed effort, compare my pr With your with a software like meld and merge my work on it :)

No, I choose the repeatable way. You still have to mark the strings and therefore read the code to make sense of it. I fail to see your point.

bomdia:
@shinuza My point is that the translation of printer state is necessary and in your pr isn't done, and i criticized that you have used a script for extracting the text strings: i think is bad practice that lead to worse translation. 
Your effort on translating the app was very low you taked a whole month to run a script basically, i've translated the app at hands in 3 days, in a more complete way, respecting i18n best practice to make a locale, that's all.

shinuza:
> @shinuza My point is that the translation of printer state is necessary and in your pr isn't done
It is.

>and i criticized that you have used a script for extracting the text strings: i think is bad practice that lead to worse translation.

On what ground? FIY, the tool is not implemented yet, so I did the extraction by hand as I mentioned.
Once you get some programming experience, you learn that tool that helps keeping consistency are always preferred. E.g: Linting.
The translation are going as good as you make them, if you don't take context in account you will indeed produce low quality translations. I'm assuming you are also Fluent in French as you sem to think there's a link between my method and the quality of my translations?

> Your effort on translating the app was very low you taked a whole month to run a script basically, i've translated the app at hands in 3 days, in a more complete way, respecting i18n best practice to make a locale, that's all.

I suppose it's first contribution and/or you don't follow the development of the app very much. Otherwise I suggest you look at the commit history, you'll see that the commit pace if fairly high. So it does take some commitment to keep such a large PR up to date.
The best-practices you mentioned were also mentioned in the other PR and are being implemented.
Now, I don't really understand why you have that kind of negative attitude, as we all try to make Fluidd a better app for non-english speakers. Since it took it so little time, I'm sure @cadriel will be more than happy to welcome your contribution for Italian once my PR is merged 😉

cadriel:
@bomdia I have merged the other PR - and am reformatting the translation files both in how I handle the keys, and possibly the format too.

I'll let you know once I've done that, and you should then have a good base to add your translations.

PR #184, opened 2021-03-23
i18n refactor

cadriel:
Throwing this here to accept feedback.

1. I feel going with yaml makes things a bit easier to read
2. fleshing the keys out into a logical structure will mean less pain down the road
3. consolidating some of the translations - lots of very common strings that I can simplify

@shinuza @bomdia would appreciate any feedback. I'll probably mess with the keys a bit more as I move the translations across.

Otherwise the initial effort @shinuza has put in re: how the language is applied etc is good for now. I'm mostly worried about the language files themselves and their structure.


bomdia:
A feauture of vue-i18n is to use already existent key with the ability of capitalize uppercase and lowercase them if needed, i think yaml can go without problem, an appoint for the variable in the message, giù fan write them simply without % before { i think is more cleaner to read

src/components/cards/configuration/PrintHistoryCard.vue
```diff
@@ -29,7 +29,7 @@ import { SocketActions } from '@/socketActions'
 })
 export default class PrinterHistoryCard extends Vue {
   handleRemoveAll () {
-    this.$confirm(this.$t('Are you sure? This will clear all history, and printer statistics.').toString())
+    this.$confirm(this.$t('app.history.msg.confirm').toString())
```
shinuza, line 32: Just FIY, this force the evaluation of the string, which also means it's not going to be reactive.

PR #185, opened 2021-03-24
add chinese language

CF3B5:
Translate Chinese UI

### DIFF
--- a/src/store/config/index.ts
+++ b/src/store/config/index.ts
@@ -16,7 +16,8 @@ export const defaultState = (): ConfigState => {
       blacklist: [],
       hosted: false,
       locales: [
-        { name: 'English', code: 'en' }
+        { name: 'English', code: 'en' },
+        { name: 'Chinese', code: 'cn' }
       ]
     },
     cardState: {},
--- a/src/locales/cn.yaml
+++ b/src/locales/cn.yaml
@@ -23,53 +23,51 @@ app:
       history: 任务历史
       config_files: 配置文件
     label:
-      current: 当前
       accel_to_decel: 加速到减速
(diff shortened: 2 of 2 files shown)


QABQ:
兄弟神速啊,哈哈哈

cadriel:
Hi, thanks for this.. very fast!

Does this need RTL support, or is it sufficient in its current form?

cadriel:
Also, it appears there are a lot of missing keys and translations.. if you have this running in a local environment, you can run

```
npm run i18n-extract
```

to see all missing.

src/store/config/index.ts
```diff
@@ -16,7 +16,8 @@ export const defaultState = (): ConfigState => {
       blacklist: [],
       hosted: false,
       locales: [
-        { name: 'English', code: 'en' }
+        { name: 'English', code: 'en' },
+        { name: 'Chinese', code: 'cn' }
```
cadriel, line 20: I think the language name might be better written in Chinese.

cadriel (commented):
As per the previous comment, there are keys missing.

src/store/config/index.ts
CF3B5: Thank you very much, I have updated the latest language file. However, I think the language name is recommended to use unified English, because the language name can be displayed normally when there is no language font file, especially when using kiosk mode.

cadriel:
There have been more changes since your last commit.

If I release before you have a chance to add the missing keys, feel free to create another PR - but this otherwise seems good so far.

cadriel:
This has been merged. I had to rebase and fix a conflict, but didn't push back to your branch so so am closing here.

PR #192, opened 2021-03-25
Fixed end triggered label not being displayed

shinuza:
Before

Before

After

After


cadriel:
Bleh, github has a button whereby it merges develop into your branch, which is horrible. I've instead cherry-picked this commit into develop, and pushed that. So,  this is merged - but github thinks it isn't. Closing. :)

PR #193, opened 2021-03-26
Add dutch translation

LeandroMarceddu:


shinuza:
Thank you for your contribution. We will be looking into it.

src/store/config/index.ts
```diff
@@ -18,7 +18,8 @@ export const defaultState = (): ConfigState => {
       locales: [
         { name: 'English', code: 'en' },
         { name: 'Français', code: 'fr' },
-        { name: 'Chinese', code: 'cn' }
+        { name: 'Chinese', code: 'cn' },
+		{ name: 'Nederlands', code: 'nl' }
```
shinuza, line 22: I take it that your editor didn't read the `.editorconfig` file, can you check if this is not a tab and/or if you have the correct amount of spaces in there?

shinuza (changes requested):

src/store/config/index.ts
LeandroMarceddu: Fixed

shinuza: Thanks

cadriel:
Merged, and thank you!

PR #242, opened 2021-04-13
Language chinese

CF3B5:
Update chinese language


cadriel:
Thank you for the contribution!

I had troubles rebasing this on develop because of the merges from develop - so I grabbed the file from the latest commit and created a new commit for it.

Closing this as the updates are now applied! Thanks!

Issue #240, opened 2021-04-12
Feature request: Mainsail-like "Printing" widget


ssorgatem:
The one thing I like better from Mainsail compared to Fluidd is the "Printing" widget, because, compared to Fluidd:

- It shows all 3 remaining time estimations at the same time.  Very useful since, at least for me, they over and underestime each very differently.
- Bigger preview image. It just looks nicer. With Fluidd sometimes it's hard to tell the object in the thumbnail.
- Layer progress. Very handy in case the print is interrupted for some reason. Also, nice,

It would be nice to be able to optionally add these features to Fluidd's "printing" widget.

cadriel:
Yep. I've been meaning to give the status panel a once over now it's been about for a while.

I'd love to hear suggestions from other members of the community on what you'd like to see here too.

StuSerious:
I think what @ssorgatem was outlining is about what's needed
the one thing I miss about octoprint is the load/print procedure.

StuSerious:
@cadriel here are a couple of ideas on top of what @ssorgatem was suggesting:
- I find that the only thing I'm craving since I left octoprint (other than Spool Manager and Dry Run) is the Load/Print routine. being able to load a print, take a look at parameters and then hit the button to start it,
- an upload or print button on the top bar, to make it easy to have a file uploaded and ready with only one click after entering the UI
- the possibility to enable a sanity check / confirmation pop-up with something like "Is the Build Plate clear?"

hofleverancier:
> Yep. I've been meaning to give the status panel a once over now it's been about for a while.
> 
> I'd love to hear suggestions from other members of the community on what you'd like to see here too.

I'd like the option of showing the ETA next to or instead of time remaining.

cadriel:
Most items discussed here are coming up in the next release, so closing this.

PR #262, opened 2021-04-25
feat: add setting and speed accounting for time left

SilverCory:
Add a setting that tweaks the calculation of time remaining to take into account the speed

Signed-off-by: Cory Redmond <code@cory.red>

**100%**
100 percent rate

**50%**
50 percent rate

**150%**
150 percent rate

**Settings page**
settings page

### DIFF
--- a/src/components/settings/GeneralSettings.vue
+++ b/src/components/settings/GeneralSettings.vue
@@ -65,6 +65,19 @@
           v-model="printTimeEstimateType">
         </v-select>
       </app-setting>
+
+      <v-divider></v-divider>
+
+      <app-setting
+        :title="$t('app.setting.label.account_for_speed')"
+      >
+        <v-switch
+          @click.native.stop
+          v-model="accountForSpeed"
+          hide-details
+          class="mb-5"
+        ></v-switch>
+      </app-setting>
     </v-card>
   </div>
 </template>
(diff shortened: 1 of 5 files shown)


cadriel (commented):
Normally, the time left and total time when at 0 seconds would reflect one another. Perhaps it might be relevant to apply a similar calculation to the total too?

Also, I wonder if its not worth thinking about displaying this a little differently? Rather than adjusting the normal values, maybe instead having an adjusted time field in brackets so it's clear to the user how the estimation has changed from the original because of the speed adjustment.

Finally, I have plans to re-work this area, see ticket https://github.com/cadriel/fluidd/issues/240

So - it might be worth holding off until I have this in place. Thanks for this tho, I think its relevant.

SilverCory:
>Normally, the time left and total time when at 0 seconds would reflect one another. Perhaps it might be relevant to apply a similar calculation to the total too?

Yeah, good point, it could throw people. Goes with the next point.

>Also, I wonder if its not worth thinking about displaying this a little differently? Rather than adjusting the normal values, maybe instead having an adjusted time field in brackets so it's clear to the user how the estimation has changed from the original because of the speed adjustment.

I agree with this completely. It's not 100% accurate but brings the estimate closer.

>Finally, I have plans to re-work this area, see ticket #240
>
>So - it might be worth holding off until I have this in place. Thanks for this tho, I think its relevant.

I look forward to this! If you make issues describing what you want I'll give it a crack.

cadriel:
Hey @SilverCory. I think the latest code in develop covers everything you had here, so closing the PR for now. Feel free to open another one or add suggestions if needed! Thanks!

PR #269, opened 2021-05-03
Update Korean localization for missing keys.

karlla1220:
I found some keys are shown as it is in Korean localization. I copied keys from latest English locale and did simple diff to find missing keys and add few more translation rather than using English.


cadriel:
Thanks. I had to remove one dupe, but otherwise good.

PR #267, opened 2021-05-02
feat(i18n-IT): add Italian translations

StuSerious:
this PR aims to add Italian translations.
#226 seems to be discontinued, so I ripped parts of the translations, made fresher ones and rearranged the .yaml file.

this should be good enough for release as-is, but some members of the discord have offered to rewiew it, so I'll wait for their approval.

@cadriel sorry for not rebasing, I forgot 😅 

massive thanks to darkestbo#9467 and luca2210#5305 for helping out on Discord!

PR #270, opened 2021-05-04
feat(i18n-IT): add Italian translations

StuSerious:
responds to #267 

Signed-off-by: Stefano Serioli <stefano.serioli@outlook.it>


cadriel:
This was merged, but I fast forwarded before force pushing the rebase so it says closed instead :)

PR #297, opened 2021-05-30
Add Firmware Retraction card.

yadotrof:
image
My JS code may look ugly because I never worked with him before.
This PR can be reviewed only as an idea, but if anyone can advise me on how to improve code I'll do my best.

### DIFF
--- a/src/components/widgets/retract/Retract.vue
+++ b/src/components/widgets/retract/Retract.vue
@@ -1,0 +1,92 @@
+<template>
+  <v-card-text>
+    <v-row class="my-0 mb-4">
+      <v-col cols="12" sm="12" class="py-0">
+        <app-slider
+          :label="$t('app.general.label.retract_length')"
+          value-suffix="mm"
+          input-sm
+          :value="retract_length"
+          :min="0"
+          :max="10"
+          :step="0.1"
+          :disabled="!klippyReady"
+          :locked="!klippyReady || isMobile"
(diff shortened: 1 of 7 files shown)


cadriel:
This looks like a great start. Here's some comments;

- I'm wondering if this shouldn't exist on the tune page instead. Happy to talk about this.
- The card should only be displayed if firmware retractions are turned on. This value should exist in `store.printer.printer.fileconfig.settings`

If you're happy with where this is, I can merge and adjust as needed. Thank you for the contribution!

yadotrof:
> I'm wondering if this shouldn't exist on the tune page instead

In my opinion, the retract configuration is a common printing procedure such as speed and temperature configuration. And often it needs to be adjusted at the time of printing.
So I think it should be presented in the dashboard.

The main question in this is "What fundamental difference between tune and home page?". Because now "Tune page" == "Bed mesh page".

Cards on the tune page can`t be hidden and the layout can`t be changed. But users may like to hide the retraction card. This is another argument for the dashboard.

> The card should only be displayed if firmware retractions are turned on. This value should exist in store.printer.printer.fileconfig.settings

Done

OFFTOP:
As retract configuration depends on the filament type, maybe it would be nice to have presets for this setting in the future.

cadriel:
This is looking good! Thanks.

I'll get this merged soon.

cadriel:
Merged, thank you!

PR #305, opened 2021-06-06
Tailscale Tutorial on docs.fluidd

denkyem:


StuSerious (commented):
hey there @denkyem!
3 quick thoughts:

1. how about adding a section to bind the host to TailScale's IP? something like:
```properties
[server]
host: 100.X.Y.Z
```
2. can the PNGs be renamed so that their function is apparent even if not open in an image viewer?

3. can this be expanded to be a generic VPN tutorial or do you think we should split them in different documents? 🤔 

anyways, looks good as a start to me. only thing git-side is that it needs to be rebased and sqashed

thanks for the PR!
~S

eduncan911:
Closing stale PR with unresolved Build errors.

Please feel free to open a new PR, referencing this original PR, once things have been resolved.

PR #306, opened 2021-06-07
fix: division by zero and infinite loop in file upload/download dialogs

twelho:
With `privacy.resistFingerprinting` enabled in Firefox-based browsers the precision of `performance.now()` is reduced to 100 ms. On most machines I've tested this leads to a division by zero in the transfer progress speed calculation when the a file upload/download starts (in less than 100 ms), subsequently the script enters an infinite loop (as speed is assigned `Infinity`) and the tab freezes. Fix this by explicitly requiring a positive delta before starting to calculate transfer speed, I can confirm that editing and saving files through Fluidd now works on all affected machines on my end.

Signed-off-by: Dennis Marttinen <twelho@welho.tech>

### DIFF
--- a/src/mixins/files.ts
+++ b/src/mixins/files.ts
@@ -84,11 +84,15 @@ export default class FilesMixin extends Vue {
       ...options,
       onDownloadProgress: (progressEvent: ProgressEvent) => {
         const units = ['kB', 'MB', 'GB']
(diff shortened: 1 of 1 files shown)


cadriel:
Greatly appreciated, thank you..

Issue #295, opened 2021-05-29
Feature Request: End Time Display


Mitsuma:
## Describe the solution you'd like:
One thing I loved, coming from OctoPrint, was having an addon that showed me the time (as in clock time) when my print will finish.
Right now we only see the remaining time and as silly as it sounds making the effort to calculate the time when it finishes is a little annoying.

Little "Mockup":
![chrome_gE0BWZdPuU](https://user-images.githubusercontent.com/8568089/120086154-d6dc4f80-c0dd-11eb-8596-0721a700294b.png)
I would like 24h but maybe add an option for AM/PM as well. (Hour/Minutes is probably enough.)
Doesn't need to show the full date but might not be a bad option to add, for multi day prints.

## Describe alternatives you've considered:
No alternatives outside of going back to OctoPrint.

## Affected users and/or setups:
Probably a good help for everybody and not just me who is too lazy to calculate the end time.

## Additional Context:
(Add any other context or screenshots about the feature request here.)

cadriel:
Hi, thank you! I'll look into this with the next status card update.

kieraneglin:
@cadriel I just discovered this project and the Vue stack is right in line with what I'm most familiar with.  I'm interested in contributing and would love to take this on as my first issue if it hasn't already been implemented!

cadriel:
Yep, we love contributions. Please feel free to join us on Discord to discuss.

https://discord.gg/GZ3D5tqfcF

cadriel:
This is now in place for next release. Thanks all!

PR #316, opened 2021-06-15
Add absolute end time estimate

kieraneglin:
Addresses #295.

Adds an estimate of the real-world print completion estimation to the print status card.  Has three main states:

- Hidden when not printing
- When the estimate is in the same day a straight hours/minutes end estimate is given
- When the estimate is further in the future the estimated end date is also included

Screen Shot 2021-06-14 at 3 59 35 PM
Screen Shot 2021-06-14 at 3 53 47 PM
Screen Shot 2021-06-14 at 4 03 07 PM

### DIFF
--- a/src/components/widgets/status/PrintStatus.vue
+++ b/src/components/widgets/status/PrintStatus.vue
@@ -20,7 +20,8 @@
               </template>
               <span>{{ $t('app.printer.status.time_left') }}</span>
             </v-tooltip>
-            {{ timeEstimates.remaining }}
+            <span>{{ timeEstimates.remaining }}</span>
+            <span class="grey--text text--darken-2" v-if="printerPrinting && printTimeEstimationsType !== 'totals'"> / {{ timeEstimates.endTime }}</span>
           </div>
           <div class="mb-1 grey--text text--lighten-1">
             <v-tooltip left>
--- a/src/plugins/filters.ts
+++ b/src/plugins/filters.ts
@@ -78,6 +78,20 @@ export const Filters = {
   },
 
   /**
+   * Formats a date from unixtime into a human readable
+   * datetime with optional formats for today and
+   * future datetimes
+   */
+  formatAbsoluteDateTime: (datetime: number, todayFormat?: string, futureFormat?: string) => {
+    const date = _Vue.$dayjs(datetime * 1000)
+    // Including a year doesn't make sense as that'll be clear from context (even on newyears-related edge cases)
+    const defaultFormat = 'MMM D, h:mm A'
+    const appropriateSpecifiedFormat = date.isToday() ? todayFormat : futureFormat
+
+    return date.format(appropriateSpecifiedFormat || defaultFormat)
+  },
+
+  /**
    * Formats a string into camel case.
    */
   camelCase: (string: string) => {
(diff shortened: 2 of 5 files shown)


src/components/widgets/status/PrintStatus.vue
```diff
@@ -20,7 +20,8 @@
               </template>
               <span>{{ $t('app.printer.status.time_left') }}</span>
             </v-tooltip>
-            {{ timeEstimates.remaining }}
+            <span>{{ timeEstimates.remaining }}</span>
+            <span class="grey--text text--darken-2" v-if="printerPrinting && printTimeEstimationsType !== 'totals'"> / {{ timeEstimates.endTime }}</span>
```
kieraneglin, line 24: It's unclear to me if `printTimeEstimationsType !== 'totals'` is redundant when checking for `printerPrinting`

src/plugins/filters.ts
```diff
@@ -77,6 +77,20 @@ export const Filters = {
...
+   * future datetimes
+   */
+  formatAbsoluteDateTime: (datetime: number, todayFormat?: string, futureFormat?: string) => {
+    const date = _Vue.$dayjs(datetime * 1000)
+    // Including a year doesn't make sense as that'll be clear from context (even on newyears-related edge cases)
+    const defaultFormat = 'MMM D, h:mm A'
+    const appropriateSpecifiedFormat = date.isToday() ? todayFormat : futureFormat
+
+    return date.format(appropriateSpecifiedFormat || defaultFormat)
+  },
```
kieraneglin, line 92: There is a similar method above (`formatDateTime`), but I can justify the creation of this new method for a few reasons:

- It's never meant to give a humanized relative offset (eg: "a few minutes from now") so it doesn't cover a large use case of the other method
- The other method could be updated, but it would get quite complex with the new format arguments and logic changes while maintaining some form of backward compat

If there is disagreement there I can see about taking a pass at refactoring the other method

src/store/printer/getters.ts
```diff
@@ -151,6 +153,7 @@ export const getters: GetterTree<PrinterState, RootState> = {
           ? current / progress
           : current
         remaining = total - current
+        endTime = endTime + remaining
```
kieraneglin, line 156: It seems like this (`case 'file'`) is the only situation in which `endTime` would be meaningful.  If that's not so then I'll update the other cases as needed!

tests/unit/filters.spec.ts
```diff
@@ -57,6 +57,19 @@ describe('formatDateTime', () => {
...
+describe('formatAbsoluteDateTime', () => {
+  Vue.use(DayJSPlugin)
+  it('renders numbers as date with offset applied', () => {
+    // This should return based on your browsers local time, given unixtime.
+    const d = new Date(0)
+    const offset = d.getTimezoneOffset()
+    const d2 = new Date(d.getTime() + offset * 60)
+    expect(Filters.formatAbsoluteDateTime(d2.getTime())).to.equal('Jan 1, 12:00 AM')
+  })
+})
```
kieraneglin, line 71: It's unclear to me how to test cases like the `todayFormat` vs. `futureFormat` in a deterministic way.  It'd be simple enough to make a naive test, but I'd need to likely mock `Date`  or DayJS in some capacity to ensure the test works even if run at 11:59pm.

If there's some precedent for mocking time (or similar), please let me know!

src/components/widgets/status/PrintStatus.vue
kieraneglin: Upon more digging it doesn't seem redundant but I'll leave this comment up for posterity

src/store/printer/getters.ts
kieraneglin: Again, on digging I see that this isn't the full story. Will update tomorrow!

tests/unit/filters.spec.ts
cadriel: Yep, i've been meaning to implement `mockdate` (search npm for mockdate). For the moment this is fine I think.

src/plugins/filters.ts
cadriel: This is all good.

PR #336, opened 2021-06-24
Keep bed mesh control options enabled when printer is busy

Adam8234:
Issue: Can't change controls during printing or busy for bed mesh control

Keep bed mesh controls enabled while printer is busy or is printing.


Adam8234:
Closing as it was fixed

PR #337, opened 2021-06-25
Bugfix: gcode viewer stops following progress between layers

kieraneglin:
Under certain (currently unknown) conditions, `fileMovePosition` can get out of sync with `moveProgress` on some layer changes, disabling the follow progress feature.

This is a naive attempt at addressing this based on some existing logic.

### DIFF
--- a/src/components/widgets/gcode-preview/GcodePreviewCard.vue
+++ b/src/components/widgets/gcode-preview/GcodePreviewCard.vue
@@ -157,8 +157,13 @@ export default class GcodePreviewCard extends Mixins(StateMixin, FilesMixin) {
     if (this.followProgress) {
       const fileMovePosition = this.$store.getters['gcodePreview/getMoveIndexByFilePosition'](this.filePosition)
 
+      // In some (yet unclear) cases, fileMovePosition can get out of sync with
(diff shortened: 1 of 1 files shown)


src/components/widgets/gcode-preview/GcodePreviewCard.vue
```diff
@@ -157,8 +157,13 @@ export default class GcodePreviewCard extends Mixins(StateMixin, FilesMixin) {
...
       const fileMovePosition = this.$store.getters['gcodePreview/getMoveIndexByFilePosition'](this.filePosition)
 
+      // In some (yet unclear) cases, fileMovePosition can get out of sync with
+      // the component's notion of moveProgress.  This seems to happen during
+      // layer changes, but not every time.  Possibly some gcode command is getting
+      // misinterpreted.
+      // This "fix" simply forces a re-sync of progress if they get out of sync
       if (fileMovePosition !== this.moveProgress) {
-        this.followProgress = false
+        this.syncMoveProgress()
```
kieraneglin, line 166: I can't figure out _why_ this was the way it was.  It doesn't make much sense to stop following when instead we can re-sync progress.

At any rate, this fix seems to work for me locally

PR #339, opened 2021-06-26
Fix: gcode preview load current file

Biorn1950:
- if the file is in a subfolder the path is incorrect
By: Biorn1950 (biorn@me.com)

### DIFF
--- a/src/mixins/files.ts
+++ b/src/mixins/files.ts
@@ -52,7 +52,7 @@ export default class FilesMixin extends Vue {
 
     if (res) {
       this.cancelTokenSource = Axios.CancelToken.source()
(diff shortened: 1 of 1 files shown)


cadriel:
The root cause of this issue was because file updates come through with the file and path included in one property.. however, in other locations this is provided as a filename and path as separate properties.

In fluidd there is a helper called `formatAsFile` which is supposed to make this conversion, however - a single line was omitted that would have converted the filename to the correct format.

I've since fixed this - and applied the necessary change in commit https://github.com/cadriel/fluidd/commit/32563d37c457d32f2e8e455062eb351a4e9a1ff6

Thanks for spotting! Closing this PR.

PR #356, opened 2021-07-08
Added power "usage" and "humidity" unit support in ThermalChart on temperature axis

brian-edward:
change units display, remain on first y-axis. maybe one needs to "open up" the widget for more sensors, because klipper might supply more than temperature output

### DIFF
--- a/src/components/widgets/thermals/TemperatureTargets.vue
+++ b/src/components/widgets/thermals/TemperatureTargets.vue
@@ -121,7 +121,7 @@
           <td class="temp-actual">
             <v-tooltip left>
               <template v-slot:activator="{ on, attrs }">
-                 <span v-bind="attrs" v-on="on">{{ item.temperature.toFixed(1) }}<small>{{ (item.name.includes('usage')) ? 'W' : ((item.name.includes('humidity')) ? '%' : '°C') }}</small></span>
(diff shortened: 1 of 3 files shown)


cadriel:
I think this isn't quite the right approach here..

I'd like to see these sensors represented as specific types instead, so we can handle them in explicit ways in the UI. At the moment it looks like they're arbitrarily added to the temp graphs based on the name.

If you take a look at the other sensors, they're all specifically typed.

Also, I think these types of sensors may be better represented via a different graph entirely, or otherwise need to have specific thought on how to best represent them - as the data shown isn't necessarily temp related.

Finally - i believe this is dependent on a klipper PR;
https://github.com/KevinOConnor/klipper/pull/4470

brian-edward:
Let me explain, I was asked to submit PRs on this and would not have done so at all by myself. The changes in the PRs  for the Klipper/Moonraker/Fluidd combination are the minimal changes required to achieve the goal (external sensors in Fluidd) without breaking the whole multi-tier concept completely and getting it done within the hour. Looking at the code gave me an idea how much rework it would be to place external sensors into this multi-tier architecture at the "optimal" place. If one does not want to wait, this is the PR to get it working today ;-)

brian-edward:
> dependent

It should not break without it, but not "look" as good. But Fluidd would not look as nice - one may argue - if there were no Klipper and Moonraker, if course.

cadriel:
Yup - understand this is an easy way to get that sensor information into the charts - and good work on accomplishing that goal. :)

brian-edward:
I am going to do no more work in that direction, but I like Fluidd a lot! Amazing work. Thanks for it!

PR #362, opened 2021-07-15
Feat: Date/Time format

Biorn1950:
- add date and time format in settings

By Biorn1950 - biorn@me.com


pedrolamas:
Thanks for your PR @Biorn1950!

The branch currently has some conflicts with base, can you please check on those so we can look into merging this in?

pedrolamas:
I do wonder if it would be preferable to just let people use whatever they want here instead of having a fixed list of formats...

Biorn1950:
> Thanks for your PR @Biorn1950!
> 
> The branch currently has some conflicts with base, can you please check on those so we can look into merging this in?

conflicts fixed.

src/components/settings/GeneralSettings.vue
```diff
@@ -38,6 +38,35 @@
...
+          dense
+          hide-details="auto"
+          :items="[
+            { text: $t('app.setting.dateformat.monthdayyear'), value: 'MMM. DD' },
+            { text: $t('app.setting.dateformat.daymonthyear'), value: 'DD MMM.' }
+          ]"
+          item-value="value"
+          item-text="text"
+          v-model="dateformat">
+        </v-select>
```
pedrolamas, line 53: A small space between the two options would be nice (either a small margin or just adding an extra `&nbsp;` between them)

pedrolamas (changes requested):

pedrolamas (approved):

src/store/config/index.ts
```diff
@@ -54,7 +54,9 @@ export const defaultState = (): ConfigState => {
         zAdjustDistances: [0.005, 0.01, 0.025, 0.050],
         enableVersionNotifications: true,
         confirmOnEstop: false,
-        confirmOnPowerDeviceChange: false
+        confirmOnPowerDeviceChange: false,
+        dateformat: 'MMM. DD',
+        timeformat: 'h:mm a'
```
eduncan911, line 59: @pedrolamas  What about defaulting to ISO 8601 for datetime formats?

    2022-01-26T20:11:23Z

eduncan911:
Branch conflicts also need to be resolved.

Biorn1950:
@eduncan911 this is based on how Craig did it:
Capture d’écran 2022-01-26 à 21 34 14
And this with the 24h and euro date:
Capture d’écran 2022-01-26 à 21 35 25

src/store/config/index.ts
eduncan911: @Biorn1950 good point (about how existing works/looks right now).  I was just thinking of the hardcoded values defaulting above, which now that I think about it is user-facing, not backend-used.

So yeah, ignore my comment.

eduncan911 (approved):
LGTM

Biorn1950:
@eduncan911 yes the idea was just to add the FR about 24h format. But at the same time I did it for the day-month vs month-day before somebody ask for it ;)

PS: think to close the 24h FR in the https://github.com/fluidd-core/fluidd/projects/2 please.

eduncan911:
> PS: think to close the 24h FR in the https://github.com/fluidd-core/fluidd/projects/2 please.

I haven't even dug into Projects yet, as i have limited permissions...

PR #367, opened 2021-07-19
filters.ts: Support endpoints hosted in a subdirectory

zhaofengli:
I host Moonraker behind a reverse proxy under a subdirectory (e.g., `https://print.myhost.tld/moonraker`), and the current behavior of Fluidd is as follows:

1. In the Add Printer dialog, I enter `https://print.myhost.tld/moonraker/`
1. `AddInstanceDialog.vue` concatenates the test URL correctly, successfully tests the endpoint and allows me to add it :heavy_check_mark: 
1. `$filters.getApiUrls` strips the pathname from the URL and the incorrect endpoint is then added :interrobang:
1. I get a sad error screen :frowning_face:

### DIFF
--- a/src/plugins/filters.ts
+++ b/src/plugins/filters.ts
@@ -195,8 +195,8 @@ export const Filters = {
     const _url = new URL(url)
     const wsProtocol = _url.protocol === 'https:' ? 'wss://' : 'ws://'
     const o = {
-      apiUrl: `${_url.protocol}//${_url.host}`,
-      socketUrl: `${wsProtocol}${_url.host}/websocket`
+      apiUrl: `${_url.protocol}//${_url.host}${_url.pathname}`,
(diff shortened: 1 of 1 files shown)


pedrolamas:
Hosting under a sub-path is not something that is currently supported, you might be able to make it work but chances are you are opening yourself to a lot more pain than you already experiencing...

It will be a lot easier to either use a different port (if port 80 is taken in your system) or just use a different root domain (which is what I personally do, via Nginx reverse proxy)

basz:
Miss reading. You want moonraker on a sub path not fluidd itself. 

You can do that by opening dev tools and modifying paths in the appInstances key of local storage. Fluidd adds 'https://example.com/' even if you add 'https://example.com/my-moonraker'. But if you modify the paths in localstorage to include the path it works. Also change the websocket path

PR #378, opened 2021-08-03
docs: adds thumbnails instructions for Cura

pedrolamas:


StuSerious (approved):
Thank you! Seems good to me!

docs/features/thumbnails.md
```diff
@@ -19,6 +19,8 @@ permalink: /features/thumbnails
 
 Both PrusaSlicer and SuperSlicer allow configuration of thumbnails to be generated alongside your `gcode`. Each slicer has a slightly different configuration.
 
+Cura requires configuring a post processing script to create the thumbnails.
```
grigi, line 22: Not strictly true.
If you use the cura-octoprint plugin, if your printer is on BEFORE cura connects to it to see its "capabilites", it will see the support for UFP documents, and upload via that instead.

So this only works if you do all of the followng:
* Cura-octoprint plugin
* octoprint_compat enabled
* printer is on when you laumch cura
* You use the "upload to cura" button.

If you use a different flow, then the post-procesing script would be needed.

pedrolamas: Thank you, I wasn't aware of that! I just tested by disabling the post-processing actions and using the "Print with OctoPrint" button and it worked exactly as you said!

I will update the text to follow that workflow instead, as it seems to make more sense to me than using the post-processing scripts!

grigi: Thanks, this looks great now :smile:

njoyard:
Maybe we should also mention [Cura2MoonrakerPlugin](https://github.com/emtrax-ltd/Cura2MoonrakerPlugin/) ?

Installation is manual since the plugin is not on cura marketplace, but usage is way easier.

Issue #386, opened 2021-08-11
Power on/off should have optional confirmation dialog


pedrolamas:
## Is your feature request related to a problem? Please describe.

Right now, all the `[power]` elements will work without any confirmation whatsoever, something that might be a problem to some

## Describe the solution you'd like:

I think that an optional confirmation popup for the `[power]` elements would be ideal, just like it currently shows for the host reboot/shutdown options.

## Affected users and/or setups:

All users would benefit from this, but specifically those that have the printer and other devices (such as controllable LEDs) on the `[power]` configuration elements.

## Additional Context:

![image](https://user-images.githubusercontent.com/85504/129057162-d9453692-8780-4b8a-bae5-0ddf33b110c2.png)

PR #388, opened 2021-08-11
feat: adds optional confirmation for power device toggle

pedrolamas:
Implements #386, by adding a new setting that allow to indicate if a confirmation dialog should be shown when the user toggles a device power.

### DIFF
--- a/src/components/common/SystemCommands.vue
+++ b/src/components/common/SystemCommands.vue
@@ -161,9 +161,19 @@ export default class SystemCommands extends Mixins(StateMixin, ServicesMixin) {
       })
   }
 
-  togglePowerDevice (device: Device, wait?: string) {
-    const state = (device.status === 'on') ? 'off' : 'on'
-    SocketActions.machineDevicePowerToggle(device.device, state, wait)
+  async togglePowerDevice (device: Device, wait?: string) {
+    const confirmOnPowerDeviceChange = this.$store.state.config.uiSettings.general.confirmOnPowerDeviceChange
+    let res: boolean | undefined = true
+    if (confirmOnPowerDeviceChange) {
+      res = await this.$confirm(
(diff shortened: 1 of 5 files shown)


dweight (approved):
LGTM :)

src/components/common/SystemCommands.vue
```diff
@@ -161,9 +161,19 @@ export default class SystemCommands extends Mixins(StateMixin, ServicesMixin) {
...
-    SocketActions.machineDevicePowerToggle(device.device, state, wait)
+  async togglePowerDevice (device: Device, wait?: string) {
+    const confirmOnPowerDeviceChange = this.$store.state.config.uiSettings.general.confirmOnPowerDeviceChange
+    let res: boolean | undefined = true
+    if (confirmOnPowerDeviceChange) {
+      res = await this.$confirm(
+        this.$tc('app.general.simple_form.msg.confirm_power_device_toggle'),
+        { title: this.$tc('app.general.label.confirm'), color: 'card-heading', icon: '$error' }
+      )
+    }
```
ForsakenHarmony, line 172: Really not a fan of this tbh, maybe something like

```suggestion
    if (
      confirmOnPowerDeviceChange
      && !await this.$confirm(
        this.$tc('app.general.simple_form.msg.confirm_power_device_toggle'),
        {
          title: this.$tc('app.general.label.confirm'),
          color: 'card-heading',
          icon: '$error'
      	}
      )
    ) {
      return
    }
```

pedrolamas: I personally prefer writing as you did above, but the reason I wrote it as I did was for consistency purposes, as there was already code written that way!

eduncan911: i agree with you both.  however, let's keep it consistent with existing coding techniques for now...

eduncan911 (approved):
LGTM

PR #393, opened 2021-08-25
Updated Spanish translation

zhantyzgz:
feat: Updated Spanish translation.

A lot of the translations where either missing, completely wrong, had typos or were weirdly worded. I've done my best to try to maintain a consistent style, but context was hard to deduce in some of the strings so there will probably be some less than ideal translations.

Signed-off-by: Zhant <zhantyzgz@gmail.com>

### DIFF
--- a/src/locales/es.yaml
+++ b/src/locales/es.yaml
@@ -1,61 +1,68 @@
 app:
   bedmesh:
     label:
+      box_scale: Escala de la caja
+      flat_surface: Mostrar plano
+      mesh_matrix: Matriz de la malla
+      probed_matrix: Matriz sondeada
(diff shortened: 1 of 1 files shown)


shawe:
When this must be released or how can I apply this translation to the actual release?

This translation seems to be much better than actual that is extremly confusing to spanish people.

pedrolamas:
> When this must be released or how can I apply this translation to the actual release?

At the moment, this is just on the "dev" branch, but will be available up on the next release (no date scheduled as of yet, but I would hope to be soon!)

shawe:
@pedrolamas thanks for your reply, I suppose that.

PR #400, opened 2021-09-01
Dockerfile Cleanup

CurtisLeeBolin:
Changed image used for stage "unzip"; alpine is much smaller and does not need unzip installed.
Removed ARG declared but never referenced.
Removed RUN previously used to install unzip in debian container.
Removed stage name "image" declared but never referenced.
Removed commented lines.


ForsakenHarmony:
I'd probably just build a single image with

```Dockerfile
RUN wget ${FRONTEND_ZIP_URL} -O frontend.zip && unzip frontend.zip -d /usr/share/nginx/html && rm frontend.zip
```

PR #407, opened 2021-09-09
Update cameras.md

simpat1zq:
Just added some more info on how to add some command line options for mjpg streamer to webcam.txt and a link to some of those options.


Biorn1950:
@simpat1zq this text block don't fit with the rest of the md file as you can see:
Capture d’écran 2022-01-27 à 17 48 29

StuSerious (approved):
Though this is an edge case, I don't see a downside in adding a bit of documentation to help with it!
Approving this, but as @Biorn1950 stated we need the .md change!

~S2

docs/features/cameras.md
```diff
@@ -7,20 +7,21 @@ permalink: /features/cameras
 ---
 
 # Cameras
+
 {: .no_toc }
 
 ---
 
-Fluidd (and FluiddPI) has built in support for web cameras.
+Fluidd (and FluiddPI) have built-in support for web cameras.
```
eduncan911, line 15: has/have
`have` is used with the pronouns I, you, we, and they, while `has` is used with he, she, and it.

since "Fluidd" is a pronoun, I think it needs "have".  correct?

Biorn1950: Not my native language but I would say HAS.

> Have is used with the pronouns I, you, we, and they. Has is used with he, she, and it.

simpat1zq: This is a tricky one. Fluidd is an "it", so for that you would use "has". But if you mention 2 things, Fluidd and FluiddPi in this case, the "it" becomes a "they" in which case you would use "have". **BUT**, since the FluiddPi is in parentheses, it isn't part of the "they" and you are back to a singular Fluidd "has".

If you want to fix it, I would say you either drop the FluiddPi completely and say "has", or drop the parentheses and say "have". No one will care either way lol.  But the way it was is correct and your proposal is not correct.

eduncan911:
image

matmen (changes requested):
Camera settings screenshot should probably be updated to reflect #500
I think it's best to include it here instead of a seperate PR

pedrolamas:
This has been superseeded by #455

PR #425, opened 2021-10-07
Add camera iframe option with height option

dragonnn:
Hi :)
I just started setting up my klipper setup with fluidd and found out that you can not setup camera as a simple iframe.
I am using https://github.com/dans98/pi-h264-to-browser with works much better then any mjpeg options.
Uses less bandwidth and runs with hw accel in your browser but needs to be setup as a simple web page.
So I add an iframe option named "HTTP Page".
I also add to it an height option, it is a not perfect workaround around the iframe not scaling automatic.
Another option to fix that would be onLoad on the iframe but that would work only if the webserwer and camera are on the same origin with isn't always the case.
My first time writing code for vue, so any suggestion are welcome if something isn't up to the standards. 
In general works really well:
image

### DIFF
--- a/src/components/settings/cameras/CameraConfigDialog.vue
+++ b/src/components/settings/cameras/CameraConfigDialog.vue
@@ -66,7 +66,8 @@
             :items="[
               { text: $t('app.setting.camera_type_options.mjpegadaptive'), value: 'mjpgadaptive' },
               { text: $t('app.setting.camera_type_options.mjpegstream'), value: 'mjpgstream' },
-              { text: $t('app.setting.camera_type_options.video'), value: 'ipstream' }
+              { text: $t('app.setting.camera_type_options.video'), value: 'ipstream' },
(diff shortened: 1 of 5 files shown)


Doridian:
Is there anything missing to get this in?

I set up [webrtc-streamer](https://github.com/mpromonet/webrtc-streamer) to get a very low latency (and low bandwidth) camera feed from my printer into a browser, sadly fluidd's options do not allow me to embed that at all.

dragonnn:
For my application I don't miss anything.
Maybe the only complain about is the need to set the height of the camera stream because it is kind a workaround, but I personal don't see an issue with it.
Also if you don't hit a cross-orgin issue the iframe code can adjust it's height automatic from inside if it has the right code for it as far I know.

ForsakenHarmony (approved):

dweight (approved):

PR #433, opened 2021-10-18
docs: Extra explanation on categories

catonlin:
Was very confused on how to move macros to different categories and there was no documentation. So I wrote an explanation on how to do it to help others. 
Signed-off-by: Caton Lin <lin.caton@gmail.com>


eduncan911 (approved):
LGTM. Thanks for helping with the docs!

PR #442, opened 2021-11-01
fix: SI units

matmen:
Currently, Fluidd displays file sizes based on powers of 2 (binary units), but uses units meant for calculations based on powers of 10 (metric units, e.g. kB, MB).
This might be confusing and leads to rounding errors / wrong values being displayed (i.e. showing 7.5GB of RAM instead of 8GB):
RPi RAM

Opting to switch to the use of powers of 10 here because most users know kB only; showing KiB in the UI would be confusing too.

ref: https://en.wikipedia.org/wiki/Byte#Multiple-byte_units

### DIFF
--- a/src/mixins/files.ts
+++ b/src/mixins/files.ts
@@ -35,7 +35,7 @@ export default class FilesMixin extends Vue {
    * Loads a gcode file and parses for the gcode-viewer.
    */
   async getGcode (file: AppFile) {
-    const sizeInMB = file.size / 1024 / 1024
+    const sizeInMB = file.size / 1000 / 1000
     let res: boolean | undefined = true
 
     if (sizeInMB >= 100) {
@@ -92,8 +92,8 @@ export default class FilesMixin extends Vue {
         const delta = performance.now() - startTime
         if (delta > 0) {
           speed = progressEvent.loaded / delta
-          while (speed > 1024) {
-            speed /= 1024
+          while (speed > 1000) {
+            speed /= 1000
             i = Math.min(2, i + 1)
           }
         }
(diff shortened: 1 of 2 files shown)


pedrolamas:
Though I get your point, I'm not sure I agree with the current changes.

Right now, Fluidd uses powers of 2 and this is what the current users expect, so I wouldn't consider this a "fix" as nothing is actually broken if people understand that this is the current behavior.

Having said that, I would be ok with an optional setting that would allow the users to pick what they want to use.

Even if we did go ahead with this change, I'd expect us to start showing "MiB" instead of the current "MB" so that would also need to be changed.

Biorn1950:
First thing if we speak about memory, isn't based on power 2?
Anyway for file and storage space must be in power 10.

And yes should be a Fix in any case since the current management is in power 2 and the unit is in MB ;)

matmen:
I agree with you. I think file sizes should be displayed in MB (/ metric units) and memory related sizes should be displayed in MiB (/ binary units) then. @pedrolamas opinions?

Biorn1950:
@matmen For memory (only) you can have based power 2 value and use MB unit iirc.

pedrolamas:
Here's the thing, can someone give me an example where file sizes displayed in SI units?

Windows displays in non-SI (so multiples of 1024) and Linux, if we take for example `ls -lh` command, that is non-SI units also (only `ls -lh --si` would do SI units)

I'm all in for SI units, but I have yet to see where in IT would someone actually use 1000 multiples instead of 1024, hence my reluctance in accepting this PR...

PR #443, opened 2021-11-02
fix: file system browser selection

matmen:
Currently, when selecting `n-1` files of a file system browser with `n` files, all files are selected. This PR fixes the broken logic for determining if all files should be selected.


pedrolamas (approved):
Thanks for the contribution, I've confirmed the bug exists and this does indeed fix the problem!

dweight (approved):

dweight:
@matmen It looks like you need to update your branch with the changes from fluidd-core:develop yourself (I can't do it from here)

Issue #33, opened 2020-11-29
Allow rotation of camera image


cadriel:
Possibly use black bars and an aspect constrain of some sort.

aepkolol:
I tried to use '-rot 90' in /boot/fluiddpi.txt and it doesn't make a difference

tjengbudi:
is there any eta for this?

sdc53:
FYI this is editable in webcam.txt from the fluidd gui:
camera_raspi_options="-x 1920 -y 1080 -fps 30 -rot 90"

tjengbudi:
@sdc53 looks like it is not from gui right? if i am not wrong you state webcam.txt, this is mjpegstreamer file file configuration right? it is used driver/hardware capability to used it right?
the first problem is i used logitech c270 instead raspi cam. the driver said it cannot rotate. that's why i asked to rotated within web code (html5 example by mjpegstreamer can do it) so the second problem is, looks like fluidd use vue, i am not understand at all. the third problem if i rotate the camera, it will crash the bed. another position is not an option cause i used enclosure and i might dumb at design also

PR #452, opened 2021-11-28
Allow rotation of camera image

max-scopp:
Really needed that feature.

I can also make it so the black bars are avoided, but this would require more effort in resizing the container.

Requires further testing on IP Cameras and HTTP pages.
HTTP pages might cause weird behavior due to the `height` setting.
IP cameras might cause weird behavior due to visible controls of the HTML5 video tag.

Tested:
- [x] MJPEG Adaptive
- [x] MJPEG Stream
- [ ] IP Camera
- [ ] HTTP page

Resolves https://github.com/fluidd-core/fluidd/issues/33

image


grrminator:
how does one put this on fluiddpi?

max-scopp:
Either run a

max-scopp:
It's not a good (non-permanent) solution, but you can clone my fork on your computer, build the frontend and copy the resulting contents of the created `dist/` folder onto your pi. Klipper is usually installed under `~/fluidd`. However, I'm not sure how the updater would handle this.

Alternatively, you can create a web server somewhere and access it from there. Simply go through the install process again and point it to the pi's ip/host.

point-source:
What is this still waiting on for a proper merge? Review?

src/components/settings/cameras/CameraConfigDialog.vue
```diff
@@ -58,6 +58,40 @@
...
+              {
+                text: $t('app.setting.camera_rotate_options.180'),
+                value: '180',
+              },
+              {
+                text: $t('app.setting.camera_rotate_options.270'),
+                value: '270',
+              },
+              {
+                text: $t('app.setting.camera_rotate_options.360'),
```
matmen, line 82: I don't really see a need for a 360° rotation

src/components/widgets/camera/CameraItem.vue
```diff
@@ -77,16 +70,33 @@ export default class CameraItem extends Vue {
...
-    transforms += (config && config.flipX) ? ' scaleX(-1)' : ''
-    transforms += (config && config.flipY) ? ' scaleY(-1)' : ''
-    // transforms += ` rotate(${config.rotate}deg)`
-    return (transforms.trimLeft().length) ? { transform: transforms.trimLeft() } : {}
+
+    if (!config.rotate) {
+      transforms += config && config.flipX ? ' scaleX(-1)' : ''
+      transforms += config && config.flipY ? ' scaleY(-1)' : ''
+    } else {
+      let scaling = element.clientHeight / element.clientWidth
```
matmen, line 87: the scaling works fine for 90° and 270° rotation, though i am not sure about the black bars (they are probably not easily removable because of the footer text tho; and maybe even fine considering each camera view is meant to split the card evenly):
90 deg rotation

for 180° (and 360°) the scaling is quite off:
180 deg rotation

src/locales/de.yaml
```diff
@@ -262,9 +262,16 @@ app:
       mjpegadaptive: MJPEG Adaptive
       mjpegstream: MJPEG Stream
       video: IP Camera
+    camera_rotate_options:
+      none: Keine
+      90: 90°
+      180: 180°
+      270: 270°
+      360: 360°
```
matmen, line 270: see first comment

src/locales/en.yaml
```diff
@@ -335,9 +335,16 @@ app:
       mjpegstream: MJPEG Stream
       video: IP Camera
       iframe: HTTP page
+    camera_rotate_options:
+      none: None
+      90: 90°
+      180: 180°
+      270: 270°
+      360: 360°
```
matmen, line 343: see first comment

src/locales/en.yaml
```diff
@@ -335,9 +335,16 @@ app:
...
     label:
       all_off: All off
       all_on: All on
+      camera_rotate_by: Rotate By
```
matmen, line 347: probably lower case "by"?

src/store/cameras/types.ts
```diff
@@ -12,5 +12,6 @@ export interface CameraConfig {
   fpstarget?: number;
   flipX: boolean;
   flipY: boolean;
+  rotate?: '90' | '180' | '270' | '360';
```
matmen, line 15: see first comment

src/locales/de.yaml
```diff
@@ -262,9 +262,16 @@ app:
       mjpegadaptive: MJPEG Adaptive
       mjpegstream: MJPEG Stream
       video: IP Camera
+    camera_rotate_options:
+      none: Keine
```
matmen, line 266: "Drehen um: Keine" seems a bit off, maybe replace this with "Keine Drehung" to be more verbose?

matmen (changes requested):

src/components/widgets/camera/CameraItem.vue
max-scopp: Yes, there are some questionable cases that would require scaling the parent elements, so I didn't intend to change too much. Maybe I'll try it for the next iteration.

My camera is currently not usable on my printer, but I'll check this presumably Monday/Tuesday evening. I also tried it with my 1080p camera, which worked on a maximized browser window. Can you leave your browser resolution and just in case the camera resolution? I don't see a reason it would be different as long as the aspect ratio stays the same, but oh well.

matmen: I'm using Chrome on Win10 on a 1080p monitor with a 1080p camera. I assume scaling at 180° can be omitted completely, because it will just result in the same canvas size.

matmen (dismissed):
LGTM

image

matmen:
One small issue, actually: When collapsing the camera view card, the animation frame handler throws `Uncaught TypeError: Cannot read properties of undefined (reading 'style')`. Probably just needs a `if(!image) return;`, though.

Also I'm not quite sure about the recursive call of animate, as it pushes to the stack with every frame. Not sure if Vue is doing some magic in the background to keep the stack from overflowing (need some time to test this, my browsers max stack size is 13953, so 15 minutes at 15fps for it to overflow. I'll comment on here when I know more).

matmen:
Doesn't seem like it overflows. Just the TypeError then. Nice work!

max-scopp:
Yikes! Totally forgot about collapsing. Good eye on the QA though!

A simple return would have worked, but the current solution is more appropriate since the error only occurs when trying to transform a frame where the component was already being unmounted/destroyed. But, just in case, I added the missing `undefined` to the refs' type.

matmen (approved):
LGTM

matmen (approved):

matmen:
docs: see https://github.com/fluidd-core/fluidd/pull/500#issuecomment-1048247653

PR #460, opened 2021-12-15
add remaing Filament weight in Printer Status

TheTiEr:
add the remaining filament weight in the printer status widget. 

This could be helpful if you have to change the spool midprint and keep sure that there is enough material on the spool. This helps with sleepless nights


eduncan911 (commented):
Could you post a screenshot of this working?

src/components/widgets/status/StatusTab.vue
```diff
@@ -341,6 +341,24 @@ export default class StatusTab extends Mixins(StateMixin, FilesMixin) {
     return this.$store.state.printer.printer.current_file.filament_total || 0
   }
 
+  /**
+   * Total filament weight according to the current file / slicer.
+   */
+  get filament_weight_total () {
```
eduncan911, line 347: could you remove this extra function, and put the single line between 354 and 355 below?

src/components/widgets/status/StatusTab.vue
```diff
@@ -71,7 +71,7 @@
               </status-label>
 
               <status-label :label="$t('app.general.label.filament')">
-                <span v-if="filament_used > 0 && printerPrinting">{{ $filters.getReadableLengthString(filament_used) }}</span>
+                <span v-if="filament_used > 0 && filament_total > 0 && printerPrinting">{{ $filters.getReadableLengthString(filament_used) }} / {{ $filters.getReadableLengthString(filament_total) }} <br /> remaining: {{filament_weight_needed.toFixed(2)}} g  </span>
```
eduncan911, line 74: so if `filament_total` is not `> 0`, this won't display at all now?

not sure we would want to hide it if so.  could we remove the condition `&& filament_total > 0`?

eduncan911 (changes requested):

TheTiEr:
Fluidd remaining Filament

i have added and changed the code to fit your annotations. Above you can see it working. 
I use Cura as Slicer. To work with Cura i had to add a plugin, which i had to write on my own, so the needed weight of the filament is written in the header.

eduncan911 (dismissed):
LGTM, pending @pedrolamas review.

eduncan911:
Yes, I was wondering how this would look, hence the screenshot request.

@pedrolamas it may be too much to ask in this PR, maybe we can track in a other issue, but what about a config setting to display either or.  And whichever is not displayed, because the tooltip mouse hover?

pedrolamas:
I'm ok we add a setting to further configure this, but I still think we should do something on how it currently looks. I would prefer that we either break this into 2 labeled entries or just set the 2nd line as a tooltip.

eduncan911:
@TheTiEr what do you think to our suggestions?

One label, with tooltip/mouse hover?
Or, two labels?

The config option can be tracked in another issue.

I would lean towards two labels, if it was me (and hide the label we don't have, like originally coded)

pedrolamas:
Following up on this topic, I just opened this feature request: https://github.com/Ultimaker/Cura/issues/11343

eduncan911 (changes requested):
Just a tweak to the UI labels as we discussed above.  Let us know if you have other suggestions.

Biorn1950:
IMHO these kinds of features must be implemented to be compatible with all the main slicers and with native values or not implemented. 

Plus adding an option in the settings to choose which kind of value you want. eg grams/meter/percent (and if you want it)

Same thing about the time like that was before. I find very annoying to have a list of value which you don't care about.

It's a very niche case, need cura + need a plugin in cura.

TheTiEr:
> @TheTiEr what do you think to our suggestions?
> 
> One label, with tooltip/mouse hover? Or, two labels?
> 
> The config option can be tracked in another issue.
> 
> I would lean towards two labels, if it was me (and hide the label we don't have, like originally coded)

I think this option is the best too. Show the remaining weight if it is available else the needed length in meters. 

The cherry on the top would be to show the remaining weight and while hovering over it, it will show all other informations, like meters and percent.

pedrolamas:
> It's a very niche case, need cura + need a plugin in cura.

I just confirmed, and Cura works out of the box (no extra plugin required!)

https://github.com/Ultimaker/Cura/issues/11343#issuecomment-1023133340

Biorn1950:
> > It's a very niche case, need cura + need a plugin in cura.
> 
> I just confirmed, and Cura works out of the box (no extra plugin required!)
> 
> [Ultimaker/Cura#11343 (comment)](https://github.com/Ultimaker/Cura/issues/11343#issuecomment-1023133340)

After a quick check `filament_weight_total` is already managed by Moonraker for the main slicers.So effectively it's ok on this point. ( @TheTiEr maybe your slicer is outdated and don't have the lasts placeholders)

IMO a select list in the settings for the filament unit stay the best solution. That not exclude to have a passover like @eduncan911 suggested.

PR #391, opened 2021-08-24
fix: display correct fan speed when using 'fan max_power' setting

Vhab:
Fixes the Fan slider misbehaving when the 'max_power' setting is used. (Fixes #390)

As the slider goes from 0 to 1, I did have the option to adjust the maximum value of the slider to max_power. This is what Mainsail has done.

However, it felt less intrusive to adjust the value received from Moonraker rescaled to 0-1 using max_power instead.

PR #477, opened 2022-01-26
Feat: Non-controllable fan On/Off

Biorn1950:
Changing the status of uncontrollable fan to On/Off instead of useless percentage.
In the same minding of this #391 .

Signed-off by Biorn : biorn@me.com

### DIFF
--- a/src/components/widgets/outputs/OutputFan.vue
+++ b/src/components/widgets/outputs/OutputFan.vue
@@ -45,7 +45,7 @@ export default class OutputFan extends Mixins(StateMixin) {
   get prettyValue () {
     return (this.value === 0)
       ? this.$t('app.general.label.off')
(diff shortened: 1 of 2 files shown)


src/components/widgets/outputs/OutputFan.vue
```diff
@@ -45,7 +45,7 @@ export default class OutputFan extends Mixins(StateMixin) {
   get prettyValue () {
     return (this.value === 0)
       ? this.$t('app.general.label.off')
-      : `${this.value.toFixed()}<small>%</small>`
+      : this.$t('app.general.label.on')
```
eduncan911, line 48: nice catch

eduncan911 (approved):
LGTM

PR #478, opened 2022-01-26
Feat: make each axis controllable after homing

Biorn1950:
Give the possibility to move an axis if the others are not homed.

like that:
Capture d’écran 2022-01-26 à 18 15 40
instead of that:
Capture d’écran 2022-01-26 à 18 36 36

Signed-off by Biorn : biorn@me.com


eduncan911 (approved):
I'll have to find time to pull down the source and set it up to check all of this Vue code.  Looks good at 2 or 3 glances.  ;)

Will wait for @pedrolamas to give a second review as he may be more versed in Vue than I.

pedrolamas (approved):
Thanks @Biorn1950, it's working great!!

PR #479, opened 2022-01-26
Feat: alias for macro

Biorn1950:
Signed-off-by: biorn biorn@me.com


eduncan911 (approved):
LGTM

mjparme:
How do you use this feature? There doesn't seem any intuitive way to set the alias for the macro in the UI.

PR #481, opened 2022-01-27
Show Moonraker warnings

pedrolamas:
I've mad some "mistakes" on my Moonraker config so I could show how this would look:

image

After this gets merged, I will start working on Klipper deprecation warnings, and will improve messaging for this also.


Biorn1950:
Is it useful to have the fluidd warning part since there is the moonraker with the detailed issue?

eduncan911 (dismissed):
LGTM.  Just review the comment from Biorn1950 .

Biorn1950:
> @Biorn1950 so drop first line of it? Sounds good to me!!

Yes only the first line

pedrolamas:
> I think this part should be removed.

The last 2 lines on that screenshot we need to keep as it might be completely different from the other part I'm adding!

Those are the moonraker "failed components", while what I am adding is moonraker "warnings"

eduncan911 (approved):

PR #483, opened 2022-01-28
Feat: Add reset history

Biorn1950:
- add the new reset history endpoint
- fix update job instead of clear

Signed-off-by: Biorn1950 biorn@me.com


Biorn1950:
I have to remove bad definition for `history.msg.confirm` and i18n Ally refracted at the same time. Which is not a bad thing IMHO.

src/components/widgets/stats/PrinterStatsCard.vue
```diff
@@ -105,6 +118,18 @@ export default class PrinterStatsCard extends Vue {
   get supportsHistoryComponent () {
     return this.$store.getters['server/componentSupport']('history')
   }
+
+  handleResetStats () {
+    this.$confirm(
+      this.$tc('app.history.msg.confirm_stats'),
+      { title: this.$tc('app.general.label.confirm'), color: 'card-heading', icon: '$error' }
+    )
+      .then(res => {
```
pedrolamas, line 127: My personal preference goes to use `async..await` when possible (it should work fine here) but I have no problem with promise-style code!

src/api/socketActions.ts
```diff
@@ -346,19 +346,25 @@ export const SocketActions = {
...
+        dispatch: 'history/onDelete',
         params
       }
     )
   },
 
+  async serverHistoryResetTotals () {
+    baseEmit(
+      'server.history.reset_totals', {
+        dispatch: 'history/onHistoryChange'
```
pedrolamas, line 363: This has no payload, so what will this actually do?

src/components/widgets/stats/PrinterStatsCard.vue
pedrolamas: Fair enough! 🙂

pedrolamas (approved):
LGTM!

PR #486, opened 2022-01-28
feat: enhanced console filtering

randellhodges:
This adds the ability to define custom console message filtering


randellhodges:
I need to make a tweak so it sets the focus when the dialog is opened. I will push changes for that later and update this comment when I am done.

pedrolamas:
This sounds like an interesting feature to have, but I think it would be a lot more useful if one can enable/disable directly on the Console view instead of using the Settings

image

randellhodges:
> This sounds like an interesting feature to have, but I think it would be a lot more useful if one can enable/disable directly on the Console view instead of using the Settings
> 
> image

I can look into adding that. I doubt there would be a lot of these custom filters so that list shouldn't expand out too much.

randellhodges:
Done

pedrolamas:
Thank you for this @randellhodges, I just tested this locally and it works great!!

pedrolamas:
@randellhodges I've just sent a PR to your repo (sorry, we are still working out the repo permissions here...) to split that `regex` property out of the `ConsoleFilter` interface (thus keeping it on a "pure" state, always ready to serialize as is)

randellhodges:
@pedrolamas Merged. I had thought about the same thing but was concerned about things getting out of order with using the index of one to get the expr from the other, but looking at your changes, I don't see that being a problem. If that was ever a problem, we could store the expr in a dictionary and use the id of the filter as the key into the expr "cache".

pedrolamas (approved):
LGTM!

Biorn1950:
@randellhodges it's a great idea, thanks.
I see an improvement which could be very helpfull for the majority of people who don't know regex, is a field where you can add a string to include/exclude. (with a switch)
Or
- two text field `include` and `exclude`

What you thing about that?

pedrolamas:
> I see an improvement which could be very helpfull for the majority of people who don't know regex...

To be honest, I see this more of the "poweruser" kind of setting, so if someone doesn't know regular expression syntax, then perhaps they should not be playing with this?

Biorn1950:
> so if someone doesn't know regular expression syntax, then perhaps they should not be playing with this?
They can't use it if it's regex only, but with my proposal everybody can use the filter.

randellhodges:
I could see a starts with, or contains type of option with a word and then converting that into a regex.

In my opinion, that would be a something to consider as a second phase or enhancement after this is merged. If we attempt to solve everything at once, we'll never get anything out there.

I would be willing to do the contains/starts with type thing at a later time.

Biorn1950:
> I could see a starts with, or contains type of option with a word and then converting that into a regex.

I didn't think about options. I mean it's a good solution.

Biorn1950 (approved):
LGTM

randellhodges:
wait one, I broke something when I went to rebase the latest changes

randellhodges:
My git isn't good and I overwrote your changes @pedrolamas.  I'm not sure how to remerge them, but I can see them, so I will re-apply them myself since that won't take a long as it will for me to figure out as git.

randellhodges:
For the contains/starts with, do you think this would ease the use for simple rules?

Case insensitive. Turns the value into an expression like .*{value}.*/i
image

Case insensitive. Turns the value into an expression like ^{value}.*/i
image

Passes the input as an express verbatim
image

Biorn1950:
I think it's very good like that.
- idea of type: contain, not contain, startWith, startWithout, expression.
- maybe I would make a larger text field

randellhodges:
For the contains/starts with, do you think this would ease the use for simple rules?

Case insensitive. Turns the value into .*{value}.*/i
image

image

image

randellhodges:
I've been avoiding the "not contain" and "without" ones because those types don't play well with others.

Not contains: timelapse
Contains: something

Every line would get filtered out unless it had the word timelapse in it, and those that do have timelapse, if it had "something" in it, it would get filtered out. My opinion is that going to cause more support/questions. If the community wants these "not" options, perhaps just putting a checkbox to "reverse" the match criteria. But again, probably overcomplicating the feature.

I agree on the text field/dialog box. I just have to figure out how to do that.

pedrolamas (approved):
Some issues that I've now fixed (check last 2 commits) but overall, this is working perfectly!

Biorn1950 (approved):
LGTM

Issue #467, opened 2022-01-05
Entering a number into the Z field uses the default XY speed


Makin-Things:
## Affected Version(s):
(The versions of fluidd this issue occurs on.)
- klipper version: v0.10.0-221-g8a3727ef 
- moonraker version: v0.7.1-280-g9d7baa1 
- fluidd version: v1.16.2 

## Platform:
Raspberry Pi 3 running FluiddPi

## Hardware Configuration:
TwoTrees Sapphire Pro (Makebase Nano v1.2)

## Reproduction Steps:
1. Home machine - all good
2. Use the arrows in the GUI to move the Z access up and down - all good, it uses the Default toolhead Z speed
3. Type a number in the Z field - all bad, it uses the Default toolhead XY speed which the machine is not capable of doing.

## Actual Results:
Lots of noise from the Z stepper and no movement.

## Screenshot(s):
(Images or gifs showing the problem, perhaps before/after.) 
![image](https://user-images.githubusercontent.com/91187456/148172963-92cb18e1-86cd-47bf-874d-54de6cf85c49.png)
![image](https://user-images.githubusercontent.com/91187456/148173055-07ca19f7-5d25-4d59-97ec-cb33f8f3b6c3.png)
At 17:39:06 and 17:39:10 the Z arrows ini the GUI where pushed. At 17:39:17 a number was entered into the Z field of the GUI. All 3 should have been executed at a feedrate of 600.

## Expected Results:
When entering a number into the Z field it should use Default toolhead Z speed for the feedrate.

## Log Files:
(The log files fluidd provides at Configure > System Control.)
- klipper log: N/A
- moonraker log: N/A

## Additional information
(Extra information relevant to the issue, links to similar issues.)

Issue #427, opened 2021-10-11
[v1.16.2] Editing Z coordinate uses XY speed instead of Z speed (caution!!)


batista16:
## Affected Version(s):
- klipper version: v0.10.0-39-g5f184e3f
- moonraker version: v0.7.1-67-gc631bff 
- fluidd version: v1.16.2

## Platform:
FluidPi on Rpi, using Chrome

## Reproduction Steps:
If you use the control arrows for Z, it uses the correct speed value configured in the "tool" section. If you change the Z coordinate in the position input boxes, it uses the XY speed instead of the Z speed. It's scary! 

## Expected Results:
Changing the Z coordinate should use the Z speed, no XY speed.

Biorn1950:
Hello, I'm not able to reproduce this issue. can you share a screen of the console history and the settings please?

PR #488, opened 2022-01-28
Fix: Z move speed issue

Biorn1950:
Relative to: https://github.com/fluidd-core/fluidd/issues/467 and https://github.com/fluidd-core/fluidd/issues/427

Signed_off_by: Biorn1950 <biorn@me.com>


pedrolamas:
@Biorn1950 can you check `sendMoveGcode` method in `ToolheadMoves.vue`?

It shows the same fix as you have here, but there's also a check for `inverted` that might make sense here too!

Biorn1950:
> @Biorn1950 can you check `sendMoveGcode` method in `ToolheadMoves.vue`?
> 
> It shows the same fix as you have here, but there's also a check for `inverted` that might make sense here too!

Invert is to have to button inverted (left and right) not to switch the value in positive and negative.

pedrolamas (approved):

PR #492, opened 2022-01-29
Fix: directory sorting

trapeznikov-ap:
The top directory ".." should always be the first in the list when sorting


pedrolamas:
Hi @trapeznikov-ap, thank you for your PR.

From what I can see on my own machine, the ".." always appears on top, no matter what file list I'm looking at, so I was just wondering where did you see this behavior broken?

trapeznikov-ap:
> Hi @trapeznikov-ap, thank you for your PR.
> 
> From what I can see on my own machine, the ".." always appears on top, no matter what file list I'm looking at, so I was just wondering where did you see this behavior broken?

At me the problem is shown at sorting in descending order by names. Tell me the sort direction and which column are you sorting by?

pedrolamas:
Can you send a screenshot showing this behavior? Are you using cyrillic chars for your filenames and folders?

From me, using regular chars it is working as normal when ordering by Name ascending and descending, so ".." is always on top.

trapeznikov-ap:
image
image
image

pedrolamas (approved):

Biorn1950 (approved):

PR #493, opened 2022-01-29
chore: Adjust page title order when printing

ChipCE:
Currently, when printing, the page title is in <code>`${instanceName} | ${pageName} | [${this.progress}%]`</code> format with the display result is <code>[Percent ring icon] MyPrinterName | Dashboard | xx%</code>. Due to the long title string and width limit of browser tab, the percent number more likely to be trimmed down. There is percent ring icon, but it's hard to tell the real number and progress, specially on long print.
By moving the percent number to the front (<code>[Percent ring icon] xx% | MyPrinterName | Dashboard</code>), it will make user easier to have quick glance about the print progress without have to switch thru browser tabs. It's not a big impact on user with a single printer, but for people with multi instance of fluidd running (farm for example), I think this change will help the manage job easier.

Signed-off-by: John Smith <login721@gmail.com>


shinuza:
Hi thanks for your contribution. This is a good idea but a tad opinionated, maybe we could offer multiple options and in a future revision offer a template with variable substitution. @StuSerious What do you think?

src/App.vue
```diff
@@ -122,7 +122,7 @@ export default class App extends Mixins(StateMixin) {
     const pageName = this.$route.name
 
     if (this.printerPrinting) {
-      return `${instanceName} | ${pageName} | [${this.progress}%]`
+      return `[${this.progress}%] | ${instanceName} | ${pageName} | [${this.progress}%]`
```
matmen, line 125: the other ` | [${this.progress}%]` can probably be removed, no?

matmen (changes requested):

ChipCE:
@matmen Oups, sorry! 
Fixed it.

pedrolamas:
Personally, I like this change, but will wait for further comments from other users.

@shinuza makes a good point of having this as a configurable setting, but in the end I do think it would be more useful and simpler to have it only as the proposed implementation in this PR.

shinuza:
I'd say we should merge this and then add the option to change it in options if people want the old behavior back 😄

shinuza (approved):
LGTM.

pedrolamas (approved):

Issue #343, opened 2021-06-28
Pause auto-scroll automatically when scrolling up in console


AndrewEllis93:
An extremely minor thing - when I scroll up in the console, it will keep bouncing me to the bottom if new messages are coming through, until I uncheck auto-scroll. 

It would be great if auto-scroll would automatically pause when scrolling up, and maybe show a buttom to jump back down to the bottom / resume.

PR #495, opened 2022-01-29
feat: pause autoscroll when back scrolling

randellhodges:
When scrolled up, it will "pause" the auto scroll to bottom until you scroll down or click the button to scroll back to the bottom. The button will only be displayed when scrolling is "paused".

This is for https://github.com/fluidd-core/fluidd/issues/343

This was lightly tested with Firefox and Edge in Ubuntu.  Additional testing would be appreciated.

### DIFF
--- a/src/components/widgets/console/ConsoleCard.vue
+++ b/src/components/widgets/console/ConsoleCard.vue
@@ -21,6 +21,24 @@
     </template>
 
     <template v-slot:menu>
+
+      <v-tooltip
+        v-if="scrollingPaused"
+        top>
+        <template v-slot:activator="{ on, attrs }">
+          <v-btn
+            @click="console.scrollToBottom(true)"
+            v-bind="attrs"
+            v-on="on"
+            fab
+            small
+            text>
+            <v-icon>$down</v-icon>
+          </v-btn>
+        </template>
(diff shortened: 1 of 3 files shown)


pedrolamas (changes requested):
There seems to be a styling/layout issue here, as the new "scroll to bottom" button is bigger than the existing ones:

image

src/components/widgets/console/ConsoleCard.vue
```diff
@@ -21,6 +21,24 @@
...
+            v-bind="attrs"
+            v-on="on"
+            fab
+            small
+            text>
+            <v-icon>$down</v-icon>
+          </v-btn>
+        </template>
+        <span>{{ $t('app.console.placeholder.scroll') }}</span>
+      </v-tooltip>
```
pedrolamas, line 40: From my understanding, the `app-btn` is the preferred way to get this button here:

```suggestion

      <app-btn
        v-if="scrollingPaused"
        @click="console.scrollToBottom(true)"
        color=""
        fab x-small text>
        <v-icon>$down</v-icon>
      </app-btn>
```

pedrolamas:
I've fixed the layout issues using the proposed solution!

pedrolamas (approved):

matmen (approved):

PR #497, opened 2022-01-30
 Fix: firmware retractation length stuck at 0

### LINKED ISSUE #366: v1.16.2 cosmetic bug: firmware retraction values less than 0.25mm display as 0mm on GUI.
probably an old bug. front GUI will not display firmware retract retraction length properly if value is less than 0.25.

## Affected Version(s):
(The versions of fluidd this issue occurs on.)
- klipper version: v0.9.1-621-gdafb74e3 
- moonraker version: v0.7.1-25-g7f740cc -> v0.7.1-29 
- fluidd version:  v1.16.2

## Platform:
firefox

## Hardware Configuration:
N/A

## Reproduction Steps:
enable firmware retraction in printer.cfg with this:

[firmware_retraction]
retract_length: 0.249
retract_speed: 60
unretract_extra_length: 0
unretract_speed: 60

## Actual Results:
GUI Retraction Length displays as 0mm. GET_RETRACTION will return the correct value which proves this is likely just a cosmetic bug.

## Expected Results:
Retraction length should display 0.249.

## Additional information
retract_length: 0.25 and above will display correctly.

Biorn1950:
- if value < 0.25 was rounding to 0 the max value.
( fix #366)

Signed-off-by: Biorn <biorn@me.com>

### DIFF
--- a/src/components/widgets/retract/Retract.vue
+++ b/src/components/widgets/retract/Retract.vue
@@ -83,7 +83,7 @@ export default class Retract extends Mixins(StateMixin) {
   }
 
   get retract_length_max () {
-    if (this.defaults.retract_length <= 0) return 15
+    if (this.defaults.retract_length <= 0) return 10
     return (this.defaults.retract_length * 2).toFixed(2)
   }
 
@@ -110,7 +110,7 @@ export default class Retract extends Mixins(StateMixin) {
   }
 
   get unretract_extra_length_max () {
-    if (this.defaults.unretract_extra_length <= 0) return 15
+    if (this.defaults.unretract_extra_length <= 0) return 10
     return (this.defaults.unretract_extra_length * 2).toFixed(2)
   }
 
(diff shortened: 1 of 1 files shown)


src/components/widgets/retract/Retract.vue
```diff
@@ -84,7 +84,7 @@ export default class Retract extends Mixins(StateMixin) {
 
   get retract_length_max () {
     if (this.defaults.retract_length <= 0) return 15
-    return Math.round(this.defaults.retract_length * 2)
+    return Number((this.defaults.retract_length * 2).toFixed(2))
```
pedrolamas, line 87: Drop (remove) `Number` that should be automatically inferred and not required.

src/components/widgets/retract/Retract.vue
```diff
@@ -111,7 +111,7 @@ export default class Retract extends Mixins(StateMixin) {
 
   get unretract_extra_length_max () {
     if (this.defaults.unretract_extra_length <= 0) return 15
-    return Math.round(this.defaults.unretract_extra_length * 2)
+    return Number((this.defaults.unretract_extra_length * 2).toFixed(2))
```
pedrolamas, line 114: Same here, `Number` is not required and can be safely removed.

pedrolamas (changes requested):

pedrolamas (approved):
LGTM!

Issue #279, opened 2021-05-09
Full Screen webcam feed


PigeonFX:
## Is your feature request related to a problem? Please describe.
(A clear and concise description of what the problem is. Ex. I'm always frustrated when [...])

## Describe the solution you'd like:
Double-clicking on the webcam to go full screen rather than a 640x480 with maybe an emergency stop button.  
## Describe alternatives you've considered:
(A clear and concise description of any alternative solutions or features you've considered. Again, if possible, think about why these alternatives are not working out.)

## Affected users and/or setups:
I monitor my printing on phone and tablet, so would find having a full size screen of the stream very handy and a quick way to trigger emergency stop.
## Additional Context:
(Add any other context or screenshots about the feature request here.)

The current system of being able to bring up the stream smaller than the one already on the screen seems a little redundant.

![Capture](https://user-images.githubusercontent.com/39096995/117564189-c9194880-b0a2-11eb-9c1b-db2c799ac003.PNG)

StuSerious:
hey there @PigeonFX!

this idea has been discussed and already has backend attention!
expect it in future releases!

cheers,
~S

maruhe:
this is the point where I currently (have to) use OctoPrint
(with the 'Fullscreen Plugin') a double-click on the camera-image switches to this view where I have
*) camear in full-window-sice
*) information from printing-widget
*) pause-button (may with extra confirm)
*) full-screen-button to switch between full-window and full-screen

![grafik](https://user-images.githubusercontent.com/49123648/124357226-ad519f00-dc1a-11eb-8c12-2a6ffe6d44e2.png)

pedrolamas:
On the next release we will provide a button to open the video feed in a new (separate) browser window/tab.

That will only have the feed, no buttons to control the print itself, so for that reason we will keep this ticket open to revisit going forward.

PR #498, opened 2022-01-30
Do not merge: Adding full(er) page camera viewing

randellhodges:
Working on adding a full page view of the cameras. Right now you just click the icon in the navbar to go to a "full page" camera view.

The full page needs some work. I'd like to figure out how to get it to expand to the available container without scrolling and scale to fit.

This is for: https://github.com/fluidd-core/fluidd/issues/279

The original request asked about double clicking on a camera. I can probably do that but would need to use CameraItem directly on the Camera page instead of the CameraCard. If I attempt to use the CameraCard, the code might get a little messy.

I'm looking for feedback on
- Exact navigation. Double click only? Navbar? Both? Something else?
- Css and structure (and code) to get the CameraItem to fill the container and scale. With the cols/rows and fill-height and the rest, it is a bit confusing.

Note: How the full page works, that is, going to the Card (with the drop down camera selection) or the Item (directly to the camera viewer) will impact a lot of things, including navigation, so that needs to be decided first.

I am not 100% committed at this point to finishing this depending on the scope of the requested changes. A simple implementation followed by enhancements is my preference. If someone wants to jump in and take this, I have no problems closing out this PR.

Here are some screenshots of how it is working initially:

Multiple Cameras (Dashboard):
image

Full screen, needs some work:
image

Single Camera (Dashboard):
image

Works ok. Needs some css work when resizing the screen smaller
image

### DIFF
--- a/src/components/layout/AppNavDrawer.vue
+++ b/src/components/layout/AppNavDrawer.vue
@@ -42,6 +42,13 @@
           </app-nav-item>
 
           <app-nav-item
+            v-if="hasCameras"
+            icon="$camera"
+            to="/camera">
+            {{ $tc('app.general.title.camera', 2) }}
+          </app-nav-item>
+
+          <app-nav-item
             v-if="supportsHistory"
             icon="$history"
             to="/history">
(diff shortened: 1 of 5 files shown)


randellhodges:
I'm not going to be able to give this the attention it deserves, and it looks like someone else is going to run with it.

PR #500, opened 2022-01-30
feat: idle FPS

### LINKED ISSUE #314: Lower FPS when not focused
Hello there,

When using the webcam feature of Fluidd my browser uses a lot of resources, which is no problem when using it, but when I have it on a second screen when doing something else it's a waste.

My request is a Lower FPS when not focused function to cut down on the resource cost.
So that I can set it on 15FPS on focus and 1FPS on not focus.

matmen:
Implementation of FR "Lower FPS when not focused" (closes #314)
Also fixes broken frame request scheduler (https://github.com/fluidd-core/fluidd/commit/f5ad536ffd8b78f702639812116439acc3d08486)

Example image showing two cameras idling with their idle FPS to 5fps and 1fps:
image

Camera settings:
image

### DIFF
--- a/src/components/widgets/camera/CameraItem.vue
+++ b/src/components/widgets/camera/CameraItem.vue
@@ -141,7 +141,7 @@ export default class CameraItem extends Vue {
       this.camera &&
       this.camera.type === 'mjpgadaptive'
     ) {
-      const fpsTarget = this.camera.fpstarget || 10
+      const fpsTarget = document.hasFocus() ? (this.camera.fpstarget || 10) : (this.camera.fpsidletarget || 1)
       const end_time = performance.now()
       const current_time = end_time - this.start_time
       this.time = (this.time * this.time_smoothing) + (current_time * (1.0 - this.time_smoothing))
--- a/src/components/widgets/camera/CameraItem.vue
+++ b/src/components/widgets/camera/CameraItem.vue
@@ -123,6 +123,7 @@ export default class CameraItem extends Vue {
         const hostUrl = new URL(document.URL)
         const url = new URL(this.cameraUrl, hostUrl.origin)
         url.searchParams.set('cacheBust', this.refresh.toString())
+        this.request_start_time = performance.now()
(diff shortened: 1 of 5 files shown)


randellhodges:
When I read the title, I thought, it would drop the FPS when the printer was idle/not printing. I see it is actually for when the window isn't focused, but I wonder if it could be extended so that it uses the idle fps setting when it isn't printing?

matmen:
I was going to implement that first, too, but wasn't sure because it's out of scope for the FR.
@StuSerious Opinions? I'm happy to implement it (maybe in another PR, tho?), just need feedback on a couple of things:
* Do we just use the idle fps while not printing (and while printing but not in focus)?
* Should this be toggleable?
* Should idle fps while not printing (or while printing?) be another setting? Not sure about this one, having 3 or 4 input fields for FPS in the camera settings may be quite confusing and cluttered.

randellhodges:
I fully support incremental improvements/updates. If you feel implementing an "idle while not printing" type of feature would be better split as a separate update/PR, that sounds good to me.

src/components/widgets/camera/CameraItem.vue
```diff
@@ -123,6 +123,7 @@ export default class CameraItem extends Vue {
         const hostUrl = new URL(document.URL)
         const url = new URL(this.cameraUrl, hostUrl.origin)
         url.searchParams.set('cacheBust', this.refresh.toString())
+        this.request_start_time = performance.now()
```
pedrolamas, line 126: What is this for?

matmen: it's a fix for the broken frame load scheduler. without it the delay between frames will eventually be 0 so it's loading a frame on every tick instead of when it's actually scheduled to load

pedrolamas (approved):
LGTM

matmen:
screenshot update in [camera docs](https://docs.fluidd.xyz/features/cameras) needed

Issue #470, opened 2022-01-16
v1.16.2 When printer is not connected cameras cannot be selected


RobRobM:
I have 2 cameras, one pi camera and one USB webcam, connect and when the printer is offline I cannot switch cameras. I can only use the camera pull down menu when the printer is connected. If the cameras are visible when the printer is online that they will remain visible when the printer goes offline.

## Affected Version(s):
(The versions of fluidd this issue occurs on.)
- klipper version: v0.10.0-221
- moonraker version: v0.7.1-280
- fluidd version: v1.16.2

## Expected Results:
Able to switch cameras anytime.

pedrolamas:
This is now fixed in the `develop` branch and will be available on the next release!

PR #503, opened 2022-02-01
fix: keep camera select menu always enabled

randellhodges:
This leaves the camera select menu enabled even when klipper is disconnected.

For issue: https://github.com/fluidd-core/fluidd/issues/470


matmen (approved):

pedrolamas (approved):
LGTM

Issue #375, opened 2021-07-31
Our telegram-bot-service is hidden from the "restart service" panel.


aka13-404:
Hello,

We have a small [telegram-bot](https://github.com/nlef/moonraker-telegram-bot/) which we use to fetch different info from moonraker. It is configured as all the other stuff around klipper, with config files, and needs to restart to read the config changes.

The bot runs as "moonraker-telegram-bot.service".

Now the problem is, that everything starting with "moonraker" is exempt from the list, see 
https://github.com/cadriel/fluidd/blob/052b16b9fcb9bc3c5bae30893b8c56fdc54dfb75/src/components/common/SystemCommands.vue#L84

We would obviously like to be able to restart the bot from fluidd webinterface.
 Do you have a suggestion, what would be the optimal way to do that? 

## Affected Version(s):
v1.16.2

aka13-404:
This is still an issue, as we are preparing a kiauh release, we need a way to restart the bot from the UI.

Issue #421, opened 2021-09-29
Multiple instances of klipper/moonraker - unable to be restarted via GUI


teikjoon:
## Affected Version(s):
- klipper version: v0.9.1-786-gc2d3a168
- moonraker version: v0.7.1-56-g2447cca
- fluidd version: v1.16.2

## Platform:
Windows 10/Chrome/Ubuntu 20.04

## Hardware Configuration:
BTT SKR1.4 Turbo, Lenovo M73 Tiny

## Reproduction Steps:
1. Tried to restart klipper or moonraker service

![image](https://user-images.githubusercontent.com/42706322/135243764-b8e02c2d-f86d-40f6-b646-9793f0f082a1.png)

## Actual Results:
Fluidd reports error 'unable to restart Klipper'

## Expected Results:
Klipper/moonraker should restart

## Additional information
Have installed multiple instances of klipper/moonraker using KIAUH, so the klipper service is actually 'klipper-1' & 'klipper-2' not 'klipper'. 

This is probably why it can't restart any of the services - also logged an issue with KIAUH dev and this was his response : https://github.com/th33xitus/kiauh/issues/37#issuecomment-928374965

randellhodges:
In my opinion, if I'm connected to Printer-1, I wouldn't want to restart a klipper or moonraker instance for Printer-2. Things like checking if it is currently printing, etc wouldn't really be possible.

It would be nice if Moonraker would filter or return the instance of the moonraker you are currently connected to, as well as the service name of the klipper instance that moonraker instance is connect to.

Then you'd still just have klipper/moonraker listed in the services and it would be the one that you currently have a websocket connect to.

It would work as a facade, translating klipper or moonraker to klipper-X or moonraker-X where -X is the instance.

randellhodges:
You can ignore about 1/2 of what I previously said.

I'm been playing with some service enhancements. It seems this issue is specific to the klipper instances? Moonraker will always restart the instance you're connect to?

So, a couple ways to handle this.

1) make the name of the klipper service configurable, default to klipper.
2) get a change made to moonraker so that it has an endpoint like `/server/restart` but instead of stopping the web server service, the 3 services (start/stop/restart) would signal klipper to do the actions?

I think option 2 might not be well received.

pedrolamas:
> It would be nice if Moonraker would filter or return the instance of the moonraker you are currently connected to, as well as the service name of the klipper instance that moonraker instance is connect to.

[...]

PR #504, opened 2022-02-02
feat: Expanded Service State

randellhodges:
Displays all services returned by moonraker and provides the ability to start/stop/restart those services.

The sidebar width has been increased from 256px to 300px so there is more room to display long service names.

image

This should resolve: https://github.com/fluidd-core/fluidd/issues/375

I am thinking of following this PR up with one that lets you define the service name for klipper for https://github.com/fluidd-core/fluidd/issues/421


pedrolamas:
I do like this, but I do wonder in what situation would one want to stop a service anyway? 🤔

pedrolamas:
What do you think about making the round buttons and icons small, matching the style for the "delete printer" button just below?

image

src/mixins/services.ts
```diff
@@ -22,59 +35,69 @@ export default class ServicesMixin extends Vue {
...
+
+  /**
+   * Start a service by name.
+   */
+  async serviceStartByName (name: string) {
+    if (name === 'klipper') {
+      SocketActions.machineServicesStart(name)
+    } else {
+      SocketActions.machineServicesStart(name)
+    }
```
pedrolamas, line 71: There's no difference here for "klipper" or not!

matmen:
> What do you think about making the round buttons and icons small, matching the style for the "delete printer" button just below?


Using just the `small` attribute on the v-icon (like the delete icon does) doesn't look great IMO:

v-icon small



The `v-icon` API also has a `dense` attribute, but it made no difference for me. Manually setting the `dense` scaling using
`style="transform: scale(calc(20/24))"` looks like this (but I'm not sure it's the right solution):

Transform Scale

pedrolamas:
> Using just the `small` attribute on the v-icon (like the delete icon does) doesn't look great IMO:

Yep doesn't look good, but I would argue that only the "restart" looks very odd... the "stop" is fine and I checked and "play" is ok too!

I would either keep as @randellhodges put on the PR or find a different icon for "restart", but I have no idea what to use instead!

src/mixins/services.ts
pedrolamas: This is now fixed.

randellhodges:
> > What do you think about making the round buttons and icons small, matching the style for the "delete printer" button just below?
> 
> Using just the `small` attribute on the v-icon (like the delete icon does) doesn't look great IMO:
> 
> v-icon small
> 
> The `v-icon` API also has a `dense` attribute, but it made no difference for me. Manually setting the `dense` scaling using `style="transform: scale(calc(20/24))"` looks like this (but I'm not sure it's the right solution):
> 
> Transform Scale

The scale option definitely looks better than the small, but then it looks like it needs a pixel or 2 of padding at the top to push it down (probably just an illusion due to the arrow at the top of the circle).

I think it is just how mdi does these icons, the restart with the circle around it is just larger than the square stop.

If there is a better icon, color, or something I'm up for that

pedrolamas:
Adding `small` to both the `v-icon` and parent `v-btn` doesn't look **that** bad, I think...

image

I enabled "hover" on the 2 buttons for illustration purposes only

image

pedrolamas:
I made some minor layout changes using mostly the vuetify base components and a bit of styling overrides, I believe it looks better now!

image

pedrolamas (dismissed):
From my tests, this is ready to merge and working fine!

src/components/common/SystemCommands.vue
```diff
@@ -55,40 +55,47 @@
 
     <v-list-group
       prepend-icon="$restart"
-      no-action>
```
pedrolamas, line 58: @randellhodges I just noticed you removed `no-action` from here, I assume this was to get the text to fully align to the left?

That change however, makes it look rather odd when we compare with the items under "Host" and "Power" entries as they still show the margin on the left, so I wonder if making the sidebar width going to 300px wouldn't be a better option:

image

randellhodges: I went back and forth with that. I didn't consider making the sidebar wider. I think that would indeed look better. I was attempting to prevent `moonraker-telegram-bot` from wrapping. I know we won't be able to prevent every name from ever wrapping, but I think your idea of making it a little wider is a better compromise.

pedrolamas: Doing it! 🙂

pedrolamas: Done!

pedrolamas (approved):

matmen (approved):

PR #510, opened 2022-02-03
fix: websocket cache issue (arrayMerge)

Biorn1950:
- DeepMerge was appending the array instead of merging.

Thanks @pedrolamas

Signed-of-by: Biorn <biorn@me.com>


pedrolamas (approved):
LGTM

PR #514, opened 2022-02-07
feat: edit on click

### LINKED ISSUE #312: Edit on click
Not a problem, just a usability enhancement. 99% of the time when clicking on a file in settings it is to edit it. Why not make this default behaviour and move other options into buttons in the edit screen?

Clicking a file takes you into edit, removing an unnecessary click.

Clicking edit.

Everyone that tinkers lots.

## Additional Context:
Loving your work. Many thanks.

matmen:
Closes #312.
Not sure about the file extension list, it probably should be stored more centralized and maybe even user-customizable? Discussion is highly welcome.


src/components/widgets/filesystem/FileSystem.vue
```diff
@@ -437,13 +437,18 @@ export default class FileSystem extends Mixins(StateMixin, FilesMixin, ServicesM
...
         this.selected = []
       } else {
-        // Open the context menu
-        this.contextMenuState.x = e.clientX
-        this.contextMenuState.y = e.clientY
-        this.contextMenuState.file = item
-        this.$nextTick(() => {
-          this.contextMenuState.open = true
-        })
+        if (e.type === 'click' && ['txt', 'md', 'conf', 'cfg', 'ini', 'sh'].includes((item as AppFile).extension)) {
```
pedrolamas, line 440: Please don't use `item as AppFile`!

TypeScript is a lot smarter than JavaScript and has a fantastic feature called [type narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html)!

`AppFile` has been defined as having `type: 'file'`, so all we need here is to check the `type` to make item be of `AppFile`.

Given your code above, just replace it with this:

```typescript
if (item.type === 'file' && e.type === 'click' && ['txt', 'md', 'conf', 'cfg', 'ini', 'sh'].includes(item.extension)) {
  this.handleFileOpenDialog(item)
// rest of stuff
```

src/components/widgets/filesystem/FileSystem.vue
```diff
@@ -437,13 +437,18 @@ export default class FileSystem extends Mixins(StateMixin, FilesMixin, ServicesM
...
-        // Open the context menu
-        this.contextMenuState.x = e.clientX
-        this.contextMenuState.y = e.clientY
-        this.contextMenuState.file = item
-        this.$nextTick(() => {
-          this.contextMenuState.open = true
-        })
+        if (e.type === 'click' && item.type === 'file' && ['txt', 'md', 'conf', 'cfg', 'ini', 'sh'].includes(item.extension)) {
+          // Open the file editor
+          this.handleFileOpenDialog(item as AppFile)
```
pedrolamas, line 442: The `as AppFile` is not needed here, TypeScript should infer that automatically from the above `type === 'file'` check!

pedrolamas:
Just tested it, it works great!

Here's another extension to add: `.log`

matmen:
Not sure about logs; presenting the users with an option to download right away seems nice. But I guess that will still be there in the context menu.

src/components/settings/FileEditorSettings.vue
```diff
@@ -0,0 +1,73 @@
...
+<script lang="ts">
+import { Component, Mixins } from 'vue-property-decorator'
+import StateMixin from '@/mixins/state'
+
+@Component({
+  components: {}
+})
+export default class FileEditorSettings extends Mixins(StateMixin) {
+  get confirmDirtyEditorClose (): boolean {
+    return this.$store.state.config.uiSettings.editor.confirmDirtyClose
```
pedrolamas, line 47: Currently, the "confirm on close feature is broken (doesn't save the state on the settings), I believe this is because this should be `confirmDirtyEditorClose` and not `confirmDirtyClose`

pedrolamas (changes requested):
AS you made the changes for the `confirmDirtyEditorClose` key, those changes also need to be made in FileEditorDialog.vue (it's still looking for the old setting position!)

pedrolamas (approved):

PR #522, opened 2022-02-16
feat: option to flip console direction

### LINKED ISSUE #511: Console direction

I find it somewhat inconvenient to have the console input field at the bottom of the console. With the default dashboard layout, the console is half visible which means I have to scroll down to type in commands in the console. It is of course possible to modify the layout, but that is in my opinion a bad solution to the problem. 

I would like to flip the console such that the input field is on the top and new output from the console is displayed on top. 

This makes the input field easier accessible and it becomes the first thing you see in the console window. Even if the console is partially visible on the dashboard it is still possible to see new output from the console and type in commands.

One possible downside of this is, if autoscroll in the console is enabled, it is easier to accidentally scroll the console window instead of the main dashboard window. 

_No response_

### Additional information

I think mainsail is using a layout similar to this.

matmen:
Adds an option to move the console input field to the top of the console widget.
Not sure if the log direction should be flipped too (most recent at top), and if "Flip layout" is the right wording here.
Closes #511

### DIFF
--- a/src/components/widgets/console/Console.vue
+++ b/src/components/widgets/console/Console.vue
@@ -1,6 +1,12 @@
 <template>
   <div
     class="console">
+    <console-command
+      v-if="!readonly && flipLayout"
+      v-model="consoleCommand"
+      @send="sendCommand"
+    >
+    </console-command>
     <v-card
       flat
       class="console-wrapper"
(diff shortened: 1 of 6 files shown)


ReinBentdal:
Nice! I won't be able to test (Holliday) but i would prefer for the content of the terminal to also flip, like you described. I think this would make it more coherent and more intuitive

pedrolamas (approved):
LGTM!

PR #523, opened 2022-02-16
feat: show qr-code with api key

### LINKED ISSUE #411: FR: QR-Code for API Key
Hey, I am the creator of the Mobileraker Flutter App.
If a user enables authorization my app only works with the API Key of Moonraker.
Currently, users have to enter/copy it by hand but I think it would be quite cool to enable users to just scan a QR code containing the API-Key.
So would it be possible to add a QR-Code button that shows the API-Key as QR-Code?

Cheers.

pedrolamas:
Fixes #411 

Adds new dependency to [qrcode.vue](https://www.npmjs.com/package/qrcode.vue)

image

### DIFF
--- a/src/components/ui/AppQrCode.vue
+++ b/src/components/ui/AppQrCode.vue
@@ -2,7 +2,7 @@
   <qrcode-vue
     :value="value"
     :size="size"
-    level="H"
+    level="L"
   ></qrcode-vue>
 </template>
 
@@ -19,7 +19,17 @@ export default class AppQrCode extends Vue {
   @Prop({ type: String, default: '' })
   value!: string;
 
-  @Prop({ type: Number, default: 300 })
+  @Prop({ type: Number, default: 260 })
   size!: number;
 }
 </script>
+
+<style lang="scss" scoped>
+  ::v-deep {
+    canvas, svg {
+      display: block;
+      padding: 10px;
+      background: #fff;
+    }
+  }
+</style>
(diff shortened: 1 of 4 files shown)


src/components/ui/AppQrCode.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+  <qrcode-vue
+    :value="value"
+    :size="size"
+    level="H"
```
matmen, line 5: An error correction of `L` probably suffices here, considering this will be displayed on a monitor and not printed out where it could be damaged. IMO it makes the QR code look a bit nicer, but I'm not sure it matters

src/components/ui/AppQrCode.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+  <qrcode-vue
```
matmen, line 2: A (white) border around the QR code would be nice, my phone failed to pick it up because the position marks had too low of a contrast compared to the background. The qrcode.vue docs mention a [`margin` property](https://github.com/scopewu/qrcode.vue#margin), but it seemed to not work in my testing. Giving it an 8px CSS padding with a background-color of white seemed to work fine.

matmen (changes requested):

pedrolamas:
@matmen thanks for reviewing, I've updated the PR following up your on comments (also updated the main image on the PR description to show the latest changes)

Clon1998:
Greate work!
Thanks, this will make setting up a printer in my app way easier :)

PR #524, opened 2022-02-18
fix: improves macro param default value handling

### LINKED ISSUE #376: Macro parameter inputs aren't populated if floats are used.
fluidd version: v1.16.2-0fa3c9b

If you have a macro defined with any decimal/float parameter defaults fluidd doesn't detect those default values correctly. It just leaves those input fields empty (= it displays the parameter name in the input fields [which also is confusing at first, you could think it passes the parameter name itself as a value until you click into the input field]).

## Reproduction Steps:
Define any macro with parameters with decimal number defaults.

In my case:
`[gcode_macro PA_TOWER_SETUP]`
`gcode:`
`    {% set S = params.S|default(0.2)|float %} ; start`
`    {% set F = params.F|default(0.02)|float %} ; factor`
`    {% set B = params.B|default(0)|float %} ; band`
`    SET_VELOCITY_LIMIT SQUARE_CORNER_VELOCITY=1 ACCEL=500`
`    TUNING_TOWER COMMAND=SET_PRESSURE_ADVANCE PARAMETER=ADVANCE START={S} FACTOR={F} BAND={B}`

## Actual Results:
In fluidd the inputs for S and F, which have decimal defaults, are left empty, the B input is correctly populated with 0.

## Screenshot(s):
Unbenannt

pedrolamas:
Improves the regular expression used to detect macro parameter defaults. This should be a bit more permissive with white-space and multiple params on the Jinja2 `default` function.

Tested with the following example:

```txt
[gcode_macro TEST]
gcode:
  {% set A = params.A | default(0, true) | float %}
  {% set B = params.B | int | default(0.005, true) | float %}
  {% set C = params.C | default("PLA", true) %}
  {% set D = params.D | default("PLA") %}
  {% set E = params.E | int | default(0.2) %}
```

The above example macro will show like this:

image

Fixes #376 

### DIFF
--- a/src/components/ui/AppMacroBtn.vue
+++ b/src/components/ui/AppMacroBtn.vue
@@ -124,12 +124,12 @@ export default class AppMacroBtn extends Mixins(StateMixin) {
   mounted () {
     if (!this.macro.config || !this.macro.config.gcode) return []
     if (this.macro.config.gcode) {
-      const regex = /params\.(\w*)\|?(default\('?(\w*)'?\))?/gmi
+      const regex = /params\.(\w+).*\|\s*default\s*\(\s*([^,)]+)/gmi
       let match = regex.exec(this.macro.config.gcode)
       do {
         if (match && match[1]) {
           const name = match[1]
-          const value = match[3] || ''
+          const value = (match[2] || '').trim()
           if (!this.params[name]) {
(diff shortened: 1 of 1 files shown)


matmen (approved):
LGTM

PR #525, opened 2022-02-18
feat: Add full-screen button to camera view

matthewlloyd:
Signed-off-by: Matthew Lloyd <github@matthewlloyd.net>

### DIFF
--- a/src/components/widgets/camera/CameraItem.vue
+++ b/src/components/widgets/camera/CameraItem.vue
@@ -43,6 +43,14 @@
       >
         fps: {{ currentFPS }}
       </div>
+      <div
+        v-if="cameraFullScreenUrl"
+        class="camera-fullscreen"
+      >
+        <a :href="cameraFullScreenUrl" target="_blank">
+          <v-icon>$fullScreen</v-icon>
+        </a>
+      </div>
     </v-sheet>
   </div>
 </template>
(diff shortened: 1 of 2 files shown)


matthewlloyd:
Your call - I can always continue rebasing this PR on my clone if you prefer the in-app approach. Personally I like opening a separate tab because:

- I get better camera FPS when the video isn't being rendered in an <iframe>, maybe because the rest of the Fluidd UI takes up browser rendering cycles. I get about 10-15 FPS with a dedicated tab, vs about 2-3 FPS in Fluidd. 
- With a dedicated tab in Chrome, the webcam fills my entire browser window without any border. I'm running a 1080P camera and I like to see the detail without having to get up close to my monitor.
- Opening a new tab in Chrome is snappier (almost instantaneous in fact) compared to clicking the side buttons in Fluidd and waiting for the app-nav-item to render.
- Having a separate tab keeps the regular Fluidd UI loaded and quickly accessible without having to wait for the Fluidd Home tab to load if I switch back, which takes about 1-2 seconds on my machine.

I can see the advantage of keeping an E-stop button available, though I find in practice on the rare occasion I need to do that while viewing the full-screen camera tab, I can just Ctrl+Shift+Tab back to Fluidd to find it.

These approaches aren't mutually exclusive, of course... one could have both the full-screen side bar version available, and also the full-screen button. I think it's a frequent enough action to warrant providing more than one method.

pedrolamas (approved):

matmen (approved):

PR #527, opened 2022-02-19
feat: TURN_OFF_HEATERS confirmation in temperature presets

matmen:
As requested by NewCalJs in the Discord server.
Adds a confirmation modal to the `Heaters off` button under temperature presets when the printer is busy.
image

### DIFF
--- a/src/components/widgets/thermals/TemperatureCard.vue
+++ b/src/components/widgets/thermals/TemperatureCard.vue
@@ -166,7 +166,18 @@ export default class TemperatureCard extends Mixins(StateMixin) {
     }
   }
 
-  handleApplyOff () {
+  async handleApplyOff () {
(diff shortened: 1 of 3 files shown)


src/components/widgets/thermals/TemperatureCard.vue
```diff
@@ -166,7 +166,18 @@ export default class TemperatureCard extends Mixins(StateMixin) {
     }
   }
 
-  handleApplyOff () {
+  async handleApplyOff () {
+    if (this.$store.getters['printer/getPrinterState'] === 'busy') {
```
pedrolamas, line 170: Should we also check for paused state?

matmen: Yep, missed some states. Thanks!

pedrolamas (approved):
Tested, working fine!!

PR #533, opened 2022-02-22
chore: adds commit validation via husky

pedrolamas:
After cloning and running `npm install`, one can optionally use `npm run prepare` to add some git hooks to validate if the staged files are linted and that the commit message follows conventional commits standards and has a Signed-off-by line.

**This is opt-in only**, but we can revisit this at a later stage and make it pre-install via `prepare` script if we decide so.


package.json
```diff
@@ -12,6 +12,7 @@
     "test:unit": "vue-cli-service test:unit",
     "test:e2e": "vue-cli-service test:e2e",
     "lint": "vue-cli-service lint",
+    "prepare": "husky install",
```
matmen, line 15: husky already runs `husky install` when it's installed via `npm i`

CONTRIBUTING.md
```diff
@@ -18,3 +18,5 @@ Fluidd is built on VueJS, using TypeScript.
 
 - By signing off on commits, you acknowledge that you agree to the [developer certificate of origin](/developer-certificate-of-origin).
 This must contain your real name and a current email address.
+
+After cloning the repo and running `npm install`, we recommend running `npm run prepare` to install a couple of git hooks that will pre-validate all commits from that moment.
```
matmen, line 22: see `package.json` comment

package.json
pedrolamas: It did do that, but they stopped doing it on the last major release. For a while, they recommended using the npm `post-install` action to do it automatically, but even that has now been discouraged as it is considered an anti-pattern by npm.

CONTRIBUTING.md
pedrolamas: Replied above! 🙂

package.json
matmen: alright sounds good then! 👍

matmen (dismissed):

package.json
pedrolamas: I might have spoken too soon there... I was under the impression `prepare` script was not running automatically, but I just removed my "node_modules" folders and when I did `npm i` to get it back, I got the same message that on your screenshot!

pedrolamas: I moved from `prepare` to `bootstrap` to ensure it is an opt-in only feature!

matmen (approved):

PR #535, opened 2022-02-22
fix: service state item layout balance

pedrolamas:
This is minor issue, but it was bugging me and I decided to fix it anyway...

With #504 we now have the possibility to start/restart/stop services, but as vuetify doesn't support multiple horizontal buttons on a single list item, we had to style it ourselves.

There was a minor issue with margins and item alignment that was causing the list item layout to become unbalanced when compared to the rest of the list.

image

On the above image, the issue is mostly noticeable on the space between "Services" label and "Klipper" just below it, but with more services it becomes even more prominent

This PR fixes this minor issue.


pecirep (approved):
LGTM

PR #538, opened 2022-02-24
feat: show pressure and humidity from sensors

### LINKED ISSUE #335: BME280
Hi. 
Is possible to show data from BME280 some where in temperature window (temperature, humidity, barometric pressure) without using macro in terminal?

Thanks in advance.

### LINKED ISSUE #397: Target Temperatures for other temperature sensors
Would it be possible to show the target temperatures for other temperatures, especially the chamber ?

A read only field would be fine, my chamber target temp is passed from the slicer to the start_print macro, and it would nice to be able to see what the target temp is as I wait for the chamber to reach a temp before continuing

Many Thanks

### LINKED ISSUE #434: Display HTU21D humidity sensor data in the dashboard and graph
Yes and no. The problem is that Fluidd does not display the humidity of a HTU21D sensor. It only displays the temperature.

The HTU21D sensors to show the current humidity level in the dashboard. Secondarily, it would be great if the the Humidity could be added to the graph/chart. 

Mainsail does show the humidity in the dashbaord.

Everyone using HTU21D sensors with Klipper

## Additional Context:
NA

pedrolamas:
Here's how it is currently showing the humidity output from a BME280 connected to a Pi Zero 2W running as MCU:

image

Fixes #335, fixes #397, fixes #434

### DIFF
--- a/src/components/widgets/thermals/TemperatureTargets.vue
+++ b/src/components/widgets/thermals/TemperatureTargets.vue
@@ -187,11 +187,9 @@
                 <span
                   v-bind="attrs"
                   v-on="on"
-                >{{ item.temperature.toFixed(1) }}
-                  <small>°C
-                    <span v-if="item.humidity"><br>{{ item.humidity.toFixed(1) }}&nbsp;%</span>
-                    <span v-if="item.pressure"><br>{{ item.pressure.toFixed(1) }}&nbsp;hpa</span>
-                  </small>
+                >{{ item.temperature.toFixed(1) }}<small>°C</small>
+                  <small v-if="item.humidity"><br>{{ item.humidity.toFixed(1) }}&nbsp;%</small>
+                  <small v-if="item.pressure"><br>{{ item.pressure.toFixed(1) }}&nbsp;hpa</small>
                 </span>
               </template>
               <span v-if="item.measured_max_temp && item.measured_min_temp">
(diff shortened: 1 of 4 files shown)


src/components/widgets/thermals/TemperatureTargets.vue
```diff
@@ -187,7 +187,12 @@
                 <span
                   v-bind="attrs"
                   v-on="on"
-                >{{ item.temperature.toFixed(1) }}<small>°C</small></span>
+                >{{ item.temperature.toFixed(1) }}
+                  <small>°C
```
matmen, line 191: The newline before this introduces an issue where the `°C` label wraps to a new line (because newlines in HTML act like spaces that turn into wrap-points):
image

I would suggest creating 2 `<small>` elements (or keeping the `°C` on the same line as the temperature):
```vue
<span
   v-bind="attrs"
  v-on="on"
>{{ item.temperature.toFixed(1) }}<small>°C</small>
  <small>
    <span><br>{{ (item.humidity || 1).toFixed(1) }}&nbsp;%</span>
    <span><br>{{ (item.pressure || 1).toFixed(1) }}&nbsp;hpa</span>
  </small>
</span>
```

matmen (changes requested):
I can't test it properly because I don't have access to any of these sensors, but I did find one small issue while testing:

src/components/widgets/thermals/TemperatureTargets.vue
pedrolamas: Nice catch! I did some changes that should now fix this issue!

matmen (approved):

PR #543, opened 2022-02-27
fix: add missing default for idle fps input

matmen:
Adds the missing type and default value for the "FPS Target when not in focus" input field in the "Add Camera" dialog


pedrolamas (approved):
LGTM

PR #544, opened 2022-02-28
fix: filename wrap in status window

samwiseg0:
This issue has bothered me for some time and I figured I would attempt to fix it. If this is not the "best" way to do this please let me know. Tested in several browsers and iOS 15.

**Before the changes:**
Screen Shot 2022-02-27 at 8 07 15 PM
Screen Shot 2022-02-27 at 8 07 29 PM

**After the changes:**
Screen Shot 2022-02-27 at 8 04 53 PM
Screen Shot 2022-02-27 at 8 04 39 PM

Signed-off-by: Phillip Hernandez samwiseg00@gmail.com

### DIFF
--- a/src/components/widgets/status/StatusLabel.vue
+++ b/src/components/widgets/status/StatusLabel.vue
@@ -30,13 +30,10 @@ export default class AppSwitch extends Mixins(StateMixin) {
   }
 
   span.value {
-    // flex: 0 0 auto;
-    // min-width: 0;
     > span {
       display: block;
-      // white-space: nowrap;
       overflow: hidden;
-      word-break: break-all;
+      text-overflow: ellipsis;
     }
   }
 </style>
--- a/src/components/widgets/status/StatusLabel.vue
+++ b/src/components/widgets/status/StatusLabel.vue
@@ -37,7 +37,6 @@ export default class AppSwitch extends Mixins(StateMixin) {
       // white-space: nowrap;
       overflow: hidden;
       word-break: break-all;
-      width: auto;
     }
   }
 </style>
--- a/src/components/widgets/status/StatusLabel.vue
+++ b/src/components/widgets/status/StatusLabel.vue
@@ -35,7 +35,6 @@ export default class AppSwitch extends Mixins(StateMixin) {
     > span {
       display: block;
       // white-space: nowrap;
-      text-overflow: ellipsis;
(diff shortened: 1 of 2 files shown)


src/components/widgets/status/StatusLabel.vue
```diff
@@ -35,9 +35,9 @@ export default class AppSwitch extends Mixins(StateMixin) {
     > span {
       display: block;
       // white-space: nowrap;
-      text-overflow: ellipsis;
       overflow: hidden;
-      text-overflow: ellipsis;
+      word-break: break-all;
+      width: auto;
```
pedrolamas, line 40: I don't see this entry actually doing anything, what is the purpose of using `width: auto` here?

samwiseg0: You are correct. That was left over from my testing. I have removed it.

pedrolamas:
Though I agree with you that the initial approach is quite flawed (and quite honestly, doesn't even seem to be doing a proper job), I am checking on the implications of such a change on a component that is used for more things than just displaying filenames!

Specifically, if there is an `M117` with a long text, it would also be truncated on letter instead of word, but that would look quite odd for that case... so a localized change for the filename only might make more sense here.

src/components/widgets/status/StatusLabel.vue
pedrolamas: I was referring just to the `width` attribute, but I do agree that `text-overflow` is not working correctly here.

samwiseg0:
> Though I agree with you that the initial approach is quite flawed (and quite honestly, doesn't even seem to be doing a proper job), I am checking on the implications of such a change on a component that is used for more things than just displaying filenames!
> 
> 
> 
> Specifically, if there is an `M117` with a long text, it would also be truncated on letter instead of word, but that would look quite odd for that case... so a localized change for the filename only might make more sense here.

That makes sense. I had not considered M117 as I do not use it. 

I am no css expert by any means and unsure of what would need to be done to make that specific to just the file name.

pedrolamas:
Not only M117 apparently...

image

> I am no css expert by any means and unsure of what would need to be done to make that specific to just the file name.

I think I can help with that! 🙂

pedrolamas (approved):
LGTM

matmen (approved):

PR #547, opened 2022-02-28
fix: z-offset apply

### LINKED ISSUE #546: Save Z_Offset attempts to apply new values using the incorrect gcode command
### What happened

When saving the z offset from the UI, it attempts to apply new offset using the incorrect gcode command. 

In this case my printer has both a z endstop and a probe. The probe is not used to home z instead the nozzle is on the z endstop. 

In this scenario the new offset should be applied to the endstop and not probe. 

Code:
https://github.com/fluidd-core/fluidd/blob/0cf6c87cc661edfc533d62d009ad43a31b2e1886/src/components/widgets/toolhead/ZHeightAdjust.vue#L105-L120

Fluidd should issue the correct gcode command to set the new z offset. 

### How to reproduce

1) Configure klipper to have a z enstop and a probe. 
2) Attempt to save a new offset
3) Observe that it issues `Z_OFFSET_APPLY_PROBE` instead of `Z_OFFSET_APPLY_ENDSTOP`

### Additional information

More details were discussed in discord. They can be found here. 
https://discord.com/channels/762863525818662974/771936287300911185/947836576480759828

Biorn1950:
Changing the condition of the probe check.
fix #546

Signed-of-by: Biorn1950 <biorn@me.com>

### DIFF
--- a/src/components/widgets/toolhead/ZHeightAdjust.vue
+++ b/src/components/widgets/toolhead/ZHeightAdjust.vue
@@ -106,15 +106,11 @@ export default class ZHeightAdjust extends Mixins(StateMixin) {
     return this.$store.getters['printer/getPrinterSettings']()
   }
 
-  get printerHasProbe (): boolean {
-    return 'probe' in this.printerSettings || 'bltouch' in this.printerSettings
-  }
-
   handleZApplyDialog () {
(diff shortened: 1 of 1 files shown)


pedrolamas (approved):
LGTM

matmen (approved):

Issue #453, opened 2021-11-29
Chart with sensors is not displayed. Monitoring tools shows that the Chart SVG is updating, but SVG is not drawing


DizzyaKZ:
## Affected Version(s):
(The versions of fluidd this issue occurs on.)
- klipper version: v0.10.0-167-g2b7d0bba
- moonraker version: v0.7.1-194-geda5623
- fluidd version: v1.16.2

## Hardware Configuration:
MCU v0.10.0-122-gc179db3d

## Reproduction Steps:
1. Installed the Mainsail
2. Changed in the Nginx the link for sites-enabled from fluidd to mainsail
3. Restore the link to fluidd (doesn't like mainsail)

## Actual Results:
The chart disappeared while the table part with sensors still exists and updating. The pressing CHART ON/OFF button is working but chart is not displayed while the container is appearing/disappearing. Have tried it in different browsers Edge,

## Expected Results:
The chart suppose to be displayed

## Log Files:
(The log files fluidd provides at Configure > System Control.)
- klipper log:
.....
'webhooks: registering remote method 'shutdown_machine' for connection id: 3041919864
webhooks: registering remote method 'reboot_machine' for connection id: 3041919864
webhooks: registering remote method 'pause_job_queue' for connection id: 3041919864
webhooks: registering remote method 'start_job_queue' for connection id: 3041919864
Stats 1133.8: gcodein=0 mcu: mcu_awake=0.000 mcu_task_avg=0.000000 mcu_task_stddev=0.000000 bytes_write=3856 bytes_read=4110 bytes_retransmit=0 bytes_invalid=0 send_seq=168 receive_seq=168 retransmit_seq=0 srtt=0.004 rttvar=0.002 rto=0.025 ready_bytes=0 stalled_bytes=0 freq=72004070  heater_bed: target=0 temp=0.0 pwm=0.000 motherboard_temp: temp=0.0 raspberry_pi: temp=48.2 sysload=0.12 cputime=27.165 memavail=3655984 print_time=0.001 buffer_time=0.000 print_stall=0 extruder: target=0 temp=0.0 pwm=0.000
[...]

PR #551, opened 2022-03-01
fix: attempt to fix temp chart issues

pedrolamas:
Related with #453


src/main.ts
```diff
@@ -33,11 +33,10 @@ import { FiltersPlugin } from './plugins/filters'
 import { SocketPlugin } from './plugins/socketClient'
 import { ColorSetPlugin } from './plugins/colorSet'
 import { DayJSPlugin } from './plugins/dayjs'
-import { plugin } from 'echarts-for-vue'
 
 // Import ECharts
-// import * as echarts from 'echarts'
-import * as echarts from 'echarts/core'
+import ECharts from 'vue-echarts'
+import { use } from 'echarts/core'
```
matmen, line 39:
```suggestion
import { use as echartsComponents } from 'echarts/core'
```

src/main.ts
```diff
@@ -70,7 +68,7 @@ Vue.component('vue-headful', vueHeadful)
 // Use any Plugins
 
 // Configure echarts
-echarts.use([
+use([
```
matmen, line 71:
```suggestion
echartsComponents([
```

matmen (commented):
LGTM, but maybe change these so the code is a little more descriptive?

src/main.ts
pedrolamas: renamed to `echartsUse`

matmen (approved):

PR #553, opened 2022-03-02
fix: idle fps behavior

matmen:
Fixes some user complaints on the Discord:
* idle fps input is no longer required
* idle fps default to fps target to restore previous behavior


pedrolamas (approved):
LGTM

PR #554, opened 2022-03-02
docs: moonraker config refactor

majcn:
## Description

Moonraker changed configuration structure some time ago. fluidd was only using `server` part which is now in `data_store` and `file_manager`,...

I removed code part of changing things which are solved https://github.com/fluidd-core/fluidd/pull/555 and only left docs part


matmen (commented):
LGTM, but maybe the corresponding section headers should be added to the examples? e.g.:
```yaml
[file_manager]
log_path: ~/klipper_logs
```

pedrolamas (approved):
LGTM

PR #560, opened 2022-03-04
chore: removes deprecated vue-headful dependency

pedrolamas:
[vue-headful](https://www.npmjs.com/package/vue-headful) has been marked as deprecated with a note to use [vue-meta](https://www.npmjs.com/package/vue-meta), which we already do anyway!

So I've completely removed the deprecated vue-headful dependency and ensured that we only use the correct vue-meta.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,12 +1,6 @@
 <template>
   <v-app v-if="loading"></v-app>
   <v-app v-else class="fluidd">
-    <vue-headful
-      :title="pageTitle"
-      :head="pageIcon"
-    >
-    </vue-headful>
-
     <app-tools-drawer v-model="toolsdrawer"></app-tools-drawer>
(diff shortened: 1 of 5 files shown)


src/App.vue
```diff
@@ -190,6 +201,16 @@ export default class App extends Mixins(StateMixin) {
       this.flashMessage.timeout = (payload && payload.timeout !== undefined) ? payload.timeout : undefined
       this.flashMessage.open = true
     })
+
+    const legacyFavIcons = document.querySelectorAll("link[rel*='icon'][type='image/png']")
```
pedrolamas, line 205: This bit here is kind of a "hack" that will remove the existing `link rel="icon"...` that is put in by the PWA stuff. We would end up with 2 entries and the icon would not update, so this will remove the ones from the generated index.html.

matmen (approved):
LGTM

matmen (approved):

PR #567, opened 2022-03-09
chore: dependency updates

pedrolamas:
Full dependency updates to latest versions, where possible!

- started by upgrading the vue-cli* dependencies to latest versions
- the upgraded remaining packages to versions that are inside the requirements of vue-cli
- ran all possible `npm run *` commands we have, they all work
- regression tested g-code viewer and code editor, both working
- typescript has been upgraded to v4
- eslint 7 now in use (we can't go for v8 as @vue/eslint-config-standard doesn't support it -- yet!)
- build works, tests pass
- some dependencies also moved from prod to dev as they are only used there
- the only npm warnings we get are from potato, we will review that at a later time

**Note:** I recommend starting fresh by deleting existing "node_modules" folder and running `npm install` before doing any tests on this.


src/util/gcode-preview.ts
```diff
@@ -1,5 +1,6 @@
 import { ArcMove, Move, Point, Rotation } from '@/store/gcodePreview/types'
 
+// eslint-disable-next-line @typescript-eslint/ban-types
```
matmen, line 3: Instead replace `Function` with `(item: any, index: number, array: any[]) => number`?

matmen (changes requested):

src/util/gcode-preview.ts
pedrolamas: Done, and made it generic so one doesn't need to use `any`!

matmen (approved):

PR #572, opened 2022-03-12
fix: individual object layer count

### LINKED ISSUE #501: Layer count is wrong when printing multiple object with "complete individual objects" enabled in Super Slicer
### What happened

I have printed multiple projects from Super Slicer today with 2 or more object completed individually and the layer count is wrong for all but one object.
layercount
.

See the layer count of each individual object or the total number of layers of all object combined.

### How to reproduce

Slice multiple objects in Super Slicer with "complete individual objects" enabled, print and watch the layer count.

### Additional information

_No response_

matmen:
Fixes #501


pedrolamas (approved):
LGTM

PR #585, opened 2022-03-17
feat(i18n-JA): adds Japanese language

Suisyan:
Add new Japanese translation.

Signed-off-by: Suisyan <99475190+Suisyan@users.noreply.github.com>

PR #586, opened 2022-03-17
update index.ts

Suisyan:
Add code:ja

Signed-off-by: Suisyan <99475190+Suisyan@users.noreply.github.com>


pedrolamas:
These changes have been merged to #585, so this is no more required.

PR #591, opened 2022-03-18
feat: adds delta support to gcode viewer

### LINKED ISSUE #519: Gcode live view not working for circular bed (delta printer)
### What happened

Gcode live view is empty when using on detla printer but it is working good on rectangular bed. I guess it is because of round bed have origin at 0;0
image

Gcode live view is visible when using round bed

### How to reproduce

Start any print on delta printer and load current file to gcode live view

### Additional information

_No response_

susisstrolch:
Fixes #519

Signed-off-by: Juergen Orschiedt <juorschiedt@gmail.com>

### DIFF
--- a/src/components/widgets/gcode-preview/GcodePreview.vue
+++ b/src/components/widgets/gcode-preview/GcodePreview.vue
@@ -71,11 +71,28 @@
         </svg>
       </defs>
       <g :transform="flipTransform">
+        <clipPath id="clipCircle">
+          <circle
+            :r="bedSize.x.max"
+            cx="0"
+            cy="0"
+          />
+        </clipPath>
         <g
           v-if="drawBackground"
           id="background"
         >
           <rect
+            v-if="isDelta"
+            :height="bedSize.y.max - bedSize.y.min"
+            :width="bedSize.x.max - bedSize.x.min"
+            fill="url(#backgroundPattern)"
+            clip-path="url(#clipCircle)"
+            :x="bedSize.x.min"
+            :y="bedSize.y.min"
+          />
+          <rect
+            v-else
             :height="bedSize.y.max - bedSize.y.min"
             :width="bedSize.x.max - bedSize.x.min"
             fill="url(#backgroundPattern)"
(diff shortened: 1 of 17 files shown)


src/components/widgets/gcode-preview/GcodePreview.vue
```diff
@@ -211,6 +228,14 @@ export default class GcodePreview extends Mixins(StateMixin) {
 
   panning = false
 
+  get isDelta (): boolean {
+    return this.$store.getters['printer/getPrinterSettings']('printer.kinematics') === 'delta'
```
pedrolamas, line 232: I believe your changes should work with both `delta` and `rotary_delta`, so can you make sure you check for both?

pedrolamas (changes requested):

susisstrolch:
Even I've added a test for "rotary_delta" I'm not sure it will work out.
I simply don't know how to compute the bedsize for a rotary.
So I'd suggest to omit the "rotary_delta" for the moment until we get a clear description how the bedsize (radius) is computed.

pedrolamas:
> So I'd suggest to omit the "rotary_delta" for the moment until we get a clear description how the bedsize (radius) is computed.

That is fair and makes sense! Let us then wait for someone with such a printer to do that bit!

pedrolamas (dismissed):
LGTM

susisstrolch:
The current implementation doesn't hurt so much, because print radius will be preset with 100.0 for rotary_delta.
However we could add an additional printer_radius setting for those who want to have an exact picture.

pedrolamas:
> The current implementation doesn't hurt so much, because print radius will be preset with 100.0 for rotary_delta. However we could add an additional printer_radius setting for those who want to have an exact picture.

My personal knowledge on how a delta works is quite limited so I can't evaluate exactly what this involves.

If this works with your delta, I'm happy to merge the changes, as code-wise, this looks perfect to me and I've regression tested this to make sure everything is fine!

susisstrolch:
Ok, so let's go on and let's wait for the rotary_delta guys...

src/components/widgets/gcode-preview/GcodePreview.vue
```diff
@@ -271,6 +297,9 @@ export default class GcodePreview extends Mixins(StateMixin) {
       this.flipY ? -(y.max - y.min) : 0
     ]
 
+    if (this.isDelta) {
```
matmen, line 300: This branch could be moved before `const transform = ...` to save a memory allocation for an unused reference

matmen (approved):
LGTM overall, just one tiny thing I would change ;)

pedrolamas (approved):

matmen (approved):
LGTM, just waiting for a delta user to chime in..

PR #597, opened 2022-03-19
chore: split FileEditor into a separate chunk

Piezoid:
This allows the Monaco editor component to be asynchronously loaded when needed.

The vendor chunk is reduced from 5642.22 KiB  to 2953.59 KiB.
The new chunk weights 2716.54 KiB.

The size could be reduced by shaving off editor features.
Available features are `'accessibilityHelp' | 'anchorSelect' | 'bracketMatching' | 'caretOperations' | 'clipboard' | 'codeAction' | 'codelens' | 'colorPicker' | 'comment' | 'contextmenu' | 'coreCommands' | 'cursorUndo' | 'dnd' | 'documentSymbols' | 'find' | 'folding' | 'fontZoom' | 'format' | 'gotoError' | 'gotoLine' | 'gotoSymbol' | 'hover' | 'iPadShowKeyboard' | 'inPlaceReplace' | 'indentation' | 'inlayHints' | 'inlineCompletions' | 'inspectTokens' | 'lineSelection' | 'linesOperations' | 'linkedEditing' | 'links' | 'multicursor' | 'parameterHints' | 'quickCommand' | 'quickHelp' | 'quickOutline' | 'referenceSearch' | 'rename' | 'smartSelect' | 'snippet' | 'suggest' | 'toggleHighContrast' | 'toggleTabFocusMode' | 'tokenization' | 'unicodeHighlighter' | 'unusualLineTerminators' | 'viewportSemanticTokens' | 'wordHighlighter' | 'wordOperations' | 'wordPartOperations'`
Curently disabled features:  `['!contextmenu', '!snippets', '!multicursor']`

Signed-off-by: Maël Kerbiriou <m431.kerbiriou@gmail.com>

### DIFF
--- a/src/components/widgets/filesystem/FileEditor.vue
+++ b/src/components/widgets/filesystem/FileEditor.vue
@@ -2,16 +2,24 @@
   <div
     ref="monaco-editor"
     class="editor"
-  />
+  >
+    <div
+      v-if="!editor"
+      class="spinner"
+    >
+      <v-progress-circular
+        indeterminate
+        size="100"
+        color="primary"
+      />
+    </div>
+  </div>
 </template>
 
 <script lang="ts">
 import { Vue, Component, Prop, Ref } from 'vue-property-decorator'
-import * as monaco from 'monaco-editor/esm/vs/editor/editor.api'
-import { IGrammarDefinition, Registry } from 'monaco-textmate'
-import { wireTmGrammars } from 'monaco-editor-textmate'
-import themeDark from '@/monaco/theme/editor.dark.theme.json'
-import themeLight from '@/monaco/theme/editor.light.theme.json'
+import type * as Monaco from 'monaco-editor/esm/vs/editor/editor.api'
+let monaco: typeof Monaco // dynamically imported
 
 @Component({})
 export default class FileEditor extends Vue {
@@ -27,7 +35,7 @@ export default class FileEditor extends Vue {
   @Ref('monaco-editor') monacoEditor!: HTMLElement
 
   // Our editor, once init'd.
-  editor: monaco.editor.IStandaloneCodeEditor | undefined = undefined
+  editor: Monaco.editor.IStandaloneCodeEditor | null = null
 
   // Base editor options.
   opts = {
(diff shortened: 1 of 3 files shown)


matmen:
Nice, thanks for your contribution! This reduces the loading times for Fluidd a lot (considering the editor isn't loaded):
|        | DOM Content Loaded   | Load           | Finish         |
|--------|----------------------|----------------|----------------|
| before | 27.74s              | 29.19s        | 40.97s        |
| after  | 19.49s (-30%)       | 21.00s (-28%) | 27.54s (-33%) |

That table assumes the "Fast 3G" preset in Chrome. However, while testing with that preset, I did notice that there's no loading screen for when the editor is loading, so it looks like it failed to load. I think adding a loading screen would be a nice addition to this PR.

Piezoid:
Thank for the feedback.

At first I was tempted to reuse the file loading dialogue, as a mean to maintain visual continuity. However there is no way to get progress feedback from the dynamic `import()`s. Beside, the file loading dialogue seems a bit too specialized for that purpose. 

A simpler solution would be adding a [centered spinner component](https://vuetifyjs.com/en/components/progress-circular/#indeterminate) and use it as a loading component for [Vue's dynamic component](https://v2.vuejs.org/v2/guide/components-dynamic-async.html?redirect=true#Handling-Loading-State). This requires another .vue file, and I can't decide if it should reside in `widget/filesystem` or in `common`. Is there already a generic component for this kind of things?

It could be potentially reused if more dynamic loading is added to the future. I'll try to split everything except main view at the router/tab level, and see if there is any benefit.

Another heavy dependency is the charting module. It is required for the temperature chart on main view, but delaying that could allow faster access to the 'Tool' panel. And I think that's what matters most in panic situations.

Piezoid:
I updated the commit, taking a different approach. Instead of splitting the FileEditor component, I moved all Monaco dependencies into a `setupMonaco.ts` which is loaded asynchronously. This reuse the the `async mounted()` in FileEditor to show a spinner while loading the dependencies, but also while loading the grammars and initializing the editor.

> I'll try to split everything except main view at the router/tab level, and see if there is any benefit.
As expected, splitting the tabs component didn't reduce the size much. Largest chunk is the Settings panel, weighting around 64 KiB. So not worth it in the end.

Imo, further efforts should target the echarts-* dependencies.

matmen:
Nice! I'm fine with the loading spinner, I think it looks good.
I'm not sure why, but the spinner seems to be pushed too far to the bottom of the editor container. It might be connected to the header bar, but I haven't tested that.
image

pedrolamas:
Thank you for doing this @Piezoid and your timing is impeccable as I was looking into this yesterday and didn't came anywhere close to the excellent work you are doing here!

I agree we can and should change the editor features... specifically, I see no reason to disable `multicursor` (I personally use it A LOT!), but there are quite a few we can disable as we don't have any use for them now (like code 'codelens' or 'colorPicker').

Piezoid:
> Thank you for doing this @Piezoid and your timing is impeccable as I was looking into this yesterday and didn't came anywhere close to the excellent work you are doing here!

You're welcome! Actually, seeing you tackling the bundle size issue is what motivated me to have a look, knowing that I wasn't the only one bothered by it.

> I agree we can and should change the editor features... specifically, I see no reason to disable `multicursor` (I personally use it A LOT!), but there are quite a few we can disable as we don't have any use for them now (like code 'codelens' or 'colorPicker').

I let you pick the editor features as I don't know what most of them are for. I don't use the built-in editor much. That's why I only mentioned them as potential future improvement.
 
> For the record, I'm currently checking the Klipper and G-code language definition files (`*.tmLanguage.json` files) and trying to get those updated to with latest keywords!

Nice!

I have two additional PRs in preparation that should shrink the chunk-vendor bellow 1MB.
After that, I will look into the spinner alignment issue. CSS is not strong suit :sweat_smile:

matmen:
> After that, I will look into the spinner alignment issue. CSS is not strong suit 😅

That's absolutely fine! If you want we can handle that part, too.
Again, tysm for your contribution, reducing the bundle size is a major part of performance tuning.

pedrolamas (approved):
LGTM

I've reviewed the code and tests it on desktop Edge and Firefox, no issues at all!

PR #599, opened 2022-03-20
chore: split VChart component and echarts* dependencies

Piezoid:
This enables faster loading of the dashboard.

The rationale is that a faster loading of the dashboard allows for faster access the tool card and jogging control. The echarts components are required for the display of the temperature and bedmesh charts. Whether faster loading is worth delaying the charts remain to be discussed. 
There is currently no visual loading state, but that could be added if needed.

The new async chunk weights 1010 KiB. Combined with #597, the chunk-vendors now weights 1033 KiB.


matmen (approved):
Nice! Thanks for your work.
This changes the dashboard load times to the following (using Chrome's Fast 3G network emulation):
|        | DOM Content Loaded | Load         | Finish      | Usable Dashboard |
|--------|--------------------|--------------|-------------|------------------|
| before | 25.90              | 27.38        | 34.27       | Finish           |
| after  | 21.53 (-17%)       | 23.01 (-16%) | 34.53 (+1%) | 29.77 (-13%)     |

PR #600, opened 2022-03-20
chore: manually tree shake lodash

Piezoid:
Replaces 'lodash' imports with 'lodash/...'.

Only 38 KiB saved this time. `chunk-vendors` is now 995 KiB, combined with previous PRs.

Alternatively, `lodash-webpack-plugin` and `babel-plugin-lodash` could be used for this, but I don't think it is worth the dependency fatigue...


pedrolamas:
Wouldn't it be easier for this one to just replace "lodash" with "lodash-es"?

PR #601, opened 2022-03-20
fix: TouchEvent is not defined on Firefox desktop

Piezoid:
On Firefox desktop, we get undefined references errors to `window.TouchEvent` in the console when the color picker is loaded.

### DIFF
--- a/src/components/ui/AppColorPicker.vue
+++ b/src/components/ui/AppColorPicker.vue
@@ -32,10 +32,10 @@
       <v-card-title
         v-if="title"
         class="card-heading mb-2"
-        style="cursor: move"
+        style="cursor: move; user-select: none;"
         @mousedown="startMouseDrag"
         @touchstart="startTouchDrag"
-        @touchmove="pointerMove"
+        @touchmove="touchMove"
       >
         {{ title }}
       </v-card-title>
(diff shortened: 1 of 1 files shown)


matmen (changes requested):
Sorry, but this doesn't seem right. Swapping the statements for TouchEvents and MouseEvents won't work here, because the order of checks is important (touch events only register on the card header, mouse events on the entire page, so this change breaks the behavior on all browsers).

matmen:
I've gone ahead and fixed this issue (and another one I found) in the above linked PR. Thanks for your effort and reporting this.

Piezoid:
You're right, fixed.

I noticed the event handler were not removed because the component is not unmounted.
Now the event handlers are only registered for the duration of the drag.

Tested with Firefox and Chrome on mobile and desktop.

matmen:
Ah, nice. I think your solution is much more elegant and fixes some more issues. I'll close my PR, thanks for the fix 😉

matmen (approved):

Piezoid:
Sorry I left the tab as a reminder for a while and  didn't see your message.

I also noticed the access `$el` on undefined. With the fix, relativeMove should no longer be called after the card is gone.
It is still not quite clear to me why the beforeUnmount wasn't fired...

pedrolamas (approved):

PR #609, opened 2022-03-24
feat: fullscreen camera page

### LINKED ISSUE #542: Fullscreen Camera View not respecting Settings
### What happened

Greetings!

[Due to how I mount my Camera in my Frame](https://i.imgur.com/3NKCEuY.jpg) I need _both_ Camera Axis flipped for the Camera View to match what I'd see standing in front of the machine.

The settings allow for such an adjustment and the Regular Camera View respects that setting but the Fullscreen Camera View does not.

I'd expect the Fullscreen Camera View to also be respecting the in the Camera Settings made changes.

### How to reproduce

Set up a Camera and in the Camera Settings have one **or** the other **or** both Camera Axis flipped:
The Regular Camera Module **will** respect those changes.
The Fullscreen Camera View **will not** respect those changes.

### Additional information

Nuffin...

matmen:
Adds a dedicated page for the fullscreen camera feed, solving the following issues:
- Camera settings are now respected (rotation, fps, ...)
- The emergency stop button (as well as all other menu items) are now available while being in the full screen view

Currently still opens in a new tab, not sure if this is the appropriate behavior

Tested with (adaptive and streamed) MJPEG sources on:
- Chrome Desktop
- Chrome Mobile (both landscape and portrait orentation)
- Firefox Desktop
- Edge Desktop

image

Closes #542 #279


matthewlloyd:
How does FPS compare against a simple stream tab, when using higher resolution or higher FPS cameras? Please see my comments on UI performance here: https://github.com/fluidd-core/fluidd/pull/525#issuecomment-1045513349

I think my comments from the full-screen console card may be worth bearing in mind here also: https://github.com/fluidd-core/fluidd/pull/613#issuecomment-1079726485

Could we make this behavior optional via a setting, or introduce an additional button for the original direct camera feed? Personally I will always prefer to have a simple direct stream opened in a tab, and I think many other users will too given that very few people are likely using the hflip/vflip options.

I think a good compromise that makes everyone happy might be as follows:

- Add a full-screen webcam nav bar item, with a full-screen view that is essentially just a big version of the existing dashboard card, showing all webcams by default but with the ability to select an individual webcam for a larger view.
- Add a full-screen icon (diagonal arrows) to the webcam dashboard card, which is a shortcut for the webcam nav bar item.
- Keep the existing "raw stream in tab" button (perhaps with a setting to show or hide?) so that people who want a quick way to access the raw stream don't lose the ability to do so.

matmen:
> How does FPS compare against a simple stream tab, when using higher resolution or higher FPS cameras? Please see my comments on UI performance here: [#525 (comment)](https://github.com/fluidd-core/fluidd/pull/525#issuecomment-1045513349)

I'm streaming my main camera at 1080p@30fps using adaptive MJPEG, and I'm getting a stable 30fps in the full screen view with this. The iframe option gives me around 60fps too. I can only see fps limiting happening on the dashboard and it looks to be because of the notifyStatusUpdate dispatching/handling.

> Could we make this behavior optional via a setting, or introduce an additional button for the original direct camera feed? Personally I will always prefer to have a simple direct stream opened in a tab, and I think many other users will too given that very few people are likely using the hflip/vflip options.

I don't really think this is an uncommon issue - from what I've seen on the Discord a lot of people are having issues with the fullscreen view not taking into account rotation and flipping. The only way for them to solve that issue would be editing the webcam config, which isn't trivial (at least much less than clicking a UI dropdown).
I'm fine with keeping the original stream URL (or adding a toggle to the settings), though.

> I think a good compromise that makes everyone happy might be as follows:
> 
> * Add a full-screen webcam nav bar item, with a full-screen view that is essentially just a big version of the existing dashboard card, showing all webcams by default but with the ability to select an individual webcam for a larger view.
> * Add a full-screen icon (diagonal arrows) to the webcam dashboard card, which is a shortcut for the webcam nav bar item.
> * Keep the existing "raw stream in tab" button (perhaps with a setting to show or hide?) so that people who want a quick way to access the raw stream don't lose the ability to do so.

This seems fine to me. Would love to get some input from @pedrolamas (and maybe @Durahl and @PigeonFX too) on this too.

matthewlloyd:
> I'm streaming my main camera at 1080p@30fps using adaptive MJPEG, and I'm getting a stable 30fps in the full screen view with this. The iframe option gives me around 60fps too. I can only see fps limiting happening on the dashboard and it looks to be because of the notifyStatusUpdate dispatching/handling.

Sounds good, thanks for confirming!

> I don't really think this is an uncommon issue

I suspect the majority (90%?) aren't using those options so they would enjoy the raw full-screen tab, and I definitely agree enough people are using rotation and flip for it to be important to provide for those users too. 

> I'm fine with keeping the original stream URL (or adding a toggle to the settings), though.

Sounds good. Can we have both please! :D Seems like we are in agreement. I can see myself using both a full-screen webcam nav bar item which shows both cameras at once, and also the raw stream URL in a new tab option, depending on what I'm trying to see.

matmen:
@matthewlloyd I'm not really happy with all the implementations involving just slapping the camera card onto its own page.. The cameras end up being too small and the button to change cameras causes a global state change to happen, which causes weird UX behavior on the dashboard. From the suggestion (https://github.com/fluidd-core/fluidd/issues/279#issuecomment-873414702) I think just showing a single camera at a time (per page) would be fine, what do you think? I've pushed something I've had stashed for the better part of 3 weeks now, but I'm not sure if I'm happy with it lol

src/components/widgets/camera/CameraItem.vue
```diff
@@ -43,14 +43,14 @@
       fps: {{ currentFPS }}
     </div>
     <div
-      v-if="cameraFullScreenUrl && !fullscreen"
+      v-if="cameraFullScreenUrl && ((fullscreen && fullscreenMode === 'embed') || !fullscreen)"
```
pedrolamas, line 46: I believe this `v-if` expression can be simplified to this:

```
cameraFullScreenUrl && (!fullscreen || fullscreenMode === 'embed')
```

src/store/config/types.ts
```diff
@@ -54,6 +54,7 @@ export interface GeneralConfig {
   timeformat: string;
   showRateOfChange: boolean;
   flipConsoleLayout: boolean;
+  cameraFullscreenAction: 'embed' | 'rawstream';
```
pedrolamas, line 57: Probably better refactor this to a reusable type (thus avoiding repeating it on CameraItem.vue)

matmen:
I went back and tried some alternative ways to solve this today, and so far I wasn't happy with any of them (UX wise, my PR included).. If anyone has specific suggestions I'm happy to implement them

matmen:
Marking this as ready for review as I haven't received more feedback yet, maybe this is worth being an intermediate solution ;)

src/components/widgets/camera/CameraItem.vue
```diff
@@ -1,67 +1,66 @@
...
+    >
+      fps: {{ currentFPS }}
+    </div>
+    <div
+      v-if="cameraFullScreenUrl && (!fullscreen || fullscreenMode === 'embed')"
+      class="camera-fullscreen"
+    >
+      <a
+        :href="cameraFullScreenUrl"
+        :target="fullscreenMode === 'rawstream' ? '_blank' : ''"
```
pedrolamas, line 51:
```suggestion
        :target="fullscreen || fullscreenMode === 'rawstream' ? '_blank' : ''"
```

This bit also needs to check for `fullscreen`, because without that check it will just replace Fluidd page with the camera

pedrolamas (changes requested):

pedrolamas (dismissed):

pedrolamas (approved):

PR #610, opened 2022-03-25
feat: timelapse browser

### LINKED ISSUE #303: Timelapse playback / assign recorded Timelapse to jobs done
since there is the possibility to take timelapses via moonraker: https://github.com/FrYakaTKoP/moonraker/blob/dev-timelapse/moonraker/components/timelapse.py

https://github.com/FrYakaTKoP/moonraker/blob/dev-timelapse/docs/configuration.md

check if files exist in
output_path: ~/timelapse/ and play

a seamlesw integration would be nice, ive read about mainsail is planning to integrate this

thanks in advance,

matmen:
Closes #303 

Currently working:
- [x] list timelapses
- [x] associate corresponding thumbnail automatically
- [x] hide thumbnail files
- [x] download video file
- [x] delete video file
- [x] delete thumbnail on video deletion
- [x] video preview
- [x] status panel (last thumbnail, number of frames, video length)
- [x] common settings on timelapse page
- [x] timelapse settings in UI settings
- [x] render settings popup
- [x] render status
- [x] view specific frame of timelapse
- [ ] assigning recorded timelapses to jobs (will follow in a later PR)
- [ ] load thumbnail files when force_logins is turned on (currently no idea why this isn't working, but gcode thumbs are..)

This was quickly hacked together, I still need to refactor this code, but it's ready for a first try.

image

### DIFF
--- a/src/components/layout/AppNavDrawer.vue
+++ b/src/components/layout/AppNavDrawer.vue
@@ -40,6 +40,13 @@
           </app-nav-item>
 
           <app-nav-item
+            icon="$console"
+            to="/console"
+          >
+            {{ $t('app.general.title.console') }}
+          </app-nav-item>
+
+          <app-nav-item
             icon="$files"
             to="/jobs"
           >
--- a/src/components/layout/AppNavDrawer.vue
+++ b/src/components/layout/AppNavDrawer.vue
@@ -123,7 +123,7 @@ export default class AppNavDrawer extends Mixins(StateMixin) {
   }
 
   get supportsTimelapse () {
-    return this.$store.getters['files/isRootAvailable']('timelapse')
+    return this.$store.getters['server/componentSupport']('timelapse')
   }
 
   get supportsVersions () {
--- a/src/components/layout/AppNavDrawer.vue
+++ b/src/components/layout/AppNavDrawer.vue
@@ -66,7 +66,7 @@
             icon="$video"
             to="/timelapse"
           >
-            {{ $t('app.timelapse.title.timelapse') }}
+            {{ $t('app.general.title.timelapse') }}
           </app-nav-item>
 
           <app-nav-item
--- a/src/components/layout/AppNavDrawer.vue
+++ b/src/components/layout/AppNavDrawer.vue
@@ -66,7 +66,7 @@
             icon="$video"
             to="/timelapse"
           >
-            {{ $t('app.general.title.timelapse') }}
+            {{ $t('app.timelapse.title.timelapse') }}
           </app-nav-item>
 
           <app-nav-item
--- a/src/components/layout/AppNavDrawer.vue
+++ b/src/components/layout/AppNavDrawer.vue
@@ -63,7 +63,7 @@
 
           <app-nav-item
             v-if="supportsTimelapse"
-            icon="$timelapse"
+            icon="$video"
             to="/timelapse"
           >
             {{ $t('app.general.title.timelapse') }}
--- a/src/components/settings/timelapse/TimelapseSettings.vue
+++ b/src/components/settings/timelapse/TimelapseSettings.vue
@@ -78,6 +78,24 @@
       </app-setting>
 
       <v-divider />
+      <app-setting :title="$tc('app.timelapse.title.render_settings')">
+        <app-btn
+          outlined
+          small
+          color="primary"
+          @click="renderSettingsDialogOpen = true"
+        >
+          <v-icon
+            small
+            left
+          >
+            $pencil
+          </v-icon>
+          {{ $t('app.general.btn.edit') }}
+        </app-btn>
+      </app-setting>
+
+      <v-divider />
       <app-setting :title="$t('app.setting.label.reset')">
         <app-btn
           outlined
@@ -89,6 +107,12 @@
         </app-btn>
       </app-setting>
     </v-card>
+
+    <timelapse-render-settings-dialog
+      v-if="renderSettingsDialogOpen"
+      v-model="renderSettingsDialogOpen"
+      :renderable="false"
+    />
   </div>
 </template>
 
@@ -104,12 +128,19 @@ import HyperlapseSettings from '@/components/settings/timelapse/subsettings/mode
 import { CameraConfig } from '@/store/cameras/types'
 import ToolheadParkingSettings from '@/components/settings/timelapse/subsettings/ToolheadParkingSettings.vue'
 import { defaultWritableSettings } from '@/store/timelapse'
+import TimelapseRenderSettingsDialog
+  from '@/components/widgets/timelapse/TimelapseRenderSettingsDialog.vue'
 
 @Component({
-  components: { ToolheadParkingSettings, HyperlapseSettings }
+  components: {
+    ToolheadParkingSettings,
+    HyperlapseSettings,
+    TimelapseRenderSettingsDialog
+  }
 })
 export default class TimelapseSettings extends Mixins(StateMixin) {
   @Ref('delayCompElement') delayCompElement!: any
+  renderSettingsDialogOpen = false
 
   rules = {
     numRequired: (v: number | string) => v !== '' || this.$t('app.general.simple_form.error.required'),
--- a/src/components/settings/timelapse/TimelapseSettings.vue
+++ b/src/components/settings/timelapse/TimelapseSettings.vue
@@ -127,8 +127,9 @@ export default class TimelapseSettings extends Mixins(StateMixin) {
     }]
   }
 
-  get cameras (): {text: string, value: string} {
-    return this.$store.getters['cameras/getCameras'].map((camera: CameraConfig) => ({ text: camera.name, value: camera.id }))
+  get cameras (): {text: string, value: string, disabled: boolean} {
+    return this.$store.getters['cameras/getCameras']
+      .map((camera: CameraConfig) => ({ text: camera.name, value: camera.id, disabled: !camera.enabled }))
   }
 
   get cameraBlocked (): boolean {
--- a/src/components/settings/timelapse/TimelapseSettings.vue
+++ b/src/components/settings/timelapse/TimelapseSettings.vue
@@ -46,6 +46,38 @@
       <toolhead-parking-settings />
 
       <v-divider />
+      <app-setting
+        :title="$t('app.timelapse.setting.stream_delay_compensation')"
+        :sub-title="subtitleIfBlocked(delayCompBlocked)"
+      >
+        <v-text-field
+          ref="delayCompElement"
+          :value="delayComp"
+          :rules="[rules.numRequired, rules.validNum, rules.numMin]"
+          :disabled="delayCompBlocked"
+          :hide-details="delayCompElement ? delayCompElement.valid : true"
+          filled
+          dense
+          single-line
+          suffix="ms"
+          @change="setDelayComp"
+        />
+      </app-setting>
+
+      <v-divider />
+      <app-setting
+        :title="$t('app.timelapse.setting.gcode_verbose')"
+        :sub-title="subtitleIfBlocked(verboseGcodeBlocked)"
+      >
+        <v-switch
+          v-model="verboseGcode"
+          hide-details
+          :disabled="verboseGcodeBlocked"
+          @click.native.stop
+        />
+      </app-setting>
+
+      <v-divider />
       <app-setting :title="$t('app.setting.label.reset')">
         <app-btn
           outlined
@@ -61,7 +93,7 @@
 </template>
 
 <script lang="ts">
-import { Component, Mixins } from 'vue-property-decorator'
+import { Component, Mixins, Ref } from 'vue-property-decorator'
 import StateMixin from '@/mixins/state'
 import {
   TimelapseMode,
@@ -77,6 +109,14 @@ import { defaultWritableSettings } from '@/store/timelapse'
   components: { ToolheadParkingSettings, HyperlapseSettings }
 })
 export default class TimelapseSettings extends Mixins(StateMixin) {
+  @Ref('delayCompElement') delayCompElement!: any
+
+  rules = {
+    numRequired: (v: number | string) => v !== '' || this.$t('app.general.simple_form.error.required'),
+    validNum: (v: string) => !isNaN(+v) || this.$t('app.general.simple_form.error.invalid_number'),
+    numMin: (v: number) => v >= 0 || this.$t('app.general.simple_form.error.min', { min: 0 })
+  }
+
   get supportedModes (): {text: string, value: TimelapseMode}[] {
     return [{
       text: this.$tc('app.timelapse.setting.mode_layermacro'),
@@ -113,6 +153,32 @@ export default class TimelapseSettings extends Mixins(StateMixin) {
 
   set mode (value: TimelapseMode) {
     SocketActions.machineTimelapseSetSettings({ mode: value })
+  }
+
+  get delayCompBlocked (): boolean {
+    return this.$store.getters['timelapse/isBlockedSetting']('stream_delay_compensation')
+  }
+
+  get delayComp (): number {
+    return this.settings?.stream_delay_compensation * 1000
+  }
+
+  setDelayComp (value: number) {
+    if (this.delayCompElement?.validate()) {
+      SocketActions.machineTimelapseSetSettings({ stream_delay_compensation: value / 1000 })
+    }
+  }
+
+  get verboseGcodeBlocked (): boolean {
+    return this.$store.getters['timelapse/isBlockedSetting']('gcode_verbose')
+  }
+
+  get verboseGcode (): boolean {
+    return this.settings?.gcode_verbose
+  }
+
+  set verboseGcode (value: boolean) {
+    SocketActions.machineTimelapseSetSettings({ gcode_verbose: value })
   }
 
   get settings (): TimelapseSettingsType {
--- a/src/components/settings/timelapse/TimelapseSettings.vue
+++ b/src/components/settings/timelapse/TimelapseSettings.vue
@@ -1,0 +1,132 @@
+<template>
+  <div>
+    <v-subheader id="timelapse">
+      {{ $t('app.timelapse.title.timelapse') }}
+    </v-subheader>
+    <v-card
+      :elevation="5"
+      dense
+      class="mb-4"
+    >
+      <app-setting
+        :title="$tc('app.general.title.camera', 1)"
+        :sub-title="subtitleIfBlocked(cameraBlocked)"
+      >
+        <v-select
+          v-model="camera"
+          filled
+          dense
+          single-line
+          hide-details="auto"
+          :items="cameras"
+          :value="camera"
+          :disabled="cameraBlocked"
+        />
+      </app-setting>
+      <v-divider />
+
+      <app-setting
+        :title="$t('app.timelapse.setting.mode')"
+        :sub-title="subtitleIfBlocked(modeBlocked)"
+      >
+        <v-select
+          v-model="mode"
+          filled
+          dense
+          single-line
+          hide-details="auto"
+          :items="supportedModes"
+          :value="mode"
+          :disabled="modeBlocked"
+        />
+      </app-setting>
+
+      <hyperlapse-settings v-if="mode === 'hyperlapse'" />
+
+      <toolhead-parking-settings />
+
+      <v-divider />
+      <app-setting :title="$t('app.setting.label.reset')">
+        <app-btn
+          outlined
+          small
+          color="primary"
+          @click="handleReset"
+        >
+          {{ $t('app.setting.btn.reset') }}
+        </app-btn>
+      </app-setting>
+    </v-card>
+  </div>
+</template>
+
+<script lang="ts">
+import { Component, Mixins } from 'vue-property-decorator'
+import StateMixin from '@/mixins/state'
+import {
+  TimelapseMode,
+  TimelapseSettings as TimelapseSettingsType
+} from '@/store/timelapse/types'
+import { SocketActions } from '@/api/socketActions'
+import HyperlapseSettings from '@/components/settings/timelapse/subsettings/modes/HyperlapseSettings.vue'
+import { CameraConfig } from '@/store/cameras/types'
+import ToolheadParkingSettings from '@/components/settings/timelapse/subsettings/ToolheadParkingSettings.vue'
+import { defaultWritableSettings } from '@/store/timelapse'
+
+@Component({
+  components: { ToolheadParkingSettings, HyperlapseSettings }
+})
+export default class TimelapseSettings extends Mixins(StateMixin) {
+  get supportedModes (): {text: string, value: TimelapseMode}[] {
+    return [{
+      text: this.$tc('app.timelapse.setting.mode_layermacro'),
+      value: 'layermacro'
+    }, {
+      text: this.$tc('app.timelapse.setting.mode_hyperlapse'),
+      value: 'hyperlapse'
+    }]
+  }
+
+  get cameras (): {text: string, value: string} {
+    return this.$store.getters['cameras/getCameras'].map((camera: CameraConfig) => ({ text: camera.name, value: camera.id }))
+  }
+
+  get cameraBlocked (): boolean {
+    return this.$store.getters['timelapse/isBlockedSetting']('camera')
+  }
+
+  get camera (): string {
+    return this.settings?.camera
+  }
+
+  set camera (value: string) {
+    SocketActions.machineTimelapseSetSettings({ camera: value })
+  }
+
+  get modeBlocked (): boolean {
+    return this.$store.getters['timelapse/isBlockedSetting']('mode')
+  }
+
+  get mode (): TimelapseMode {
+    return this.settings?.mode
+  }
+
+  set mode (value: TimelapseMode) {
+    SocketActions.machineTimelapseSetSettings({ mode: value })
+  }
+
+  get settings (): TimelapseSettingsType {
+    return this.$store.getters['timelapse/getSettings']
+  }
+
+  subtitleIfBlocked (blocked: boolean): string {
+    return blocked ? this.$tc('app.timelapse.tooltip.managed_by_moonraker') : ''
+  }
+
+  handleReset () {
+    const nonBlockedEntries = Object.entries(defaultWritableSettings)
+      .filter(([key]) => !this.$store.getters['timelapse/isBlockedSetting'](key))
+    SocketActions.machineTimelapseSetSettings(Object.fromEntries(nonBlockedEntries))
+  }
+}
+</script>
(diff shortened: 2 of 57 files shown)


matthewlloyd:
Tried this out, already very useful, thanks! Beats using WinSCP. Ability to preview the videos will be just icing on the cake.

The nav bar icon is really similar to the history icon - I wonder if we can find one that is more clearly camera-related, e.g. something like this:

image

matmen:
Yep, wasn't sure which icon to use, so I just used whatever mdi offered for `timelapse`
What do you think about [file-video-outline](https://icon-sets.iconify.design/mdi/file-video-outline/)? image
Feel free to check the [available icons](https://icon-sets.iconify.design/mdi/) for any alternatives, I couldn't find one really matching your suggestion 🙂

matthewlloyd:
> What do you think about [file-video-outline](https://icon-sets.iconify.design/mdi/file-video-outline/)?

That one looks great, I think!

matmen:
> Ability to preview the videos will be just icing on the cake.

I just pushed a file preview dialog for this, let me know what you think 😉

matthewlloyd:
Awesome! That works really well. Thanks! This is immediately getting merged into my production build. :)

One small suggestion, I think left clicking should open the preview by default, for consistency with the Config tab / file editor. Can we also get an "X" close icon at the top right of the preview view? I found myself muscle-memory clicking up there.

Incidentally, the way the full-screen button works on the previewer (at least on Chrome) is really nice, taking up the full screen with a little UI chrome at the bottom for controls, and might be a good model to follow for the full-screen webcam view.

src/components/widgets/filesystem/FileSystem.vue
```diff
@@ -370,6 +405,38 @@ export default class FileSystem extends Mixins(StateMixin, FilesMixin, ServicesM
     return this.$store.state.server.info.registered_directories || []
   }
 
+  transformTimelapseItems (items: AppFile[] | AppDirectory[]) {
+    const timelapses: {[key: string]: AppFile} = {}
+
+    for (const item of items) {
+      if (item.type === 'file' && item.extension !== 'jpg') {
+        timelapses[item.filename.slice(0, -(item.extension.length + 2))] = item
```
matthewlloyd, line 413: Should be `slice(0, -(item.extension.length + 1))`.

image

src/components/widgets/filesystem/FileSystem.vue
```diff
@@ -370,6 +405,38 @@ export default class FileSystem extends Mixins(StateMixin, FilesMixin, ServicesM
...
+      }
+    }
+
+    for (const item of items) {
+      if (item.type === 'file' && item.extension === 'jpg') {
+        const name = item.filename.slice(0, -5)
```
matthewlloyd, line 419: Should be `slice(0, -4)`:

image

matmen:
> One small suggestion, I think left clicking should open the preview by default, for consistency with the Config tab / file editor.

This is fine with me, though the behavior in the jobs tab for this is also different (it always opens the context menu). I do think having the preview auto-open makes sense here though.

> Can we also get an "X" close icon at the top right of the preview view? I found myself muscle-memory clicking up there.

Sure! Having two buttons to exit the preview seems kinda redundant, but I don't really mind that, considering it improves UX.
I'm also fine with removing the Close button at the bottom, not sure what option I would prefer here.
image

matthewlloyd:
> This is fine with me, though the behavior in the jobs tab for this is also different (it always opens the context menu). I do think having the preview auto-open makes sense here though.

Ah, good point!

In my personal opinion, I think this needs reworking across all of Fluidd. Currently we're breaking UI conventions established by Windows and Mac OS for mouse behavior when interacting with file lists (e.g. Explorer, Finder):

* Left single click = select (e.g. for bulk operations like delete)
* Left double click = open / edit
* Right single click = popup menu

We should follow these conventions across the Jobs, Configuration, and Timelapse tabs, both for consistency within Fluidd and with established norms. The History tab could also do this, offer selection, and offer the Remove and Print/Reprint actions in the popup menu.

If we can all agree on this, maybe this would make sense for another PR on a later date.

matthewlloyd:
> Sure! Having two buttons to exit the preview seems kinda redundant, but I don't really mind that, considering it improves UX.
> I'm also fine with removing the Close button at the bottom, not sure what option I would prefer here.

Nice! I'm not sure either. If it's just a Close button at the bottom on its own, I guess I'd prefer removing it. On the other hand that might be a good place to put Download and Delete buttons, in which case I'd prefer keeping it.

matthewlloyd:
Another small suggestion: I found I could increase the width of the thumbnails from 24 to 60 without affecting the vertical spacing of the rows:

image

Unfortunately the only up to width=52 works on the Jobs tab without affecting the layout, and the G-code thumbs seem to be only 32x32 (not sure whether it's possible to request larger images from Moonraker), so this would need to be special-cased. This image shows the appearance at width=52 and it looks ugly due to the upscaling:

image

matmen:
> Another small suggestion: I found I could increase the width of the thumbnails from 24 to 60 without affecting the vertical spacing of the rows:

@matthewlloyd I'm not entirely sure about this, considering timelapses can have different aspect ratios too.. If testing shows that this works with those too, I like it. Keep in mind that the big thumbnail is shown in the context menu too ;)

matthewlloyd:
> I'm not entirely sure about this, considering timelapses can have different aspect ratios too..

Ah good call, yes we should be setting the thumbnail size with the height attribute instead of width, so it will work for any aspect ratio. Here's how they look with height=36 (the rows seem to be able to accommodate up to height=38, but it's nice to have a little spacing):

image

I find the small thumbnails too tiny to be useful as they currently are... :)

Later on I'd like to add a secondary view mode option with large thumbnails in a grid layout, here and also for the Jobs tab.

matmen:
@matthewlloyd Hey, I've just implemented your suggestions, lmk what you think :)

matthewlloyd:
> Hey, I've just implemented your suggestion, lmk what you think :)

Thank you, looks great!

Seeing them all lined up now, I might have been a little aggressive with the density and would be open to slightly smaller thumbnails if you think so too... (height=32 maybe?) But also happy with 36!

matmen:
I think 32 looks a lot better - fixed

matmen:
Not too sure about the styling on the footer buttons - any other ideas?

image

matthewlloyd:
> Not too sure about the styling on the footer buttons - any other ideas?

Nice! Looks good to me as you already have it. Something consistent with the other modals would work, I'd think, and you're already there with the colored text, spacing, and right-hand alignment.

image

The footer Close button is possibly redundant now that I see it in context... I could go either way on keeping it.

matthewlloyd:
If I may make some suggestions for these:

> * [ ]  status display (rendering, last frame, ...) as page header

Currently the files list has a lot of unused horizontal space, so perhaps a 1/3 width column on the right hand side of your new view would be a good place for a status panel.

> * [ ]  settings panel above timelapse browser (or on dashboard?)

For settings that might get changed often (enable/disable timelapse, autorender) putting them in the status panel might make sense. Then the status panel could be duplicated in its own (optional) dashboard card, maybe with a mini file browser showing the last few timelapse videos.

For settings that are infrequently accessed (URL setup, render quality settings, etc.) I think a section in the main Settings view after Camera might be good.

Just my 2c. So happy to see this getting added!

matthewlloyd:
> Updated the layout to more closely match the Rename File dialog and pushed those changes 😉

That looks really nice!

matmen:
Currently I have something like this in mind for the status / common settings cards. Feel free to suggest changes to the layout, and also if there should be more content / actions in the status card ;)
For the (common) settings tab, what settings do you think should be on there?

image

matmen:
@phil2sat @sdc53 Feel free to chime in with your suggestions here 😉

FrYakaTKoP:
@matmen  
>For settings that are infrequently accessed (URL setup, render quality settings, etc.) I think a section in the main Settings view after Camera might be good. 

Just note that the snapshoturl isn't writeable via settings endpoint, the recommended implementation is to use the webcams namespace in the moonraker DB. We introduced this because we liked to streamline the snapshot setting between MS and the plugin, would be great if fluidd also changing to this implementation. It would streamline frontend camera settings between the clients and the plugin. 

just to make it clear if you have questions about the settings api or in general usage of the plugin, don't hesitate to ping me here or in Discord (FrY Sennberg).

FrY

matmen:
I pushed my recent changes - feel free to take a look at them. I like your idea of being able to scrub through previous frames.

I'll probably implement the settings section for set-and-forget- settings over the weekend.
I'm not sure which settings should go directly on the timelapse page - I also just use the config file to edit those :p I guess just changes that make sense to be changed on the fly:
* enabled
* park travel speed / retract/extrude distance
* autorender
* CRF

Not sure if these make sense too:
* framerate
* variable frame rate
* hyperlapse cycle (considering it would just speed up the video in the middle of a print?)

I think the park position setting doesn't really belong on the timelapse page, considering it rarely will be changed during a print. It (and all the other settings) are better off living on the usual settings page IMO.
I'm open for discussion or suggestions tho 😉

FrYakaTKoP:
in Mainsail we only included "enabled" directly on the timelapse page, but there is a render popup which allows the user to adjust fps (or activate variable fps & settings). Everything other is offloaded to the Settings page. Actually i doubt if CRF is something people fiddle around with on "per print basis" normally it's set to a "sane" value and left there. 

Other thing which would make alot of sense: Is a switch when starting the print (in the Start print popup) to enable the plugin. This feature have proven to be very useful.

matmen:
I put some more thoughts into this, and IMO having timelapse settings in the fluidd settings section doesn't really make sense.
The settings so far only really affect fluidd and its behavior - not any external parts (printer settings are managed by dedicated pages or cards on the dasbboard).
Additionally, the timelapse settings aren't persistent - they are only stored until the next reboot (not written to the config), so it really only makes sense to set them on a per-render basis (in a popup when clicking the render button, like FrY already mentioned), or on a per-print basis (for the ones that make sense to be changed for individual prints (if there are any other than enable and autorender?); probably in the card on the timelapse page, or maybe in a popup card on print start? But the latter seems kind of invasive).

MichelBarneveld:
@matmen Great work! I'm loving it already! Great to have a way in fluidd to view and download the timelapses on my phone.
For me personally I don't need all the settings of timelapse. Once setup they don't change that much and if needed I am ok to go into the config files just like you need to for a lot of other settings.
I do like the current options to disable or prevent rendering on a current print, that does make sense to have available in the GUI. 
I hope this ends up in a release very soon.

FrYakaTKoP:
> Additionally, the timelapse settings aren't persistent - they are only stored until the next reboot (not written to the config)

Thats not true, timelapse stores it settings in the moonraker Database. Yes it doesn't change the moonraker.conf, but it's not possible to write to the conf from the plugin that's why i used the DB to store them. Again if a user has a setting in the conf this will win over the database setting. Note that there are settings not available in the settings endpoint/database:

```
#output_path: ~/timelapse/
##   Directory where the generated video will be saved
#frame_path: /tmp/timelapse/
##   Directory where the temporary frames are saved
#ffmpeg_binary_path: /usr/bin/ffmpeg
##   Directory where ffmpeg is installed
```
does shouldn't be need to be changed during runtime that's why i excluded them.

also there are other settings which where also not present, which are more a legacy thing and shouldn't be used in combination with a proper intergrated frontend eg. snapshoturl & rotation (which are replaced with the setting in Webcams db namespace). 

Sorry for still not having proper documentation on the api, but again feel free to ask me. Also good starting point is to just GET request the settings endpoint and it will list all available settings.

matmen:
> Thats not true, timelapse stores it settings in the moonraker Database. Yes it doesn't change the moonraker.conf, but it's not possible to write to the conf from the plugin that's why i used the DB to store them.

Ah, that makes sense. Thanks for clearing that up - totally makes sense for the settings to be managable through fluidd then 😃

> Note that there are settings not available in the settings endpoint/database:
> 
> ```
> #output_path: ~/timelapse/
> ##   Directory where the generated video will be saved
> #frame_path: /tmp/timelapse/
> ##   Directory where the temporary frames are saved
> #ffmpeg_binary_path: /usr/bin/ffmpeg
> ##   Directory where ffmpeg is installed
> ```
> 
> does shouldn't be need to be changed during runtime that's why i excluded them.

Yep, this seems fine - it's set-and-forget options anyway, I don't really see a need for having those in the frontend.

> also there are other settings which where also not present, which are more a legacy thing and shouldn't be used in combination with a proper intergrated frontend eg. snapshoturl & rotation (which are replaced with the setting in Webcams db namespace).
> Sorry for still not having proper documentation on the api, but again feel free to ask me. Also good starting point is to just GET request the settings endpoint and it will list all available settings.

Not present in the TimelapseSettings interface? Both the `snapshoturl` and `rotation` seem to be includes in that, but maybe I misunderstood your statement. For now I just deferred the TimelapseSettings interface from the `machine.timelapse.get_settings` response, I assume this has all the settings in it? I'll go ahead and mark the legacy options as readonly in the typings.
As for the timelapse section in the settings, I am going to implement this over the weekend; I think the timelapse page is probably fine just having the `Enable` and `Auto Render` switch? All the render based settings can then be changed in a popup card when clicking the `Render` button, and the non-render related settings are probably off fine being with the other UI settings.

Hope this makes sense - feel free to comment on this or suggest other approaches

FrYakaTKoP:
> Not present in the TimelapseSettings interface? Both the snapshoturl and rotation seem to be includes in that, but maybe I misunderstood your statement. For now I just deferred the TimelapseSettings interface from the machine.timelapse.get_settings response, I assume this has all the settings in it? I'll go ahead and mark the legacy options as readonly in the typings.

Snapshoturl and rotation are presented in the settings endpoint, but snapshoturl is readonly and rotation shouldn't be used. Both got replaced by the settings in the Webcams namespace or the settings associated with the selected cam. 

As said before in mainsail we decoupled the webcam settings from the rest of the Mainsail settings. Now there is a new namespace for webcams. It looks like this:

```
/server/database/item?namespace=webcams
{
	"result": {
		"namespace": "webcams",
		"key": null,
		"value": {
			"5dff4bd9-925a-48d7-a9c4-266b88dec98b": {
				"name": "test2",
				"icon": "mdi-webcam",
				"service": "mjpegstreamer-adaptive",
				"targetFps": 15,
				"urlStream": "/webcam/?action=stream",
				"urlSnapshot": "/webcam/?action=snapshot",
				"flipX": true,
				"flipY": true
			}
		}
	}
}
```
which includes the urlSnapshot and flipX/Y (which is now used instead of rotation). Then in the timelapse settings the user selects the camera (by name but value is the UUID) he want to use for Timelapse and the plugin gets the snapshoturl and "rotation" from the webcam namespace.

For Timelapse the coresponding camera setting would be: (shorted)
``` 
/machine/timelapse/settings
{
  "result": {
    "camera": "5dff4bd9-925a-48d7-a9c4-266b88dec98b",
  }
}
```

My recommendation is to adapt this for fluidd so user don't have to setup camera in two interfaces if using both.

All other settings (again excluding snapshoturl and rotation) presented by the settings endpoint are active and will stay. I recommend to not include snapshoturl and rotation at all (not only mark them as readonly, just ignore them completely).

Also did you get what i meant with the blockedsettings? Every setting present in the moonraker.conf will be prioritised and reported in "blockedsettings", which then need to be disabled / set readonly in the frontend. Just double checked the code and the settings can be written in runtime but on restart of the plugin, the ones in the .conf will get priority again. As said the best solution is to parse blockedsettings and disable the settings in the frontend. There are some cross correlations like "camera" should be deactivated too if "snapshoturl" is set in the moonraker.conf. I need to dig this up again to compile you a list of those linked settings.

matmen:
Added the non-render-related settings to the UI:
image

I'll move the commonly changed render settings to a separate popup, accessible via the `Render` button on the timelapse page.
The uncommon render settings (CRF, saveframes, previewimage, time_format_code, pixelformat and extraoutputparams) are probably off fine not being with the other render settings.

matmen:
Attached are some screenshots:

### Timelapse page
image
### Render dialog
image
### Video preview
image
### Timelapse settings
image

Please let me know if anything needs changing or is missing.

src/components/layout/AppNavDrawer.vue
```diff
@@ -61,6 +61,14 @@
             {{ $t('app.general.title.history') }}
           </app-nav-item>
 
+          <app-nav-item
+            v-if="supportsTimelapse"
+            icon="$video"
+            to="/timelapse"
+          >
+            {{ $t('app.timelapse.title.timelapse') }}
```
pedrolamas, line 69: For consistency purposes, this should be a new key called `app.general.title.timelapse`

src/components/settings/timelapse/TimelapseSettings.vue
```diff
@@ -0,0 +1,199 @@
...
+    <v-subheader id="timelapse">
+      {{ $t('app.timelapse.title.timelapse') }}
+    </v-subheader>
+    <v-card
+      :elevation="5"
+      dense
+      class="mb-4"
+    >
+      <app-setting
+        :title="$tc('app.general.title.camera', 1)"
```
pedrolamas, line 12: It seems the existing settings labels are under `app.setting.label.*`, so maybe move all of these there for consistency reasons?

src/components/settings/timelapse/TimelapseSettings.vue
```diff
@@ -0,0 +1,199 @@
+<template>
+  <div>
+    <v-subheader id="timelapse">
+      {{ $t('app.timelapse.title.timelapse') }}
```
pedrolamas, line 4: `app.setting.title.timelapse` ?

src/components/settings/timelapse/TimelapseSettings.vue
matmen: I'm not a big fan of duplicating translations, though.. Seems like `app.general.title.camera` is used everywhere else, and it fits here perfectly, so I don't see a need to create a new label for it.

src/components/settings/timelapse/TimelapseSettings.vue
matmen: Don't see a need to duplicate this either? I think it's fine if it's changed to `app.general.title.timelapse` like suggested in your first comment.

src/components/settings/timelapse/TimelapseSettings.vue
matmen: Ah, I may have misunderstood - do you want me to move the `app.timelapse.setting.*` keys to `app.setting.label.*`?

pedrolamas: Correct, on this case I'm just saying to move the parent path of these translation keys!

src/components/settings/timelapse/TimelapseSettings.vue
pedrolamas: I do get your point, but I personally try to make little assumptions on this kind of stuff, as the same words can have different meanings depending on region and how they are presented on the screen.

src/locales/en.yaml
```diff
@@ -501,3 +502,56 @@ app:
...
       release_notes: Release notes
+  timelapse:
+    btn:
+      render: Render
+      save_frames: Save Frames
+      open_render_settings: Open Render Settings
+    error:
+      newframe: Error taking timelapse frame
+    label:
+      frames: %{frames} frame | %{frames} frames
```
pedrolamas, line 513:
```suggestion
      frames: '%{frames} frame | %{frames} frames'
```

i18n Ally was showing an error in VSCode for this: as this starts with "%", it needs to be quoted to work properly, or use a ">-" line break!

src/components/settings/timelapse/TimelapseSettings.vue
matmen: I think it's fine as-is for now, when more translations are added and there's conflicts this can always be changed

src/components/settings/timelapse/TimelapseSettings.vue
matmen: Honestly not sure about this, as this is an entirely new "component" and dynamically enabled when the timelapse component is supported in moonraker (bedmesh has its own section too). I think it would clutter the settings labels subkey too much.. What do you think?

src/components/widgets/timelapse/TimelapseRenderSettingsDialog.vue
```diff
@@ -0,0 +1,347 @@
...
+        </v-col>
+
+        <v-col
+          class="text-right"
+          cols="1"
+        >
+          <v-icon @click="$emit('close')">
+            $close
+          </v-icon>
+        </v-col>
```
pedrolamas, line 25:
```suggestion
        <span class="focus--text">{{ $t('app.timelapse.title.render_settings') }}</span>

        <v-spacer />
        <app-btn
          color=""
          icon
          @click="$emit('close')"
        >
          <v-icon>$close</v-icon>
        </app-btn>
```

This should make it match the look of other dialog windows (like "keyboard shortcuts" or "version information")

image

src/components/settings/timelapse/TimelapseSettings.vue
pedrolamas: Fair enough, let's leave it be!

src/components/widgets/timelapse/TimelapseSettingsCard.vue
```diff
@@ -0,0 +1,96 @@
...
+      :r-cols="2"
+    >
+      <v-switch
+        v-model="autoRender"
+        hide-details
+        :disabled="autoRenderBlocked"
+        @click.native.stop
+      />
+    </app-setting>
+    <div v-if="!frameCount">
```
pedrolamas, line 32: What do you think if we just disable the button instead of hiding this whole thing!

Hiding UX stuff is considered an anti-pattern and should be avoided!

src/components/ui/AppSlider.vue
```diff
@@ -252,3 +256,9 @@ export default class AppSlider extends Mixins(StateMixin) {
   }
 }
 </script>
+
+<style>
```
pedrolamas, line 260:
```suggestion
<style lang="scss" scoped>
```

src/components/widgets/filesystem/FilePreviewDialog.vue
```diff
@@ -0,0 +1,111 @@
...
+    return window.innerWidth * (this.$vuetify.breakpoint.mdAndDown ? 1 : 0.5)
+  }
+
+  get isVideo () {
+    return this.file?.type.startsWith('video/')
+  }
+}
+</script>
+
+<style>
```
pedrolamas, line 106:
```suggestion
<style lang="scss" scoped>
```

src/components/widgets/timelapse/TimelapseStatusCard.vue
```diff
@@ -0,0 +1,163 @@
...
+    return this.$store.getters['timelapse/getRenderStatus']
+  }
+
+  get apiUrl () {
+    return this.$store.state.config.apiUrl
+  }
+}
+</script>
+
+<style>
```
pedrolamas, line 152:
```suggestion
<style lang="scss" scoped>
```

src/components/widgets/timelapse/TimelapseSettingsCard.vue
matmen: I think we could move this button to the other timelapse settings. Most of the time the "Render" button will be there anyway, and for cases where it isn't I think having it in the usual settings seems fine. What do you think?

pedrolamas: Makes sense to me!

matmen: Fixed in https://github.com/fluidd-core/fluidd/pull/610/commits/9decaabdffec0a24d9e5fbbf5f65c3eceabecc63 😉

src/components/widgets/filesystem/FileSystem.vue
```diff
@@ -435,7 +507,10 @@ export default class FileSystem extends Mixins(StateMixin, FilesMixin, ServicesM
       } else {
         if (
           e.type === 'click' && item.type === 'file' &&
-          this.$store.state.config.uiSettings.editor.autoEditExtensions.includes(`.${item.extension}`)
+          (
+            this.$store.state.config.uiSettings.editor.autoEditExtensions.includes(`.${item.extension}`) ||
+            this.timelapseBrowser
```
pedrolamas, line 512: Why not leave the default behavior here (follow the settings) instead of forcing auto-open?

src/components/widgets/filesystem/FileSystem.vue
```diff
@@ -621,11 +735,31 @@ export default class FileSystem extends Mixins(StateMixin, FilesMixin, ServicesM
     SocketActions.serverFilesMove(src, dest)
   }
 
-  handleRemove (file: AppFile | AppFileWithMeta | AppDirectory | (AppFile | AppFileWithMeta | AppDirectory)[]) {
+  handleRemove (file: FileBrowserEntry | AppFileWithMeta | (FileBrowserEntry | AppFileWithMeta)[], callback?: () => void) {
```
pedrolamas, line 738: Is `callback` here actually in use? I can't seem to find it anywhere...

pedrolamas (commented):
I think the part that bothers me most here is part of the changes in `FileSystem.vue` that are based on having the `timelapseBrowser` enabled or not...

Would prefer to make this more generic (maybe passing some filter callbacks) and have the consumer page handle it somehow!

src/components/widgets/filesystem/FileSystem.vue
matmen: It's used [here](https://github.com/fluidd-core/fluidd/pull/610/files#diff-7a8df2f9b1d14dee2891d7abf461ac805319ba4cff8599196b6b7e8aba91911cR51)

matthewlloyd:
Great to see so much progress made on this! I merged the latest changes from your branch and have been testing them out. It's looking great. Just some small thoughts here from a user's perspective:

- I do find it a little confusing that the "Timelapse Status" panel on the Timelapse tab only appears once the print has started. At first I thought it might have been removed completely. It makes it less discoverable since a new user might not know that it will eventually appear, and might wrongly conclude that there is no such panel or that it is somewhere else to be found. My suggestion would be to make it always visible, and show a livestream of the webcam before the print starts. That way it is both discoverable, and also useful for setting up camera angles etc. before the print starts.
- After the print starts but before the first frame is captured, it currently shows a broken image icon. Maybe this is also a good time to be showing a live webcam preview, so the user has a chance to correct any issues while the first layer is printing before the first frame is captured:
image
- I couldn't see a way to access the render settings from the Timelapse tab, but maybe I am missing something...
- Can we please add some more render quality settings to the Render Settings dialog? I know Mainsail doesn't have it but I would think the CRF setting would be at least as useful as the Variable FPS and Output Framerate settings, if not more-so. I agree pixel format is not likely to be changed and could go either way on vcodec. Almost every video editing / rendering program ever made offers users the option to adjust output quality settings on the render dialog.

Thanks again!

pedrolamas (approved):

PR #613, opened 2022-03-26
feat: add full-screen console view

matthewlloyd:
image

### DIFF
--- a/src/components/layout/AppNavDrawer.vue
+++ b/src/components/layout/AppNavDrawer.vue
@@ -40,6 +40,13 @@
           </app-nav-item>
 
           <app-nav-item
+            icon="$console"
+            to="/console"
+          >
+            {{ $t('app.general.title.console') }}
+          </app-nav-item>
+
+          <app-nav-item
             icon="$files"
             to="/jobs"
           >
--- a/src/components/widgets/console/ConsoleCard.vue
+++ b/src/components/widgets/console/ConsoleCard.vue
@@ -6,7 +6,8 @@
     content-classes="flex-grow-1 flow-shrink-0"
     menu-breakpoint="none"
     menu-icon="$cog"
-    :draggable="true"
+    :draggable="!fullScreen"
+    :collapsable="!fullScreen"
     layout-path="dashboard.console-card"
     @collapsed="handleCollapseChange"
   >
@@ -81,7 +82,7 @@
       ref="console"
       :scrolling-paused.sync="scrollingPaused"
       :items="items"
-      :height="300"
+      :height="fullScreen ? 816 : 300"
     />
   </collapsable-card>
 </template>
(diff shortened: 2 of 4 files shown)


matmen:
This looks good to me so far, but I'm just wondering: What do you think of having a fullscreen button on the console card instead of a new nav item? IMO the nav bar feels kinda cluttered already and this would fit great into the console card header.
I've gone ahead and pushed an example to my fork, let me know what you think 😉 https://github.com/matmen/fluidd/commit/83af896bec8eb3a5d1f98c6ef178953e7f9dab7b

matthewlloyd:
Thanks for taking a look. Here are a few things to think about:

- Consistency is a key principle of good UI design. Here are two ways removing the full screen console nav item will break consistency:
  - Currently there is a 1:1 mapping between which button is selected in the nav bar, and what is shown in the main section of the UI. Having full-screen routes without corresponding nav buttons breaks this principle. That means you either have no nav button selected while in a full-screen view, which is confusing, or you have the Dashboard button selected while the route is not the dashboard route which means the user has to click on an already selected nav button to get back to the Dashboard.
  - We currently have two types of dashboard cards - cards that are standalone status cards (Cameras, Status, Tool, Limits, Thermals, etc.), and cards which are mini versions of full-screen cards also accessible via a nav bar item (currently just Jobs I think). Having full-screen cards that don't have a nav bar item will increase this to three "kinds", and break consistency with the existing Jobs card.
- Usability. 
  - With just a single full-screen button in the card header, there will be no way to access the Console full-screen view directly without going via the Dashboard, which tends to be slow to load. Also, because the Console card may be below the fold, accessing the button may require scrolling down. An important and primary task like accessing the console should not be so hard to find.
- Mixing metaphors. This is something I think Mainsail gets wrong and Fluidd gets right, which contributes to Fluidd's elegant and lightweight feel.
  - Right now Fluidd has just one UI metaphor in play, which is the mobile metaphor. That's a task-based metaphor, a tabbed interface with just one full-screen window (or task) selectable via large buttons which are easy to press with your thumb. In situations where more information or options need to be shown, we have popup menus. There is never more information on screen than you need for your current task, e.g. we do not have the Dashboard visible while editing Settings.
  - There are two UI metaphors in play in Mainsail, the mobile one, and also the "windowed" desktop metaphor with many small windows on screen, which in theory could be moved around, resized, maximized, and so on. However because of the limitations of the UI framework, you can't actually move the windows, you can't resize them, etc., which is confusing, and also introduces clutter.
  - I think having a full-screen button on the console card will mix the mobile and windowed metaphors in a confusing way.

I think having a full-screen button might still be OK, but only if it is a shortcut for selecting the full-screen console nav bar item.

matmen (dismissed):
Those are good points, thank you for your feedback. This is fine with me then.

matthewlloyd:
If you don't mind, @matmen... I tried it and do really like your idea of having a shortcut button on the console card, so I have added a second commit to this PR which adds one, reusing the existing full-screen icon we have for the webcam. It's essentially a shortcut for the nav bar item. Best of both worlds.

image

pedrolamas (approved):
LGTM!

This is a great improvement, thank you for doing it!! 🙂

matmen (approved):

PR #618, opened 2022-03-29
Added thumbnail URI encoding

bhelm:
Noticed that a thumbnail was not displaying when i uploaded a gcode file with "infill 10%" in its name. Turned out the thumbnail name is basically the gcode name and also contained the % character which was not url encoded. Fixed that and also for the path, as the same problem may happen there.

The gcode name was auto-generated by the cura "Gcode Filename Format Plus" plugin.

### DIFF
--- a/src/store/helpers.ts
+++ b/src/store/helpers.ts
@@ -27,8 +27,8 @@ export const getThumb = (thumbnails: Thumbnail[], path: string, large = true) =>
         if (thumb.relative_path && thumb.relative_path.length > 0) {
           const url = new URL(apiUrl ?? document.location.origin)
           url.pathname = (path === '')
-            ? `/server/files/gcodes/${thumb.relative_path}`
-            : `/server/files/gcodes/${path}/${thumb.relative_path}`
+            ? '/server/files/gcodes/' + encodeURIComponent(thumb.relative_path)
+            : '/server/files/gcodes/' + encodeURIComponent(path) + '/' + encodeURIComponent(thumb.relative_path)
 
           return {
             ...thumb,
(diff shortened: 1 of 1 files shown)


src/store/helpers.ts
```diff
@@ -27,8 +27,8 @@ export const getThumb = (thumbnails: Thumbnail[], path: string, large = true) =>
         if (thumb.relative_path && thumb.relative_path.length > 0) {
           const url = new URL(apiUrl ?? document.location.origin)
           url.pathname = (path === '')
-            ? `/server/files/gcodes/${thumb.relative_path}`
-            : `/server/files/gcodes/${path}/${thumb.relative_path}`
+            ? '/server/files/gcodes/' + encodeURIComponent(thumb.relative_path)
+            : '/server/files/gcodes/' + encodeURIComponent(path) + '/' + encodeURIComponent(thumb.relative_path)
```
pedrolamas, line 31: I haven't tested this yet but the fix makes sense.

Having said that, it would be preferable to keep using string interpolation, like this:

```suggestion
            ? `/server/files/gcodes/${encodeURIComponent(thumb.relative_path)}`
            : `/server/files/gcodes/${encodeURI(path)}/${encodeURI(thumb.relative_path)}`
```

Please note that I also replaced `encodeURIComponent` with `encodeURI` as I believe the former would encode characters like "/" when we need to keep those as is ([context](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI#encodeuri_vs_encodeuricomponent)).

bhelm: yes, ok. Please note that you missed one replacement on line 30.

pedrolamas: Good eye @bhelm, I've just updated the suggestion!

pedrolamas (approved):
LGTM

matmen (approved):

PR #633, opened 2022-04-19
feat: calculate linear progress

pedrolamas:
Klipper calculates progress by calculating `current_file_position / file_length`, but that can lead to progress starting on a non 0% value and the print complete before it reaches 100% (in my tests, I've seen it start at 46% and end at 97%)

The reason for this is that G-code files can have a lot of metadata on the start (thumbnails for example) and end of the file, thus causing the above problem.

As Moonraker reports the start byte and end byte of the G-code, we can use that together with the current file position and calculate a more correct progress value!

Having a "better" progress value will also lead to more accurate timing reports!

Before:

Animation

After:

Animation2

### DIFF
--- a/src/store/printer/getters.ts
+++ b/src/store/printer/getters.ts
@@ -101,6 +101,19 @@ export const getters: GetterTree<PrinterState, RootState> = {
   },
 
   getPrintProgress: (state) => {
+    const { gcode_start_byte, gcode_end_byte, filename } = state.printer.current_file ?? {}
+    const { file_position } = state.printer.virtual_sdcard ?? {}
+
+    if (gcode_start_byte && gcode_end_byte && file_position && filename === state.printer.print_stats?.filename) {
+      if (file_position <= gcode_start_byte) return 0
+      if (file_position >= gcode_end_byte) return 1
+
+      const currentPosition = file_position - gcode_start_byte
+      const endPosition = gcode_end_byte - gcode_start_byte
+
+      if (currentPosition > 0 && endPosition > 0) return currentPosition / endPosition
+    }
+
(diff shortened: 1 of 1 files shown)


matmen (approved):
LGTM, though I'd like to have some feedback from the others too

pedrolamas:
Fair enough, will wait a few days for feedback from the rest of the team!

src/store/printer/getters.ts
```diff
@@ -101,6 +101,19 @@ export const getters: GetterTree<PrinterState, RootState> = {
   },
 
   getPrintProgress: (state) => {
+    const { gcode_start_byte, gcode_end_byte, filename } = state.printer.current_file ?? {}
+    const { file_position } = state.printer.virtual_sdcard ?? {}
+
+    if (gcode_start_byte && gcode_end_byte && file_position && filename === state.printer.print_stats?.filename) {
```
eduncan911, line 107: `filename === state.printer.print_stats?.filename`

With the `const` on line 104 defining `filename` in the scope of this method, and the way this `filename` is being defined inline creates an in-scope variable of, well, `filename`.

However, this variable is not used anywhere in this scope below, between lines 108 to 114.

Do we need this check?

eduncan911 (approved):
LGTM!  Just a coding question, nothing with the actual logic.

src/store/printer/getters.ts
pedrolamas: The `filename` on 104 is from `state.printer.current_file`, the check on 107 is with `state.printer.print_stats` (different parent objects).

This is a sanity check as the 2 parent objects can be updated at different times, making them out-of-sync.

PR #634, opened 2022-04-20
feat(i18n-HU): Update Hungarian translation

AntoszHUN:
Update gcode captions.


src/locales/hu.yaml
```diff
@@ -1,508 +1,499 @@
...
-      release_notes: Kiadási megjegyzések
+app:
+  bedmesh:
+    label:
+      box_scale: Skála
+      flat_surface: Lapos sík megjelenítése
+      mesh_matrix: Hálós mátrix
+      probed_matrix: Próbamátrix
+      profile_name: Profil Név
+      remove_profile: %{name} profil eltávolítása
```
pedrolamas, line 9: The single quote `'` around this translation needs to be kept as this starts with a special replacement sequence.

src/locales/hu.yaml
```diff
@@ -1,508 +1,499 @@
...
+        A nagyításhoz tartsd lenyomva a Shift billentyűt.<br />Kattints egy
+        elemre a grafikonon való megjelenítéshez.<br />Kattints az erősségre a
+        grafikonon való megjelenítéshez.
+  console:
+    label:
+      auto_scroll: Automatikus görgetés
+      flip_layout: Fordított elrendezés
+      hide_temp_waits: Hőmérsékletek elrejtése
+    placeholder:
+      command: Az automatikus kiegészítéshez használd a 'tab'-ot, parancslistához
```
pedrolamas, line 45: As this is a multiline translation, this line needs to keep the `>-` and then put the rest of the translation on the next line indented by 2 spaces (as it was before the change)

src/locales/hu.yaml
```diff
@@ -1,508 +1,499 @@
...
+      edit: Szerkesztés
+      exit_layout: Kilépés az elrendezés módból
+      extrude: Extrudálás
+      heaters_off: Fűtések KI
+      load_all: Az összes betöltése
+      more_information: Több információ
+      pause: Szünet
+      preheat: Előfűtés
+      presets: Előbeállítások
+      preview_gcode: A G-Kód előnézete
```
pedrolamas, line 144: Up to you on this one, but normally I wouldn't expect for "G-Code" to be translated as it is a specific technical term/format.

src/locales/hu.yaml
```diff
@@ -1,508 +1,499 @@
...
+    camera_type_options:
+      mjpegadaptive: MJPEG Adaptív
+      mjpegstream: MJPEG Sugárzás
+      video: IP Kamera
+      iframe: HTTP oldal
+    camera_rotate_options:
+      none: Nincs
+      90: 90°
+      180: 180°
+      270: 270°
```
pedrolamas, line 369: The keys need to be surrounded by single quotes `'`, just as it was before the change!

pedrolamas (changes requested):
You need to keep the line endings on the file as LF but they have been changed to CRLF...

Whatever editor you are using to make these changes, the line endings must be kept as LF (Unix style) instead of CRLF (Windows style).

pedrolamas:
I'm on Windows too!

My advice, use [Visual Studio Code](https://code.visualstudio.com/) and install these 2 extensions:

- [EditorConfig for VS Code](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) - this will ensure you use the same formats as we need to
- [i18n Ally](https://marketplace.visualstudio.com/items?itemName=Lokalise.i18n-ally) - helper extension for translations!
image

AntoszHUN:
I fixed the bugs. Can it be good now?

pedrolamas:
> I fixed the bugs. Can it be good now?

The line endings were still incorrect, but I fixed that for you on the PR (assuming you don't push more updates here, we should be good to merge it)

pedrolamas (approved):

AntoszHUN:
> 



> > I fixed the bugs. Can it be good now?
> 
> The line endings were still incorrect, but I fixed that for you on the PR (assuming you don't push more updates here, we should be good to merge it)

All right. Thanks! I won't update then.

src/locales/hu.yaml
matmen: I think having this as "G-Code" makes sense, the new chinese translation PR also uses the english version and IIRC so does the german translation

matmen:
ah, my bad, it should be `Gcode` not `G-Code`, will fix that

matmen (approved):

AntoszHUN:
> 



> ah, my bad, it should be `Gcode` not `G-Code`, will fix that

Why did I have to rewrite it? In English, G-Code means something. The Hungarian free translation of this is G-Code. Code = Kód. Why wasn't it good now? I translated the klipper3d.org website into Hungarian in 2.5 months. Everywhere the G-Code was rewritten into Hungarian G-Kód. No one fixed it because it wasn't a bug.

matmen:
From what I've seen so far fluidd's convention here is to stick to the english locale for "Gcode" - as you can see [here](https://github.com/fluidd-core/fluidd/blob/develop/src/locales/cn.yaml#L380) and [here](https://github.com/fluidd-core/fluidd/blob/develop/src/locales/de.yaml#L147) for example. I think it makes sense to stick to this, as I think it's more universal.

AntoszHUN:
> 

The translation does not change its meaning.

matmen:
I am aware of that, but fluidd's localization convention is to leave it untranslated. It's untranslated in all the other translations too.

pedrolamas (approved):

AntoszHUN:
> I'm on Windows too!
> 
> My advice, use [Visual Studio Code](https://code.visualstudio.com/) and install these 2 extensions:
> 
>     * [EditorConfig for VS Code](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) - this will ensure you use the same formats as we need to
> 
>     * [i18n Ally](https://marketplace.visualstudio.com/items?itemName=Lokalise.i18n-ally) - helper extension for translations!
>       image

I installed the programs. But for the i18N Ally, how can it be made to work? He's not doing anything to me.

pedrolamas:
if you open the repo (folder) in Visual Studio, it should just detect the translations!

PR #636, opened 2022-04-20
fix: layout database data handling

matmen:
Fixes an issue where the moonraker db entry for the layout would be missing a whole side (container) of the dashboard, making other components fail to load:
```
mutations.ts:46 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'push')
    at mutations.ts:46:50
    at Array.forEach (<anonymous>)
    at mutations.ts:44:27
    at v.setInitLayout (mutations.ts:22:53)

```


pedrolamas (approved):
I've seen that error on my console log quite a few times... and great to see it fixed!! 🙂

PR #509, opened 2022-02-03
Fix: bypass websocket cache for toolhead and motion_report

Biorn1950:
Get instant info from toolhead position and motion_report instead of using cache.
Useful to have the toolhead vue more responsive.

Signed-of-by: Biorn <biorn@me.com>

PR #645, opened 2022-04-26
feat: show live position

pedrolamas:
This is based on #509 but only for "motion_report" (I couldn't see "toolhead" making that much of a difference, but happy to be corrected!)

Live motion changes are shown in the below animation on the top inside `[ ]`:

Animation2

I've also improved a bit on the websocket cache code timings.


src/plugins/socketClient.ts
```diff
@@ -150,26 +150,33 @@ export class WebSocketClient {
...
+
+                // If any of these properties exist, bypass the cache and send immediately
+                for (const key of ['motion_report']) {
+                  if (this.store && key in params) {
+                    this.store.dispatch('printer/onFastNotifyStatusUpdate', { key, payload: params[key] }, { root: true })
+                    delete params[key]
+                  }
+                }
+
+                const timestamp = eventtime ? eventtime * 1000 : new Date().getTime()
```
matmen, line 170: would be best to use `Date.now()` here, as `new Date().getTime()` is about a [third slower](https://jsbench.me/r7l2hmqupz/1)
```suggestion
                const timestamp = eventtime ? eventtime * 1000 : Date.now()
```

matmen (dismissed):
Overall I like this, though it reduces the browser's idle time about 10% for my setup. I've spotted one tiny thing that's not optimal (see comment), but in general I think the entire dashboard needs to be optimized at a later point in time. So far this LGTM tho

src/plugins/socketClient.ts
pedrolamas: I've merged your suggestion, but in all fairness, I doubt the code ever goes there as Moonraker always sends an `eventtime`! 🙂

matmen (approved):

PR #649, opened 2022-04-28
fix: show simple editor on mobile

### LINKED ISSUE #380: Select/Copy/Paste doesn't work on iOS when editing config files
Latest Fluidd build won't let me select/copy/paste bits of config when using the build-in editor.

This worked in earlier builds, before the more fancy VCCode style editor was added.

pedrolamas:
Monaco Editor is not supported on mobile devices (it says so on [their page](https://microsoft.github.io/monaco-editor/)), so instead of having a broken editing experience in mobile devices, better to just use a simple text area!

This change can be easily tested by changing the browser User Agent to a mobile one.

Screenshot from my own OnePlus Android phone:

image

Fixes #380

### DIFF
--- a/src/components/widgets/filesystem/FileEditorTextOnly.vue
+++ b/src/components/widgets/filesystem/FileEditorTextOnly.vue
@@ -3,6 +3,7 @@
     class="editor v-input v-textarea theme--dark"
     :readonly="readonly"
     :value="value"
+    :spellcheck="false"
     @change="emitChange($event.target.value)"
   />
 </template>
@@ -35,7 +36,6 @@ export default class FileEditorText extends Vue {
     font-size: 1rem;
     font-weight: 100 !important;
     min-width: 100%;
-    height: 90% !important;
     height: calc(100% - 48px) !important;
     resize: none;
   }
--- a/src/components/widgets/filesystem/FileEditorTextOnly.vue
+++ b/src/components/widgets/filesystem/FileEditorTextOnly.vue
@@ -1,6 +1,6 @@
 <template>
   <textarea
-    class="editor v-input v-textarea theme--dark"
+    class="editor v-input v-textarea theme--dark px-2"
     :readonly="readonly"
     :value="value"
     :spellcheck="false"
--- a/src/components/widgets/filesystem/FileEditorTextOnly.vue
+++ b/src/components/widgets/filesystem/FileEditorTextOnly.vue
@@ -1,0 +1,42 @@
+<template>
+  <textarea
+    class="editor v-input v-textarea theme--dark"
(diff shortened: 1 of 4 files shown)


src/components/widgets/filesystem/FileEditorTextOnly.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+  <textarea
+    class="editor v-input v-textarea theme--dark"
+    :readonly="readonly"
+    :value="value"
```
matmen, line 5:
```suggestion
    :value="value"
    :spellcheck="false"
```

src/components/widgets/filesystem/FileEditorTextOnly.vue
```diff
@@ -0,0 +1,42 @@
...
+}
+</script>
+
+<style lang="scss" scoped>
+  .editor {
+    font-family: monospace;
+    font-size: 1rem;
+    font-weight: 100 !important;
+    min-width: 100%;
+    height: 90% !important;
```
matmen, line 38:
```suggestion
```

src/util/is-mobile.ts
```diff
@@ -0,0 +1,5 @@
+const isMobile = () => {
+  return /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent)
```
matmen, line 2:
```suggestion
  return (navigator as any).userAgentData?.mobile ?? /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent)
```
for browsers that support it (probably more reliable), fallback to your way

src/components/widgets/filesystem/FileEditorTextOnly.vue
matmen: overwritten by next line

matmen (changes requested):

src/components/widgets/filesystem/FileEditorTextOnly.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+  <textarea
+    class="editor v-input v-textarea theme--dark"
```
matmen, line 3: looks like i forgot to add this comment: i think line starts are kinda hard to read on mobile, so a padding would be good here (just like monaco has)
```suggestion
    class="editor v-input v-textarea theme--dark px-2"
```

other than that it lgtm :)

matmen (changes requested):

matmen (approved):

PR #657, opened 2022-05-06
feat: File editor structure view

CF3B5:
Add editor structure sidebar
structure sidebar


pedrolamas:
This looks really neat, thanks for you contribution @CF3B5!

Have you tested this in a mobile device?

src/components/widgets/filesystem/FileEditor.vue
```diff
@@ -52,10 +126,26 @@ export default class FileEditor extends Vue {
...
   get isMobile () {
     return this.$vuetify.breakpoint.mobile
   }
 
+  get configLink () {
+    const configMap = this.$store.getters['server/getConfigMapByFilename'](this.filename)
+    let link = configMap.link
+    if (configMap.service) {
+      const configLink = 'app.file_system.config.' + configMap.service + '.link'
+      if (this.$t(configLink)) {
```
matmen, line 142: `$t()` always returns a string (it has a fallback to the string being input), so the wrong href is added to the helper links

```suggestion
      if (this.$t(configLink) !== configLink) {
```

src/components/widgets/filesystem/FileEditorDialog.vue
```diff
@@ -57,6 +57,19 @@
             </v-icon>
             <span v-if="!$vuetify.breakpoint.smAndDown">{{ $t('app.general.btn.config_reference') }}</span>
           </app-btn>
+          <app-btn
+            v-if="!printerPrinting && configMap.link"
```
matmen, line 61: The structure toggle button is hidden while printing, I don't believe that should be the case

```suggestion
            v-if="configMap.link"
```

matmen (changes requested):
Thanks for your contribution, I love this! I just took a quick look at it, and noticed two small things:

src/components/widgets/filesystem/FileEditor.vue
```diff
@@ -1,38 +1,112 @@
...
+          {{ $t('app.file_system.label.structure') }}
+        </v-card-subtitle>
+        <v-divider />
+        <div class="pa-2 heading">
+          <div
+            v-for="(item, index) in structures"
+            :id="'line_'+item.index"
+            :key="index"
+            style="cursor: pointer"
+            :class="(item.section!==undefined?'':'pl-5')+' px-2 d-flex justify-space-between align-center'"
```
pedrolamas, line 37:
```suggestion
            :class="{'pl-5': item.section === undefined}"
            class="px-2 d-flex justify-space-between align-center"
```

src/components/widgets/filesystem/FileEditor.vue
```diff
@@ -1,38 +1,112 @@
...
+          <v-icon small>
+            $tree
+          </v-icon>
+          {{ $t('app.file_system.label.structure') }}
+        </v-card-subtitle>
+        <v-divider />
+        <div class="pa-2 heading">
+          <div
+            v-for="(item, index) in structures"
+            :id="'line_'+item.index"
```
pedrolamas, line 34:
```suggestion
            :id="`line_${item.index}`"
```

src/components/widgets/filesystem/FileEditor.vue
```diff
@@ -1,38 +1,112 @@
...
+            :key="index"
+            style="cursor: pointer"
+            :class="(item.section!==undefined?'':'pl-5')+' px-2 d-flex justify-space-between align-center'"
+            @click="gotoLine(item.index)"
+          >
+            <div
+              v-if="item.section!==undefined"
+              class="py-1"
+            >
+              <span :class="$vuetify.theme.dark?'mtk1':'mtk1'">[</span> <span
```
pedrolamas, line 44: Class is the same no matter if dark theme is in use or not... is this what you wanted?

src/components/widgets/filesystem/FileEditor.vue
```diff
@@ -1,38 +1,112 @@
...
+                :class="$vuetify.theme.dark?'mtk13':'mtk4'"
+              >{{ item.section }}</span> <span :class="$vuetify.theme.dark?'mtk1':'mtk1'">]</span>
```
pedrolamas, line 46: Class is the same no matter if dark theme is in use or not... is this what you wanted?

src/components/widgets/filesystem/FileEditor.vue
```diff
@@ -1,38 +1,112 @@
...
+              v-else
+              class="py-1"
+            >
+              <span :class="$vuetify.theme.dark?'mtk1':'mtk13'">{{
+                item.name
+              }}</span>
+            </div>
+            <a
+              v-if="item.section!==undefined && configLink"
+              :href="configLink + '#' + item.section_name"
```
pedrolamas, line 58:
```suggestion
              :href="`${configLink}#${item.section_name}`"
```

src/store/files/types.ts
```diff
@@ -146,3 +149,11 @@ export interface FilePreviewState {
   type: string;
   appFile?: AppFile;
 }
+
+export interface Structure {
+  index: number;
+  section: string | undefined;
+  section_name: string | undefined;
```
pedrolamas, line 156:
```suggestion
  section?: string;
  section_name?: string;
```

pedrolamas (commented):
Most of the things I pointed out in my review are cosmetic, but are there in order to keep code consistency.

I did however noticed while testing this with MS Edge that there is no scrollbar for this sidebar, so scrolling can only be performed with the mouse wheel. Is this expected?

Having said that, I do like this feature but it does seem to me it adds a bit to the existing Monaco sidebar (we actually end up with TWO sidebars).

Having said that, this suggestion did make me wonder if we shouldn't add some codelens capabilites to the code editor for contextual help! 🤔

CF3B5:
Thank you very much for your help! I rarely used TypeScript to write code before, so there should be many low-level mistakes in it :-P, I will check it again in the following days.

src/components/widgets/filesystem/FileEditor.vue
CF3B5: It used to be different, but the latest version I found is the same, I'm not sure that, so I left this code till now.

src/components/widgets/filesystem/FileEditor.vue
CF3B5: It used to be different, but the latest version I found is the same, I'm not sure that, so I left this code till now.

pedrolamas:
@CF3B5 thank you for your work on this, but I think we are going in another way as we've now started work to add Codelens, region folding, and other features to the Monaco editor, I think that will be the way going forward for the internal code editor: https://github.com/fluidd-core/fluidd/pull/665

CF3B5:
> @CF3B5 thank you for your work on this, but I think we are going in another way as we've now started work to add Codelens, region folding, and other features to the Monaco editor, I think that will be the way going forward for the internal code editor: #665

Have you considered the CodeMirror editor? Monaco editor experience on mobile devices is not good.

CF3B5:
I think the structure view is there to find the configuration faster, if you can, so it's better to make them work together :-)

matmen:
Sorry, but I don't exactly see how that makes finding things easier/faster, it basically just shows the whole config but without the configured values, no? The Codelens PR also added support for section folding, which would make finding entries even faster I think.

CF3B5:
> Sorry, but I don't exactly see how that makes finding things easier/faster, it basically just shows the whole config but without the configured values, no? The Codelens PR also added support for section folding, which would make finding entries even faster I think.

Sorry for not expressing clearly because my English is really bad! :-)
The structure view is not just for display, clicking on the item in the structure view will jump to the key/value row. Maybe most IDEs will have a structure view, so for long code, I'm used to looking in the structure view, just like the catalog of a book, and I don't need to scroll the entire file repeatedly. I think value is not the purpose of structure, because it is not to replace the configuration file, it is just to quickly locate the position that needs to be modified, and quickly tell me what functions are in this configuration file, that's all.

sjqlwy:
Very nice & convenient function!

PR #665, opened 2022-05-09
feat: adds Codelens and code folding support

pedrolamas:
Adds Codelens support to the editor, showing direct links to the Klipper and Moonraker documentation websites!

Also adds code folding support for each section and comments!

Pressing F1 will bring all the enabled commands in Monaco, including "fold all"!

image

image

Clicking on the links above will open the correct documentation website!

I've checked the build outputs and seems there was little impact on the bundle size, but it is something to keep an eye on!


src/locales/en.yaml
```diff
@@ -82,6 +82,8 @@ app:
       file_name: Filename
       transfer_rate: Transfer rate
       uploaded: Uploaded
+      open_klipper_config: Open Klipper config
+      open_moonraker_config: Open Moonraker config
```
matmen, line 86: Maybe these can be merged together and replaced with `Open Documentation`? Or maybe something dynamic, like `View temperature_sensor documentation`

pedrolamas: Done!! I've updated the screenshots with the changes!

PR #668, opened 2022-05-12
fix: job card history metadata race condition

### LINKED ISSUE #662: Jobs Does Not Display Print Duration or Total Duration, but History Does
### Browser

Chrome

### Device

Desktop PC

### Operating System

Windows

### What happened

When I access the Jobs section on the home page, or the Jobs sidebar page, Print Duration and Total Duration are empty. When I visit History, they show correctly.

Home Page Jobs: https://i.imgur.com/r49ltdW.png
Jobs Page: https://i.imgur.com/QBRbWYH.png
History: https://i.imgur.com/Wh3clmU.png

I expect PD and TD to show in jobs sections.

### How to reproduce

Confirmed with another user on Reddit that they have the same issue.

### Additional information

Willing to post whatever necessary in terms of logs/code, let me know.

matmen:
Closes #662

### DIFF
--- a/src/store/history/actions.ts
+++ b/src/store/history/actions.ts
@@ -38,6 +38,7 @@ export const actions: ActionTree<HistoryState, RootState> = {
   async onHistoryList ({ commit }, payload) {
     if (payload) {
       commit('setHistoryList', payload)
+      SocketActions.serverFilesGetDirectory('gcodes', 'gcodes')
     }
   },
 


pedrolamas:
Can't we just force a dispatch from `onHistoryList` action instead to reset the history on all files?

matmen:
The problem is that the file system browser isn't reactive to the history state changing. `onHistoryList` being dispatched doesn't update the stats in the file browser either.

pedrolamas (approved):

PR #678, opened 2022-05-15
feat: clear console

pedrolamas:
image


src/store/console/mutations.ts
```diff
@@ -114,5 +114,9 @@ export const mutations: MutationTree<ConsoleState> = {
       state.consoleFilters.splice(i, 1)
       state.consoleFiltersRegexp.splice(i, 1)
     }
+  },
+
+  setLastCleared (state) {
+    Vue.set(state, 'lastCleared', Date.now())
```
pedrolamas, line 120: Ideally, we would use the latest received server timestamp here instead of `Date.now()` but for now this seems to do the trick just fine!

matmen (approved):

PR #680, opened 2022-05-17
feat: replaces Keyboard Shortcuts with Command Palette

pedrolamas:
No point showing a custom incomplete Keyboard Shortcuts dialog when we can just show the full Monaco Command Palette!

image

The above is shown when the user presses the <kbd>F1</kbd> key while focused on the Monaco editor. In full VSCode, the default is <kbd>Ctrl+Shift+P</kbd>.


matmen (dismissed):
lgtm!

src/components/widgets/filesystem/FileEditorDialog.vue
```diff
@@ -32,17 +32,16 @@
         <v-spacer />
         <v-toolbar-items>
           <app-btn
-            v-if="!printerPrinting"
-            target="_blank"
-            @click="handleKeyboardShortcuts"
+            v-if="!printerPrinting && !isMobile"
```
matmen, line 35:
```suggestion
            v-if="!isMobile"
```

matmen (changes requested):

matmen (approved):

PR #700, opened 2022-05-28
fix: manual controls no longer cause out-of-range errors

### LINKED ISSUE #395: Jogging the printer out of range leaves its state in relative mode
## Affected Version(s):
(The versions of fluidd this issue occurs on.)
- klipper version: v0.9.1-659-g69d9497d
- moonraker version: v0.7.1-34-g5522e15
- fluidd version: v1.16.2

## Platform:
Windows 10 x64 + Chrome

## Hardware Configuration:
RPi 3B+, SKR1.3

## Reproduction Steps:
1. Home printer
2. Attempt to move printer out of bounds using **jog controls**
3. Attempt to move printer inside printer bounds using **console**

## Actual Results:
1. After move out of range error, printer ends up in relative mode (presumably because the G90 of the jog controls gets ignored)
2. Console based move yields unexpected results (relative move instead of absolute)

## Screenshot(s):
image

## Expected Results:
1. Even after move out of range error, printer is returned to absolute mode
2. Console based move yields expected result

## Log Files:
(necessary log is shared in screenshot)

## Additional information
As can be seen from the screenshot, the printer ends up at `Y=100` which is a relative move from the previous `Y=70` position. The expected result was an absolute move to `Y=30`

My interpretation of this error is that, upon a `Move out of range` error, the remainder of the commands sent by the jog controls are discarded, so the printer ends up in relative mode (due to the G91 command at the beginning of the jog move)

Cheers,

Florian

matmen:
Closes #395

Fixes the following two issues:
* The printer will no longer try to move out-of-range (stops at min/max axis position now)
* The coordinate space will now always return to whatever was set before the manual move (no longer stays in relative mode after erroring)

Caveats:
* Spamming the move buttons will no longer move multiples of the distance at a time (the toolhead position needs to update first) - maybe the buttons can be disabled during the gcode wait for better UX?

### DIFF
--- a/src/components/widgets/toolhead/ToolheadMoves.vue
+++ b/src/components/widgets/toolhead/ToolheadMoves.vue
@@ -221,6 +221,18 @@ export default class ToolheadMoves extends Mixins(StateMixin, ToolheadMixin) {
     this.moveLength = val
   }
 
+  get toolheadPosition () {
+    return this.$store.state.printer.printer.toolhead.position
+  }
+
+  get printerConfig () {
+    return this.$store.getters['printer/getPrinterConfig']()
+  }
+
+  get usesAbsolutePositioning () {
+    return this.$store.state.printer.printer.gcode_move.absolute_coordinates
+  }
+
   /**
    * Send a move gcode script.
    */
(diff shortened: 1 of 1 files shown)


pedrolamas:
Not being able to spam the button is quite a loss IMHO...

I was thinking that maybe we can add a compromise solution here: add a button to the toolhead controls that allows toggling between relative/absolute positioning, while showing the current state!

That way there will be no surprise to users when they send commands directly from the command line!

PR #706, opened 2022-06-01
feat: moves MCU info to separate panel

pedrolamas:
Moves MCU information from System panel to separate panels (one for each MCU) and shows more information on dialog.

image

image


matmen (approved):
love this!!

PR #718, opened 2022-06-14
fix: show absolute/relative positioning controls

pedrolamas:
Adds controls to change between relative/absolute positioning mode.

image

image

image

Related to #395


matmen (approved):
I do like this, but I don't think it really solves #395 (it's not obvious to the user that they need to reset to absolute moves when relative moves error). I think after merging we do still have to figure something out for that (though I'm not sure what would be fitting considering the disadvantages of #700)

pedrolamas:
That's quite fair, to be honest, I'm not sure what's the best way to solve this issue either...

In an ideal world we would have some type of option to send "on error" commands, but that would require some upstream changes.

PR #719, opened 2022-06-14
feat: allows 4 layout columns in dashboard

### LINKED ISSUE #189: 3 Column Layout with Resizable Cards
I think it would be nice to have an option to have a 3 column layout on the dashboard. Also, the ability to resize individual cards would be helpful. I would prefer to not have to scroll to see everything while printing, and currently some cards are just too large for my liking.

Thanks for the consideration

pedrolamas:
image

image

Closes #189 


pedrolamas:
@matmen this is in "working condition", meaning that it does the job assuming that users who will be using this have wide screens!

What I mean is that for "regular" screens (4k @ 100% scale for example), the card buttons will not collapse to "..." so it will look a bit ugly...

I did some work to fix this, but quite honestly I'm not even sure it is worth it... what do you think?

matmen:
I took a look at this a couple weeks back and liked it. For normal use 3 columns is enough IMO, though 4 columns do make sense for bigger (ultrawide?) screens. Eventually I think we can make layouts user-based, that way you could have different layouts per user (or per device if you log in with a different username on mobile), and I think fixing the wrapping issues is worth it as of now (if it's an annoyance the user can just switch to 3 columns).
I'll take another look this evening and add my review

pedrolamas:
Fair enough! I think the "easiest" way to fix the buttons wrapping/collapse is by storing in state the "column count" for each layout, and then the components can use that to make layout decisions (in this case, the "button group" currently collapses for "md and lower", but for more columns probably should be "lg and lower")

I'll see if I can take another look into this today! 🙂

pedrolamas:
@matmen I've now added a way to dynamically collapse the menus given the column width!

I couldn't find a way of doing this with pure CSS, so this is the best I could think of...

matmen (approved):
nice!

PR #721, opened 2022-06-17
feat: ldap authentication support

matmen:
Adds a authentication source dropdown to the login page (only if LDAP authentication is configured) with the corresponding default:

Disables the corresponding user controls when a user is LDAP sourced:

### DIFF
--- a/src/views/Login.vue
+++ b/src/views/Login.vue
@@ -22,12 +22,12 @@
             v-if="error"
             type="error"
           >
-            Invalid credentials
+            {{ $t('app.general.simple_form.error.credentials') }}
           </v-alert>
 
           <v-text-field
             v-model="username"
-            label="Username"
+            :label="$t('app.general.label.username')"
             autocomplete="username"
             filled
             dense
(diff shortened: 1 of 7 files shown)


matmen:
@lucaschoeneberg feel free to review too :)

lucaschoeneberg (dismissed):
I like what you've done.

src/views/Login.vue
```diff
@@ -102,12 +115,24 @@ export default class Login extends Vue {
...
+
+  mounted () {
+    httpClient.get('/access/info', { withAuth: false })
+      .then(r => {
+        this.source = r.data.result.default_source
+        this.availableSources = r.data.result.available_sources
+      }).catch(() => {
+        // ignore
+      })
+  }
```
pedrolamas, line 129: Maybe move this to `auth.api.ts`, I thinlk it would make much more sense there!

pedrolamas (changes requested):

src/views/Login.vue
matmen: Moved it to the auth store actions, seemed more fitting to avoid multiple imports of the auth API

src/views/Login.vue
```diff
@@ -102,12 +114,19 @@ export default class Login extends Vue {
   valid = true
   error = false
   loading = false
+  source = 'moonraker'
+  availableSources = [this.source]
+
+  async mounted () {
+    const authInfo = await this.$store.dispatch('auth/getAuthInfo')
+    Object.assign(this, authInfo)
```
pedrolamas, line 122: Personal opinion here, but I was never a big fan of this type of approach as it hides what is actually being updated on the target. Nevertheless, this is correct, so all good!

pedrolamas (approved):

pedrolamas (approved):

PR #723, opened 2022-06-17
fix: LED controls behavior

### LINKED ISSUE #641: RGBW LED modal doesn't include white

My feature request is related to a problem. I recently installed a strip of SK6812 RGBW lights on my printer. They function well, however, in fluidd the modal that shows up by default doesn't include a white channel slider. I've found that if I manually assign the gcode `SET_LED LED=overhead_leds RED=0 GREEN=0 BLUE=0 WHITE=1` it will show the white channel but if I use that same white channel slider to turn off the white channel subsequently it will show just red, green, and blue without white.

Here is the modal with the white channel slider (after manually issuing the custom gcode):
image

When I slide the white channel down, the modal immediately refreshes to this version, without the white channel slider:
image

I think the best solution would include some way to read the printer.cfg file to determine the style by looking at `color_order`, then show the appropriate modal.

Alternative solutions may be a way to tell fluidd that these are RGBW so it knows to always show the white channel slider. Worst case, always display the white channel but that doesn't feel very "nice". :)

### Additional information

_No response_

matmen:
Closes #641

### DIFF
--- a/src/components/ui/AppColorPicker.vue
+++ b/src/components/ui/AppColorPicker.vue
@@ -48,7 +48,7 @@
         </v-icon>
 
         <v-icon
-          v-if="white"
+          v-if="supportedChannels.includes('W')"
           :color="whiteColor.hexString"
           large
         >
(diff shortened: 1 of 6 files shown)


pedrolamas (dismissed):
@matmen I've approved this PR, but you might want to check this before merging: https://github.com/matmen/fluidd/pull/3

matmen:
Looks good, thanks

pedrolamas (approved):

PR #726, opened 2022-06-21
feat: adds Save Config & Restart button

pedrolamas:
Adds a new "Save config & restart" button that will call `SAVE_CONFIG` if there are pending configuration changes.

## Wide

image

## Medium

image

## Small

image

## Preview popup

image

Also adds tooltips to the other app bar buttons:

image

image


pedrolamas:
Agreed, the current UX is bad... switching the two buttons probably makes more sense than the current approach (personally, disabling the button just to maintain the position doesn't feel right), or showing it elsewhere....

As for the existing "Save" buttons for bed mesh leveling, z-offset, etc., these are based on state of those specific functions.

The state of this button is based on the value of the Klipper `configfile.save_config_pending` flag, meaning that even manual invocation of commands such as `MANUAL_PROBE`, `Z_ENDSTOP_CALIBRATE`, `PROBE_CALIBRATE`, `DELTA_CALIBRATE` (that are not available in Fluidd UI, but will show the `SAVE_CONFIG` on command line)

There is also a new `configfile.save_config_pending_items` that was recently introduced in Klipper and this has all the pending changes that will be committed on `SAVE_CONFIG`

matmen:
As this is only for manually invoked commands (from what I understood), maybe it makes sense to put it in the header of the console card? It doesn't really only affect the console, of course, but it would be a subtle way to place it. It feels kind of out of place in the main header. Lmk what you think :)
Additionally, I think a tooltip or modal that shows which settings have been changed (/ will be saved) would be nice (we already have the data; though it of course isn't necessary).

pedrolamas:
> As this is only for manually invoked commands (from what I understood), maybe it makes sense to put it in the header of the console card? It doesn't really only affect the console, of course, but it would be a subtle way to place it. It feels kind of out of place in the main header. Lmk what you think :)

Honestly don't know... my point here was not to be subtle and ensure we "remember" the user that he has pending changes that have not been saved yet!

> Additionally, I think a tooltip or modal that shows which settings have been changed (/ will be saved) would be nice (we already have the data; though it of course isn't necessary).

Very good point, I will add this in! 😁

pedrolamas:
How about something like this?

image

Adding a button to Console doesn't feel right to me (these changes can be caused by macros!), but I was thinking that if the top is not an option, we could instead use the status card (2nd position on the screenshot)

matmen:
I think having it at the top formatted like a button like in your screenshot looks good 👍

pedrolamas:
@matmen PR updated with the changes we discussed. For now it is using a `textarea` for the changes preview as I had some issues ensuring all nodes were open on popup!

matmen (approved):
Still not sure about the text box tbh, but I've tried it and it's functional, so I'm fine with that :)

pedrolamas:
I've merged as-is, but we can revisit and test other layouts before we release the next version!

PR #733, opened 2022-06-29
chore: all refs strongly typed

pedrolamas:
Uses strong types for all Refs and replaces `this.$refs` usage with `Ref` decorator.


matmen (approved):
nice!

Issue #729, opened 2022-06-24
Power Button on Top Bar


Martyfree123:
### Is your feature request related to a problem? Please describe

May sound stupid, but clicking the sidebar menu, then power, then printer is an unnecessary few extra steps every time you want to power on/off the machine. 

### Describe the solution you'd like

Instead, would it be possible to have the Moonraker printer power feature/button moved to the Top Bar for easier access? Just a simple one click on/off icon right next to the user icon would be perfect.

For other power devices they could reside in the current menu since they (_probably_) aren't used as often.

### Describe alternatives you've considered

_N/A_

### Additional information

Definitely a very minor complaint, but I figured I'd put it out there. Thank you for the amazing interface, I love Fluidd very much!

Martyfree123:
Didn't mean to close it. My bad.

Martyfree123:
Lol

PR #736, opened 2022-07-03
feat: add icon to shutdown on AppBar

BastelKlug:
Closes: #729 

- Shutdown icon on AppBar/TopBar
- Entry in the settings to (de)activate

Here you can see it in action:

https://user-images.githubusercontent.com/44060099/177038146-131b4d90-e044-4868-8f4e-533ad995fa6f.mov

### DIFF
--- a/src/components/layout/AppBar.vue
+++ b/src/components/layout/AppBar.vue
@@ -87,7 +87,7 @@
               </v-icon>
             </app-btn>
           </template>
-          <span>{{ $t('app.general.btn.shutdown') }}</span>
+          <span>{{ $t('app.general.btn.shutdown') }} ({{ $t('app.general.label.host') }})</span>
         </v-tooltip>
       </div>
 
--- a/src/components/layout/AppBar.vue
+++ b/src/components/layout/AppBar.vue
@@ -226,17 +226,15 @@ export default class AppBar extends Mixins(StateMixin, ServicesMixin) {
     })
   }
 
-  handleHostShutdown () {
-    this.$confirm(
+  async handleHostShutdown () {
+    const res = await this.$confirm(
       this.$tc('app.general.simple_form.msg.confirm_shutdown_host'),
       { title: this.$tc('app.general.label.confirm'), color: 'card-heading', icon: '$error' }
     )
-      .then(res => {
-        if (res) {
-          this.$emit('click')
-          this.hostShutdown()
-        }
-      })
+
+    if (res) {
+      this.hostShutdown()
+    }
   }
 
   saveConfigAndRestart (force = false) {
--- a/src/components/layout/AppBar.vue
+++ b/src/components/layout/AppBar.vue
@@ -70,6 +70,28 @@
       </div>
 
       <div
+        v-if="authenticated && socketConnected && shutdownOnAppBar"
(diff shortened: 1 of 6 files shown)


matmen (dismissed):
Thanks for your PR! This looks good to me already, though I'm going to mention my UX concerns here again (https://github.com/fluidd-core/fluidd/pull/726#issuecomment-1162354264): Moving the e-stop button conditionally may break its UX (muscle memory wise), though I think it's not too bad in this case (considering the user actively has to toggle it and it doesn't toggle visibility based on an internal state). In the above mentioned PR the button was moved to the left of the e-stop button, but I'm fine with either position in this case, so fully up to you.

pedrolamas (changes requested):
I wonder if we should be more specific and instead of just "Shutdown" have "Host Shutdown" as that is what it is all about... @matmen what do you think?

src/components/layout/AppBar.vue
```diff
@@ -199,6 +226,19 @@ export default class AppBar extends Mixins(StateMixin) {
...
+      this.$tc('app.general.simple_form.msg.confirm_shutdown_host'),
+      { title: this.$tc('app.general.label.confirm'), color: 'card-heading', icon: '$error' }
+    )
+      .then(res => {
+        if (res) {
+          this.$emit('click')
+          this.hostShutdown()
+        }
+      })
+  }
```
pedrolamas, line 240:
```suggestion
  async handleHostShutdown () {
    const res = await this.$confirm(
      this.$tc('app.general.simple_form.msg.confirm_shutdown_host'),
      { title: this.$tc('app.general.label.confirm'), color: 'card-heading', icon: '$error' }
    )

    if (res) {
      this.hostShutdown()
    }
  }
```

That `$emit` call should not be needed here (that was on the `SystemCommands.vue` as it probably was needed to dismiss the popup menu, but that is not the case here!

Ideally, I think we should move from `Promise` based code to more modern `async/await` calls, so I went ahead and refactored this method!

matmen:
> I wonder if we should be more specific and instead of just "Shutdown" have "Host Shutdown" as that is what it is all about...

@pedrolamas Sure, sounds good to me

BastelKlug:
Thank you for your improvements. 

I have updated the code, so the text looks like this:
Bildschirmfoto 2022-07-03 um 23 12 00

pedrolamas (approved):

Martyfree123:
Thank you to @BastelKlug for making my feature request into a reality. This is awesome.

Martyfree123:
When will this be included in a stable release?

Martyfree123:
Awesome thank you!

matmen:
Actually @Martyfree123, I just realized your feature request asked for a button to toggle _printer_ power.. This PR added a button to shut down the _host_ (Raspberry Pi), so I think there was some kind of misinterpretation of your FR.. Just to clarify - would you like a toggle button for the `Power` side-bar entries (image)?

Martyfree123:
Yes, I was looking for a button to toggle printer power instead of digging into the side bar for it.

Martyfree123:
I missed that from the video as well. Damn I was really excited :/

PR #739, opened 2022-07-04
feat: device power toggle in top navigation

### LINKED ISSUE #729: Power Button on Top Bar

matmen:
Closes #729 - Supersedes #736
image
image

Also swaps the power switch icon with toggle switch icons for more intuitive UX:
image image
image image

Signed-off-by: Mathis Mensing <matmen@dreadful.tech>


src/locales/en.yaml
```diff
@@ -240,6 +240,7 @@ app:
       'off': 'Off'
       password: Password
       power: Power
+      power_toggle_caption: Toggle power for device %{device}
```
pedrolamas, line 243:
```suggestion
      power_toggle_caption: Toggle %{device} power
```

pedrolamas (approved):
This is great!! 🙂

I've suggested an alternative translation for the tooltip, seems to me more to the point.

Now that I've had the chance to play with this a bit more, I do wonder if the button should have the same style as the others (white and round) instead of using the same layout as the Emergency Stop...

Martyfree123:
You guys are so fast, thanks so much!

Martyfree123:
Personally I think it would be nice if it retains the colour matching but maybe uses the standard power icon?

matmen:
> Now that I've had the chance to play with this a bit more, I do wonder if the button should have the same style as the others (white and round) instead of using the same layout as the Emergency Stop...

> Personally I think it would be nice if it retains the colour matching but maybe uses the standard power icon?

Both suggestions sound good to me, I'll implement those later 😉

matmen:
> Personally I think it would be nice if it retains the colour matching but maybe uses the standard power icon?

@Martyfree123 Just to clarify - the green in the icon is only because of my fluidd theme. Would you like the icon to be colored (green and red?) accordingly or for it to just be white (or theme colored)?

Martyfree123:
I would prefer it was theme colourd. :)

pedrolamas (approved):

PR #743, opened 2022-07-05
feat: Support full range of klipper PA scale

Haxk20:
Klipper has the pressure advance step set to 6 decimal points.

While Smooth time doesnt need such a precise step the actual pressure advance value
can definitely benefit from the increase in resolution when making changes during print as
when tuned finely it can go easily to 5 decimal points.

Thus lets increase the step of the pressure advance value 

Signed-off-by: Martin Botka <martin.botka@somainline.org>


pedrolamas:
Thank you for this @Haxk20 

I admit I do have my doubts on this in terms of presentation, but also in terms of usability (it's impossible to take advantage of that level of precision with the slider control)

At the same time I write this, I'm discussing with Klipper devs to understand the expected impact on that precision level (or at what precision should we expect real impact)

But I am also curious on one thing: how are you calculating/calibrating a PA value that has a precision higher than 0.001?

Haxk20:
> Thank you for this @Haxk20
> 
> I admit I do have my doubts on this in terms of presentation, but also in terms of usability (it's impossible to take advantage of that level of precision with the slider control)
> 
> At the same time I write this, I'm discussing with Klipper devs to understand the expected impact on that precision level (or at what precision should we expect real impact)
> 
> But I am also curious on one thing: how are you calculating/calibrating a PA value that has a precision higher than 0.001?

Its just the final tuning on it. Im on VORON V2.4 currently at 0.0215 on PA. I can understand that this would make probably no sense for most machines but for machines that print at 300mm/s and want to still get OK top layer on corners then fine tuned PA is crucial alongside reducing Smoothing time to 0.02.

pedrolamas:
> If you want me to make it just 4 decimal resolution do let me know :)

Fair enough, 4 decimals sounds a lot more reasonable! Let's go with that then!

src/components/widgets/toolhead/PressureAdvanceAdjust.vue
```diff
@@ -14,7 +14,7 @@
         :locked="(!klippyReady || isMobile)"
         :min="0"
         :max="2"
-        :step="0.001"
+        :step="0.000001"
```
pedrolamas, line 17:
```suggestion
        :step="0.0001"
```

pedrolamas (changes requested):

pedrolamas (approved):

matmen (approved):

Issue #494, opened 2022-01-29
force_move toggle


Biorn1950:
Adding a slider in the toolhead card if the force_move is allowed to freely move the toolhead for maintenance usage.

pedrolamas:
This sounds interesting and I think we should take a look into it, but how would this really work? What commands would we be sending to Klipper?

matmen:
> This sounds interesting and I think we should take a look into it, but how would this really work? What commands would we be sending to Klipper?

FORCE_MOVE is the command you're looking for 😉

pedrolamas:
Given this most likely will put the printer in an incorrectly homed stat, I wonder if it would make more sense to have it on the "Tune" panel instead of the main one?

matmen:
I think this would make sense if the printer isn't homed at all, with big warnings telling users that it's a bad idea and may physically break the printer. Maybe a checkbox (or a button to toggle until next home) in a menu in the tool card header that just enables the movement arrows?

PR #750, opened 2022-07-08
feat(ui): Add FORCE_MOVE support to tool controls

BastelKlug:
Closes: #494 

Hi,
I am here with another PR 😁 I hope I have not misunderstood the request.

I have made a video again to demonstrate the function:

https://user-images.githubusercontent.com/44060099/177890838-768aa72c-2c1f-42e2-b1f1-78fba8b4be3b.mov

I have not included a warning, anyone who manually activates force_move should actually know the dangers. Personally, I would find it annoying. If it is still desired, I can work on it.

This PR is not complete and needs some css work. I've been working on it for days but failed, maybe someone else can help. There is a comment under Toolhead.vue. 

You can see the css disaster by marking the background:
missing-css

If anyone has any questions about the code, please contact me on Discord: BastelKlug#5605


BastelKlug:
Hi, I have fixed the style errors. In theory it is now adoptable.

So here is the current status:
update-force_move

pedrolamas:
Thanks for this @BastelKlug, I've just taken it for a quick test and it works very well! 😁

Having said that, I personally am not a fan of the override positioning...

I do wonder if it would look better as a button here:

image

If you look at the temperature charts, we have a button there that has "Chart Off/On", and I think the same approach can be taken here...

@matmen what do you think?

matmen:
I agree, having it in the card header would be much better placement-wise. I also think that some kind of warning definitely is necessary here, optimally every time FORCE_MOVE is enabled, but I'm also fine with having a setting under General that shows/hides the checkbox in the tool card (which then displays a warning when it's toggled on).

BastelKlug:
Hi, I will work on the changes and upload them when they are ready.

pedrolamas (dismissed):

pedrolamas (approved):

pedrolamas:
This is really great, thank you for your work here @BastelKlug! 🙂

I've now approved the PR, but made a couple of changes before we merge this down, specifically:

- dropped the visibility toggle for FORCE_MOVE button (if it is enabled in printer.cfg, then always show)
- some minor styling tweaks for consistency
- added `public` to any Prop for proper lint alignment
- refactored logic on toolhead buttons `color` and `disabled` states

pedrolamas:
I do wonder if we should also change the movement buttons from the existing up/down/left/right arrows to minus/plus instead... any thoughts on that @BastelKlug @matmen ?

BastelKlug:
Bildschirmfoto 2022-07-13 um 12 49 10
Bildschirmfoto 2022-07-13 um 12 53 28

I think the old design with the arrows looks better 🤔

pedrolamas:
> ...but I'm also fine with having a setting under General that shows/hides the checkbox in the tool card (which then displays a warning when it's toggled on).

@matmen I notice now that you actually mentioned having a toggle button to show/hide the FORCE_MOVE button on the interface - the one I just removed (oops!)

I wonder if we actually need this toggle, given the button will only show if enabled in the printer.cfg (same as the QGL, bed screws, z_tilt, etc.)

matmen:
My thoughts were that users that sometimes use FORCE_MOVE could disable the toggle for a less cluttered UI, but seeing the changes I don't really mind having it there anymore. I'll review this later :)

matmen:
Ah! I remember why I wanted it 😉
image

BastelKlug:
> Ah! I remember why I wanted it 😉 image

Oh I don't know if it's realistic, but if there are too many buttons, it should actually change to the mobile layout (with the dropdown menu) I had already noticed it when I was developing

matmen (commented):
Functionality wise it works fine, though I'm wondering it it maybe should be disabled when the printer is homed? Maybe I'm missing something here, but what's the use case of using FORCE_MOVE when the printer is already homed and can be moved - is it just for moving it out of range (that's what I use FORCE_MOVE for mostly)?

pedrolamas:
> ...maybe should be disabled when the printer is homed?

True, but if Klipper allows it, why should we limit the feature ourselves? It's not like we disable the homing buttons when the printer is homed either! 🙂

I know some users have overridden the default homing procedure and use FORCE_MOVE instead (I believe there's even an "advanced homing macro" somewhere, though I can't remember where I saw it now...)

matmen:
> > ...maybe should be disabled when the printer is homed?
> 
> True, but if Klipper allows it, why should we limit the feature ourselves? It's not like we disable the homing buttons when the printer is homed either! 🙂

True, fair enough 😂

matmen (approved):

PR #754, opened 2022-07-08
feat: Exclude Objects

### LINKED ISSUE #359: Exclude Region
## Thank you for using fluidd!

Nono! Thank YOU  for all your hard work!

The main plugin I miss from my Octoprint/Klipper days is the "exclude region" one.

I'd like to see it implemented as part of the GCode Preview box,.

There are no alternatives that I know of.

Users that have ever had a print consisting of many objects, and had one get messed up, will love you for it.

matmen:
Closes #359 
Credits and big thanks to @troy-jacobson and @kageurufu for doing most of the work here!

Using the Gcode viewer to cancel objects:
image
image
image
image

Alternatively using the cancel button at the top of the Gcode viewer:
image


kageurufu:
Woohoo!

```
Signed-off-by: Troy Jacobson <troy.d.jacobson@gmail.com>
Co-authored-by: Franklyn Tackitt <git@frank.af>
```

src/components/widgets/gcode-preview/GcodePreviewCard.vue
```diff
@@ -17,6 +17,18 @@
         </app-btn>
       </app-btn-collapse-group>
 
+      <app-btn
+        color=""
+        class="ml-1"
+        fab
+        x-small
+        text
+        :disabled="!parts.length"
```
pedrolamas, line 26: This didn't work for me until I changed it to this:
```suggestion
        :disabled="parts.length === 0"
```

src/store/parts/mutations.ts
```diff
@@ -0,0 +1,54 @@
...
+    if ('excluded_objects' in payload) {
+      Vue.set(state, 'excludedParts', payload.excluded_objects)
+    }
+
+    if ('objects' in payload) {
+      const partMap: { [key: string]: Part} = {}
+      payload.objects.forEach((obj: PartObject) => {
+        const name = obj.name
+        const part: Part = {
+          name: name,
```
pedrolamas, line 28:
```suggestion
          name,
```

src/components/widgets/gcode-preview/GcodePreviewCard.vue
pedrolamas: Actually, this is incorrect too, as parts is an object, not an array!

pedrolamas, line 26: Start a print, wait for this to enable, press CANCEL... still enabled!
image
```suggestion
        :disabled="!klippyReady || !printerPrinting || !printerFileLoaded || !parts.length"
```

pedrolamas (changes requested):
Something feels off about this dialog! 
image

matmen:
> Something feels off about this dialog!

I agree, it doesn't look very great, but I'm not sure how else to style it. If you have any suggestions please let me know ;)
Maybe dynamic width of the dialog as well as having a fixed size for the actions column would be enough here, I'll need to try that

src/components/widgets/gcode-preview/GcodePreview.vue
```diff
@@ -276,6 +285,19 @@ export default class GcodePreview extends Mixins(StateMixin) {
     return this.panning ? 'optimizeSpeed' : 'geometricPrecision'
   }
 
+  get showExcludeObjects () {
+    const file = this.$store.getters['gcodePreview/getFile']
+    if (!file) {
+      return true
```
pedrolamas, line 291: I believe this should be `false` (no file loaded, so don't show any exclude objects)
```suggestion
      return false
```
If true, I see the cancel objects even if the current file loaded is not the one printing!

pedrolamas:
> If you have any suggestions please let me know

is the issue horizontal overflow? We could just set it to show "..." (and add a tooltip if you think it's worth it) in that case and revert the window size changes, no?

src/components/widgets/gcode-preview/GcodePreview.vue
matmen: Actually, this change doesn't seem right. It's there so you can cancel objects without having to have the gcode preview loaded. I tried this, and it looks like if the gcode preview has another file loaded it's disabled accordingly..

pedrolamas: Ah I see!

However, it seems a bit weird to me to show the cancellation object on the gcode preview, but no gcode preview...

matmen: I've tried it and UX wise it's fine I think.. And it has the big advantage of not having to parse the entire Gcode file on the client

src/components/widgets/exclude-objects/ExcludeObjects.vue
```diff
@@ -41,30 +41,9 @@ import { Icons } from '@/globals'
...
-    const names = []
-    for (const name in parts) {
-      const excluded = this.isPartExcluded(name)
-      if (parts[name].target &&
-        ((state === 'printing' || state === 'paused') ||
-        ((state === 'complete' || state === 'cancelled' || state === 'error') && excluded))) {
-        names.push(name)
-      }
-    }
-    return names
```
matmen: From my testing this isn't actually needed and causes some weird behavior where outlines are shown but icons aren't. Looking at the state the target and outline always seems to be set.

matmen:
> is the issue horizontal overflow? We could just set it to show "..." (and add a tooltip if you think it's worth it) in that case and revert the window size changes, no?

Yep, that's exactly the issue. If you want to take a shot at it feel free to, my brain can't figure out how to use CSS properly right now 🤣

src/components/widgets/exclude-objects/ExcludeObjectsDialog.vue
```diff
@@ -0,0 +1,85 @@
...
+      </v-card-title>
+
+      <v-card-text class="py-3 px-5">
+        <v-simple-table>
+          <tbody>
+            <tr
+              v-for="part in parts"
+              :key="part.name"
+            >
+              <td class="partName">{{ part.name }}</td>
```
pedrolamas, line 24: Linting:
```suggestion
              <td class="partName">
                {{ part.name }}
              </td>
```

pedrolamas (approved):

pedrolamas:
Just tested the latest and feature wise everything seems to be working fine now! 🙂

PR #756, opened 2022-07-10
feat: Increase resolution of speed and flow sliders to single decimal…

Haxk20:
… point

Usually no decimal value is fine for tuning but in some cases it may not be enough.
Klipper already supports flow and speed adjustment to a few decimal points but such
resolution is kinda pointless. 1 decimal point is more then enough for most users.

Acceleration values and corner speeds support the same as they are float as well but
these sliders were left untouched as they truly do not need decimal resolution (You can't see
the difference in prints).

Signed-off-by: Martin Botka <martin.botka@somainline.org>

### DIFF
--- a/src/components/widgets/toolhead/SpeedAndFlowAdjust.vue
+++ b/src/components/widgets/toolhead/SpeedAndFlowAdjust.vue
@@ -14,6 +14,7 @@
         :locked="(!klippyReady || isMobile)"
         :min="1"
         :max="200"
+        :step="0.1"
         @change="handleSetSpeed"
       />
     </v-col>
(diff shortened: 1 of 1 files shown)


matmen:
I'm hesitating to approve this because this feels odd UI-wise and I honestly don't really see the need for the extra precision here..
Even on a 24 hour print, the added precision (for speed) only amounts to about 1m26s of "tunability" (though I'm not sure what speed would need to be tuned for anyway?).
For flow calibration this would account for one gram of filament per entire 1kg spool, which is below what I would consider material uniformity alone could deliver - even my prusament spools have a deviation of about 17 grams per 1kg.

pedrolamas:
I agree with @matmen here. While I see value in your previous PR to fine-tune PA values during a print, the impact on having more precision on speed and flow with this change should not be significant in terms of print quality (nor I'd expect any major difference in fine-tuning)

Issue #406, opened 2021-09-09
Background Color on Mobile only Matches Default Theme


Kaiede:
## Affected Version(s):
(The versions of fluidd this issue occurs on.)
- klipper version: 0.9.1-773-gc84956be
- moonraker version: 1.0.2-96-g2ab00c6
- fluidd version: 1.16.2

## Platform:
iPhone 12 (iOS 15)

## Hardware Configuration:
Pi 4B, Spider Board

## Reproduction Steps:
1. Browse to your fluidd instance on iPhone.
2. Configure a theme color other than the default fluidd blue.
3. Scroll past the bottom to trigger the rubber band effect in either Safari or while it's pinned to your home screen as a PWA.

## Actual Results / Screenshot:
![IMG_2226](https://user-images.githubusercontent.com/41651119/132738213-0a7703fb-d08b-440c-a755-9ccab4dedeb1.PNG)

![IMG_2227](https://user-images.githubusercontent.com/41651119/132738346-e57227d6-d570-4867-8353-a4568c09ce9c.PNG)

Notice that while the theme color being used is green, the background and status bar are using the fluidd blue color instead. Unfortunately, I don't know if this reproduces on iOS 14.

iOS 15 is also tinting the status bar to try to match the website content now in Safari and PWAs, which makes the mismatch a bit more obvious.

## Expected Results:
While a bit of a nit, it would be great if the background matched up with the theme better, either by using the accent color selected in settings (if you want the contrast), or the background color (if wanting it to blend in better).

## Log Files:
(The log files fluidd provides at Configure > System Control.)
- klipper log:
- moonraker log:

## Additional information
(Extra information relevant to the issue, links to similar issues.)

PR #774, opened 2022-07-17
feat: match theme-color with primary-color

BastelKlug:
Closes: #406

Hi, I think the changes are self-explaining. 
Here is a demo video:

https://user-images.githubusercontent.com/44060099/179393694-a09d7429-03a4-4e2a-86b5-5c5831f70527.mp4

### DIFF
--- a/vue.config.js
+++ b/vue.config.js
@@ -19,7 +19,7 @@ module.exports = defineConfig({
     }
   },
   pwa: {
-    themeColor: '#2196F3',
+    themeColor: '',
     msTileColor: '#000000',
     appleMobileWebAppCache: 'yes',
     manifestOptions: {
--- a/src/App.vue
+++ b/src/App.vue
@@ -89,7 +89,10 @@ import { LinkPropertyHref } from 'vue-meta'
 
     return {
       title: pageTitle,
-      link: pageIcon
+      link: pageIcon,
+      meta: [
+        { name: 'theme-color', content: this.primaryColor }
+      ]
     }
   }
 })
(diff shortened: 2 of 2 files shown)


pedrolamas (approved):
Thank you @BastelKlug, I tested on Windows 11 with Edge browser, working great! 🙂

matmen (approved):

matmen:
neat, thanks!

vue.config.js
```diff
@@ -19,7 +19,7 @@ module.exports = defineConfig({
     }
   },
   pwa: {
-    themeColor: '#2196F3',
+    themeColor: '',
```
pedrolamas, line 22: @BastelKlug was there any specific reason for this line to change?

I ask this as I've noticed this warning on the console caused by it:

image

BastelKlug: Yes, when I removed the line, it took the default value from the plugin. So I left it blank. Maybe there is another way 🤔

pedrolamas: MS Edge in Win11 seems to work fine with the value in... What browser/phone did you notice that you needed to clear that? Might be the case here...

BastelKlug: Chrome on Windows 11. Maybe it was a bug with the Vue Cli Hot reloading?

BastelKlug: Chrome on MacOS, also does not work:

https://user-images.githubusercontent.com/44060099/181530634-372550b6-c623-44e8-ae4b-c375b877c059.mov

And also the bar in Safari that you see when you scroll up too high:
Bildschirmfoto 2022-07-28 um 16 17 19

pedrolamas: Ah, I just tested with Chrome on Win11 and worked fine... guess this is a Mac thing only then?

PR #776, opened 2022-07-18
chore: updated GitHub Pages dependencies

pedrolamas:
No real impact expected from this change (output is static files!), only reason I'm updating to latest versions is to ensure full browser compatibility and mitigate some security warnings that are showing on GitHub.


matmen (approved):
Nice! This also fixes the issue where jekyll would complain about the remote_theme not being found, so it solves one of the issues with local building/serving of the docs; the other one being having to install webrick.. Maybe we can add this to the Gemfile too? Not sure if that would break something with GH pages, but I don't think it should

PR #777, opened 2022-07-19
chore: removes @0x4447/potato from devdependencies

pedrolamas:
We use [@0x4447/potato](https://www.npmjs.com/package/@0x4447/potato) to publish the compiled files to AWS, however this package is old and I believe deprecated as the author deleted the source repo for it...

It's old and it is "polluting" the other dependencies we have, but it is not maintained anymore so there is no way of updating it!

Ideally, we should move away from it ASAP, but for now I am just removing it from the development dependencies and just using it directly via `npx` instead.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,7 +47,6 @@
         "vuex": "^3.6.2"
       },
       "devDependencies": {
-        "@0x4447/potato": "^2.4.7",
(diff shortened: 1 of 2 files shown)


matmen (approved):
From what I can see `actions/setup-node` sets up npx, so this should work, though I didn't test it

PR #784, opened 2022-07-21
chore: code refactor & cleanup

pedrolamas:
Refactors ZHeightAdjust.vue to clear typing warnings.


matmen (changes requested):
Looks like this breaks the radio buttons (I pressed up/down on each entry):
image
image

pedrolamas:
Amazingly, all it took was for me to remove the `_` prefix from the variable... I learned the hard way that in Vue, one should NOT prefix with underscore as those are reserved...

matmen (approved):

PR #785, opened 2022-07-23
feat: adds new Manual Probe helper dialog

### LINKED ISSUE #761: UI for manual bed levelling

I recently got an Ender 3 S1 Pro, and unfortunately got unlucky with a misbehaving CRTouch. Repeatability is fine, but there's probe location bias that I haven't been able to track down. No major problem though, as I can create a manual mesh that works pretty damn well. However, the lack of a UI wizard or utility makes this really annoying to deal with in fluidd. Manually typing in the TESTZ commands is painful and makes the whole process much slower.

A simple UI that provides easy and quick methods for starting a manual mesh level and issuing the testz commands to find the right height.

It doesn't need to be very complex, just a simple layout to start the process, then 6 buttons to move the head up and down by various amounts (1mm, 0.1mm, 0.01mm). I've found what works well for me is going down 1mm at a time until the nozzle touches the paper, then going up 0.1mm at a time until the paper moves freely, then going down 0.01mm until the paper touches again.

Creating macros for issuing the correct commands does work, but the layout is cumbersome and makes it slower to work through a larger mesh. Ideally the UI would be as easy as possible to encourage people without ABL (or with broken ABL) to actually create and update a bed mesh instead of relying purely on corner levelling.

### Additional information

_No response_

### LINKED ISSUE #782: Manual Probe helper UI

Assuming https://github.com/Klipper3d/klipper/pull/5652 gets merged, we will be able to know when a manual probe is running and report the Z position values and send `TESTZ` commands!

Show a UI to easily manage manual probe

Use the command line as we currently do...

### Additional information

_No response_

pedrolamas:
Will show buttons for all the "Z Adjust values" from settings + `1` and `0.1` in case they are not there as this might need more broad changes.

There's a new "Show Manual Probe dialog automatically" setting that will automatically open the dialog if we are on the Dashboard or Tune page and a Manual Probe command is initiated.

Dialog:

image

Button added to dashboard page (only visible when a Manual Probe command is running):

image

Button added to tune page (only visible when a Manual Probe command is running):

image

Fixes #761
Fixes #782

### DIFF
--- a/src/components/common/ManualProbeDialog.vue
+++ b/src/components/common/ManualProbeDialog.vue
@@ -1,0 +1,242 @@
+<template>
+  <v-dialog
+    :value="value"
+    :max-width="450"
+    scrollable
+    @input="$emit('input', $event)"
+  >
+    <v-form @submit.prevent="sendAccept">
+      <v-card>
+        <v-card-title class="card-heading py-2">
+          <span class="focus--text">{{ $t('app.tool.tooltip.manual_probe') }}</span>
+        </v-card-title>
+
+        <v-divider />
+
+        <v-card-text>
+          <v-row>
+            <v-col>
+              <v-text-field
+                label="Z Min"
+                outlined
+                hide-details
+                dense
+                disabled
+                :value="zPositionLower"
+              />
+            </v-col>
+            <v-col>
+              <v-text-field
+                label="Z"
+                outlined
+                hide-details
+                dense
+                disabled
+                :value="zPosition"
+              />
+            </v-col>
+            <v-col>
+              <v-text-field
+                label="Z Max"
+                outlined
+                hide-details
+                dense
+                disabled
+                :value="zPositionUpper"
+              />
+            </v-col>
+          </v-row>
+
+          <v-row class="bysect-row">
+            <v-col
+              cols="3"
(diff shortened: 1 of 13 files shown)


matmen:
This is super useful!! Works perfectly, just one small question - I think using --/- and ++/+ is also possible when no low/high spot is set? Not sure I would use it (it's just not coarse enough to get the nozzle close enough to the bed), but I don't think it should be disabled when it's functional
image
Manual `TESTZ Z=--` moves `0.2mm` down
image

Just also noticed that this breaks the horizontal layout of the card header when using 3 columns, don't think this needs to be fixed in this PR tho:
image

pedrolamas:
> I think using --/- and ++/+ is also possible when no low/high spot is set?

I just tested, you are correct, so will enable that!

> Just also noticed that this breaks the horizontal layout of the card header when using 3 columns, don't think this needs to be fixed in this PR tho:

Will fix that on a separate PR (probably will just put everything under a menu except MOTORS OFF and this new button as it only shows if a manual probe tool is running and will go away after)

pedrolamas:
The +/- buttons are now enabled by default!

matmen (approved):

PR #787, opened 2022-07-23
feat: toolhead diagnostics

### LINKED ISSUE #454: Abilty to graph additional metrics, specificly flow rate and speed.
I think it would be very useful to be able to chart arbitrary values with basic custom annotations. For instance, I think the maximum flow rate of my printer is 16 mm3/s. A plot with both flow rate and speed on the y axis and time on the x axis with a dashed horizontal threshold marker at 16mm3/s would help me to see if print speed is causing under extrusions.

matmen:
Toolhead diagnostics charts using Klipper's `motion_report` (recording flow rate and speeds for now)
Maybe other diagnostic statistics would be useful here too, not sure which additional ones to log..

Disabled by default to save browser processing power:
image
image

Closes #454

### DIFF
--- a/src/components/layout/AppNavDrawer.vue
+++ b/src/components/layout/AppNavDrawer.vue
@@ -77,6 +77,14 @@
           </app-nav-item>
 
           <app-nav-item
+            v-if="diagnosticsEnabled"
+            icon="$chart"
+            to="/diagnostics"
+          >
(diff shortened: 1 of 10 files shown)


matmen:
Closing this in favor of the approach described in the 2nd quote in https://github.com/fluidd-core/fluidd/issues/454#issuecomment-1193257744

PR #798, opened 2022-07-27
fix: hide sub-navigation if socket not connected

pedrolamas:
When we are on Settings page and moonraker restarts (or socket gets closed) we get the "reconnect" buttons, but sub-navigation is still visible.

Before:

image

After:

image


matmen (changes requested):
The check should probably be a getter to avoid duplicate code

matmen (approved):

PR #801, opened 2022-07-28
fix: Check Bed_Mesh Support

BastelKlug:
Only shows the bed_mesh card if bed_mesh is available and klipper is ready (important for the dashboard card).

### DIFF
--- a/src/components/widgets/bedmesh/BedMeshCard.vue
+++ b/src/components/widgets/bedmesh/BedMeshCard.vue
@@ -1,5 +1,6 @@
 <template>
   <collapsable-card
+    v-if="supportsBedMesh && klippyReady"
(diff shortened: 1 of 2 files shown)


matmen (approved):

pedrolamas:
I haven't tested these yet, but I wonder if these would be better suited directly on `Dashboard.vue`, on the `filtered` method?

matmen:
True! I forgot about the filtered method..

BastelKlug:
> I haven't tested these yet, but I wonder if these would be better suited directly on `Dashboard.vue`, on the `filtered` method?

The difference is, when using the filtered method the card will be showed up in layout mode.

pedrolamas (approved):

PR #809, opened 2022-08-02
fix: Enable vue-echarts auto-resizing

matmen:
Fixes the issue where the (thermals and others) chart wouldn't be sized properly after resizing the window or switching from the settings page to the dashboard.

Before
image

After
image


pedrolamas:
That has been bothering me for a while now, didn't expect for such a clean fix!! Will test that after dinner! 🙂

pedrolamas (approved):
Perfect!!

Issue #768, opened 2022-07-16
hide backups in configuration tab


ReinBentdal:
### Is your feature request related to a problem? Please describe

the configuration tab fast gets clogged with many backup files, as seen in the screenshot
<img width="724" alt="image" src="https://user-images.githubusercontent.com/21336626/179327887-61f0ef5b-8d48-43fa-beec-cc33f852bcd3.png">

### Describe the solution you'd like

having a simple option to hide backup files. Not sure how best to implement it. At least possible to identify the number format and in that way determine if its a backup or not

### Describe alternatives you've considered

Backup files could even be its own tab under "other files"
<img width="711" alt="image" src="https://user-images.githubusercontent.com/21336626/179328016-81dc935f-ec02-47dc-9505-2ae5b6a2fcb5.png">

matmen:
There's a PR open for Klipper that allows backups to be moved to a different directory (https://github.com/Klipper3d/klipper/pull/5643). If it doesn't get merged I'm sure we can add a way to hide them :)

PR #812, opened 2022-08-04
feat(ui): Hide files in /config

BastelKlug:
Closes: #768 

Demo:

https://user-images.githubusercontent.com/44060099/182890156-92f1d41c-367d-445c-86c9-3104bb0d70c5.mov

### DIFF
--- a/src/components/widgets/filesystem/FileSystemToolbar.vue
+++ b/src/components/widgets/filesystem/FileSystemToolbar.vue
@@ -63,7 +63,7 @@
     />
 
     <file-system-filter-menu
-      v-if="!readonly && root === 'gcodes' && supportsHistoryComponent"
+      v-if="!readonly && ['gcodes', 'config'].includes(root)"
       :root="root"
       :disabled="disabled"
       @change="$emit('filter', $event)"
@@ -104,47 +104,6 @@
       />
     </div>
 
-    <v-menu
-      v-if="hideableFilesAvailable"
-      transition="slide-y-transition"
-      left
-      offset-y
-      :close-on-content-click="false"
-    >
-      <template #activator="{ on, attrs }">
-        <app-btn
-          fab
-          small
-          text
-          color=""
-          v-bind="attrs"
-          class="ml-1"
-          v-on="on"
-        >
-          <v-icon>$cogs</v-icon>
-        </app-btn>
-      </template>
-      <v-sheet
-        elevation="0"
-        class="pa-2"
-      >
-        <v-checkbox
-          v-model="showHiddenFiles"
-          :label="$t('app.file_system.label.show_hidden_files')"
-          color="primary"
-          hide-details
-          class="mx-2 mt-2"
-        />
-        <v-checkbox
-          v-model="showKlipperBackupFiles"
-          :label="$t('app.file_system.label.show_klipper_backup_files')"
-          color="primary"
-          hide-details
-          class="mx-2 mb-2"
-        />
-      </v-sheet>
-    </v-menu>
-
     <template
       v-if="roots.length > 1"
       #extension
@@ -168,6 +127,7 @@ import StatesMixin from '@/mixins/state'
 import FileSystemMenu from './FileSystemMenu.vue'
 import FileSystemFilterMenu from './FileSystemFilterMenu.vue'
 import { AppTableHeader } from '@/types'
+import { FileFilter } from '@/store/files/types'
 
 @Component({
   components: {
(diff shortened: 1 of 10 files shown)


src/components/widgets/filesystem/FileSystemToolbar.vue
```diff
@@ -194,6 +238,32 @@ export default class FileSystemToolbar extends Mixins(StatesMixin) {
     return this.roots.filter(r => this.$store.state.server.info.registered_directories.includes(r))
   }
 
+  get showKlipperBackupFiles () {
+    return this.$store.state.config.uiSettings.fileSystem.showKlipperBackupFiles
+  }
+
+  set showKlipperBackupFiles (value: boolean) {
+    console.log(value)
```
matmen, line 246:
```suggestion
```

src/components/widgets/filesystem/FileSystemToolbar.vue
```diff
@@ -194,6 +238,32 @@ export default class FileSystemToolbar extends Mixins(StatesMixin) {
...
+      server: true
+    })
+  }
+
+  get showHiddenFiles () {
+    return this.$store.state.config.uiSettings.fileSystem.showHiddenFiles
+  }
+
+  set showHiddenFiles (value: boolean) {
+    console.log(value)
```
matmen, line 259:
```suggestion
```

src/views/Configure.vue
```diff
@@ -12,6 +12,7 @@
         <file-system
           :roots="['config']"
           :max-height="816"
+          :hideable-files-available="true"
```
matmen, line 15: What's the use case for this? Seems redundant if it's hardcoded to true..

src/locales/en.yaml
```diff
@@ -83,6 +83,8 @@ app:
       uploaded: Uploaded
       view_section_documentation: >-
         View '%{section}' documentation
+      show_hidden_files: Show hidden files
+      show_klipper_backup_files: Show klipper backup files
```
matmen, line 87: While the capitalization of Klipper isn't consistent across the repo, the translations all seem to use it in upper case
```suggestion
      show_klipper_backup_files: Show Klipper backup files
```

matmen (changes requested):

src/views/Configure.vue
BastelKlug: It is to avoid unnecessary filtering in other folders. But I could change it on a simple check with the root (=== 'config')

matmen: Ahh, I see. I'm fine with either, I'll let @pedrolamas decide

pedrolamas:
There's already a `FileSystemFilterMenu` that I think we should use for this.

I've refactored the code to make this component more generic and to be able to use it in multiple roots.

I ran a few tests and seems to be working fine now, though it is not persisting the settings (something I'm not very worried, but I'm sure we can also add that in)

@BastelKlug take a look and tell me what do you think! 🙂

src/views/Configure.vue
pedrolamas: That is now gone, so should not be a problem anymore!

pedrolamas (dismissed):

matmen:
I think having the options persist like before would be much better UX wise..

Also, I think it would look much better if https://github.com/fluidd-core/fluidd/blob/38669b500717133fa2534c9cd59622249ab6d3d8/src/components/widgets/filesystem/FileSystemFilterMenu.vue#L36 was replaced with something like
```js
:two-line="filters.filter(f => f.desc).length"
```
Before
image

After
image

matmen:
Actually, I think the `two-lines` property can safely be removed, the jobs filters look fine without it:
image

pedrolamas:
> Actually, I think the `two-lines` property can safely be removed, the jobs filters look fine without it

Agreed, and I just removed it!

I will take a look at the state management now!

pedrolamas:
> I think having the options persist like before would be much better UX wise..

Added back, and also included the existing "printed" filter for the "gcodes" root! 😁

matmen:
Unfortunately this doesn't seem to work (race condition?), after a reload the files will still be shown until it's toggled off an on
image

pedrolamas:
Ah, dammit, I have a feeling this is the same issue we've experienced before with the history stuff (failure to refresh)

Back to the drawing board...

pedrolamas:
@matmen I think I've fixed the issue with the initial filter state!

matmen (approved):
LGTM

matmen (approved):
LGTM

PR #815, opened 2022-08-06
fix: adds ACCEL parameter to FORCE_MOVE commands

### LINKED ISSUE #814: FORCE_MOVE moves the printer without acceleration!
### Fluidd Version

1.19.1

### Browser

Chrome

### Device

Desktop PC

### Operating System

Windows

### What happened

Using the new `FORCE_MOVE` Fluidd feature, the toolhead moves without accelerating because the optional `ACCEL` parameter is never used:

image

According to the Klipper docs, the full syntax for the `FORCE_MOVE` command is:

`FORCE_MOVE STEPPER=<config_name> DISTANCE=<value> VELOCITY=<value> ACCEL=<value>`

This causes the toolhead to go at full speed immediately. My poor Anycubic i3 Mega didn't like it and only made noise instead of moving. Please retrieve the correct acceleration value from the printer limits and set it in the `FORCE_MOVE` commands.

Thanks!

Printhead accelerating.

### How to reproduce

Press on `FORCE_MOVE` and move the printer with the controls below.

### Additional information

_No response_

pedrolamas:
Send `ACCEL` parameter with `FORCE_MOVE` commands.

For X and Y it will use the currently set acceleration, for Z it will take the `max_z_accel` from the printer configuration

Fixes #814

### DIFF
--- a/src/components/widgets/toolhead/ToolheadMoves.vue
+++ b/src/components/widgets/toolhead/ToolheadMoves.vue
@@ -249,7 +249,10 @@ export default class ToolheadMoves extends Mixins(StateMixin, ToolheadMixin) {
       : distance
 
     if (this.forceMove) {
-      this.sendGcode(`FORCE_MOVE STEPPER=stepper_${axis} DISTANCE=${distance} VELOCITY=${rate}`)
+      const accel = (axis.toLowerCase() === 'z')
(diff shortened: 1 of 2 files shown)


BastelKlug (approved):

matmen (approved):
Tested this yesterday and worked great! :)

Issue #363, opened 2021-07-15
Gcode preview (beta) not showing whole printing area


Afterschnitzel:
Hi, 
I have V-Core3 300x300 and I have printing area 310x310mm.
I sliced object 300x200cm large in Prusaslicer.

Gcode preview is not showing whole printing (just 29x29cm) area and the left edge of the object in preview is far from the edge - in reality it is 0.5cm from the edge.

![image](https://user-images.githubusercontent.com/60582475/125836444-a6fa6974-979e-4e0b-95bc-8c356ce5a87d.png)

pedrolamas:
@Afterschnitzel any chance that you can provide us with the gcode file you used here so we can test on our side?

matmen:
Closing because of staleness. Feel free to re-open if this is still an ongoing issue.

corrado-c:
I also noticed this defect,
from me the decentralization is greater,
i think it depends a lot on how far endstop is from x0y0,.
it is done with any print.

`[stepper_x]
rotation_distance: 40  
microsteps: 32
step_pin: PF0
dir_pin: PF1
enable_pin: !PD7
endstop_pin: ^!PE5
position_endstop: -66
position_min: -66
position_max: 230
homing_speed: 40


[stepper_y]
rotation_distance: 40
microsteps: 32
step_pin: PF6
dir_pin: PF7
enable_pin: !PF2
endstop_pin: ^!PJ1
position_endstop: -43
position_min: -43`

pedrolamas:
Thanks @corrado-c, can you confirm if you are experiencing this on the latest release 1.18.0?

corrado-c:
> Thanks @corrado-c, can you confirm if you are experiencing this on the latest release 1.18.0?
yes i tried right after i updated everything,
I noticed that sometimes the zoom works, almost always does not, but the image always remains cropped

![preview1](https://user-images.githubusercontent.com/22625950/166154382-e6a639ec-abb6-45f8-a773-8c0b9bf9cd64.jpg)
![preview2](https://user-images.githubusercontent.com/22625950/166154389-d5a898be-3547-4c8c-ba37-4135f6ad7d94.jpg)
![seperslicer1](https://user-images.githubusercontent.com/22625950/166154394-2b637a63-477c-4876-a2c3-14a2b82051b4.jpg)

corrado-c:
[...]

PR #816, opened 2022-08-06
fix: Correct SVG transformations in GCode Preview

rianadon:
If a printer is not configured to have a (0,0) (xmin, ymin) coordinate, the GCode preview currently does not show up like in the picture below. For example, my printer's x goes from -113 to +113 and y from -74 to +74.

image

I fixed two issues with the SVG generation:
- The final 2 parameters of [`viewBox`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/viewBox) are the width & height of the image, not the +x and +y values.
- When flipping the image, the min & max coordinates should be added instead of subtracted. Imagine the print bed has minimum x 100 and maximum x 200. When flipped across the origin, it's going to end up far away and need to be translated 300px back.

Now, the preview shows up:
image

Signed-off-by: Ryan Adolf ryanadolf123@gmail.com

### DIFF
--- a/src/components/widgets/gcode-preview/GcodePreview.vue
+++ b/src/components/widgets/gcode-preview/GcodePreview.vue
@@ -324,8 +324,8 @@ export default class GcodePreview extends Mixins(StateMixin) {
     }
 
     const transform = [
-      this.flipX ? -(x.max - x.min) : 0,
-      this.flipY ? -(y.max - y.min) : 0
+      this.flipX ? -(x.max + x.min) : 0,
+      this.flipY ? -(y.max + y.min) : 0
     ]
 
     return `scale(${scale.join()}) translate(${transform.join()})`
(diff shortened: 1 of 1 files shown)


pedrolamas:
Thanks for this @rianadon.

Am I correct in saying that this will fix #363 ?

rianadon:
I didn't see that issue! But yes, it will fix #363.

I had the same issues with a nonzero `position_min` and the flip image settings, but those went away after the changes.

matmen (approved):
LGTM

pedrolamas (approved):

pedrolamas:
@rianadon Please add a footer to the main PR comment with a "Signed-off-by" line, following the [CONTRIBUTING](https://github.com/fluidd-core/fluidd/blob/develop/CONTRIBUTING.md) guidelines.

Thank you!

rianadon:
Sorry my bad! I've updated the description.

Issue #821, opened 2022-08-09
Internal file editor fails to load on MS Edge browser


cobyn:
### Fluidd Version

1.19.1

### Browser

Edge

### Device

Desktop PC

### Operating System

Windows

### What happened

Openeding a file in configuration shows a blank screen, looking at console log i see this error every time i open a file.
![image](https://user-images.githubusercontent.com/2914460/183764739-cdf2c3a8-1169-45b9-bcb9-cb2a0e489ab7.png)
 
This only is happening on my windows 11 machine running Chromium Edge seem to work fine on android chrome and windows 10 edge. 

### What did you expect to happen

file editor opens with the content of the file

### How to reproduce

try opening a file in configuration manager while on edge on windows 11

pedrolamas:
Thank you for reporting this, I have an idea of what this might be!

Please check the following:

- In the address bar, click the small icon that appears just left of the address
- On the popup, click the "Enhance security for this site"
- Disable the "Use enhanced security for this site"
- Refresh the page and try again.

Can you follow these steps and report back if that fixed the issue please?

cobyn:
Nice that fixed it!

pedrolamas:
Thanks for confirming it, I had suspected this was the case as Microsoft seems to have changed what non-HTTPS sites can do (and in this case, block them from running Web Assembly bits)

Guess we need to add some new instructions for Edge browser...

PR #822, opened 2022-08-10
fix: check for webassembly support

pedrolamas:
Check if WebAssembly is supported/enabled before attempting to load Monaco Editor

Mitigates #821


pedrolamas:
While this fixes the problem for when the browser doesn't have WebAssembly support (or it is disabled), we should probably add a warning to use another browser or check settings on the current one!

src/util/is-web-assembly-supported.ts
```diff
@@ -0,0 +1,16 @@
+const isWebAssemblySupported = () => {
+  try {
+    if (typeof WebAssembly === 'object' && typeof WebAssembly.instantiate === 'function') {
+      const module = new WebAssembly.Module(Uint8Array.of(0x0, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00))
+
+      if (module instanceof WebAssembly.Module) {
+        return new WebAssembly.Instance(module) instanceof WebAssembly.Instance
+      }
+    }
```
matmen, line 9: Where's this sourced from? Wouldn't a simple check for `window.WebAssembly`/`typeof WebAssembly` work here?

pedrolamas: https://stackoverflow.com/a/47880734 😉

I'm happy to use the simpler solution shown on the last comment on that solution!

matmen: That answer does a nice job explaining it. Thanks, LGTM

matmen (approved):

PR #828, opened 2022-08-11
fix: Allow uploading theme files

BastelKlug:
Add .css, .png, .jpg, .jpeg, .gif as accepted upload file types.


pedrolamas:
This is perfect, I'm just waiting for feedback on this first https://github.com/Arksine/moonraker/pull/485

If we get a "public" root available, then we might just make use of it and have the themes (and other stuff) being uploaded there instead of the config root!

matmen (approved):

pedrolamas (approved):

PR #839, opened 2022-08-20
chore: migrates "fluidd/cameras" to "webcams"

### LINKED ISSUE #788: chore: camera namespace migration
For tracking:
Currently, our webcam settings are stored in the `fluidd` namespace under the `cameras` key. This breaks part of the timelapse settings (camera selection) because the moonraker-timelapse plugin cannot find the specified camera.
We should migrate our internal fluidd cameras object to be compatible with [Moonraker's specification](https://moonraker.readthedocs.io/en/latest/web_api/#webcam-apis); as this will fix multiple issues (the above mentioned bug (https://github.com/mainsail-crew/moonraker-timelapse/issues/52) as well as conformity).

pedrolamas:
So this completely migrates our data under "fluidd/cameras" to "webcams" - **THIS IS A ONE WAY OPERATION!**

We are not using the new Moonraker Webcams API as this does not allow for extra attributes, so we are accessing the "webcams" namespace directly.

We are however using the new Webcam API to get cameras configured in the Moonraker API (we show these as read-only cameras), and to refresh our store if a camera is added/changes/removed via API.

Fixes #788


pedrolamas:
This is mostly done and working, but having said that, I had to [refactor](https://github.com/fluidd-core/fluidd/pull/839/files#diff-0d7bcaccc3b45793252745363044bb5372e356db132737d2213e69911e73dbd9R152-R167) the `MOONRAKER_DB` in globals.ts to include multiple namespaces, and it made it a bit too "verbose"...

The alternative to that could be to keep the existing `MOONRAKER_DB` as is, and add a new `MOONRAKER_DB_WEBCAMS` maybe?

matmen:
Refactoring `MOONRAKER_DB` is a good solution IMO, as it makes the entire store more dynamic for future use. I don't think it's too verbose.

I've tested this and it seems to work well (it also fixes https://github.com/mainsail-crew/moonraker-timelapse/issues/52), though your last point made me think that maybe we can handle the `socket/notifyWebcamsChanged` event to update our state (in case updates come in via another frontend)? I've uploaded the corresponding WS message here (triggered after a `POST` to `/server/webcams/item`): https://gist.github.com/matmen/6c04a6bd0a73229c42f32aa85408e19b

pedrolamas:
I've added some code so that when a `notify-webcams-changed` message is received, we will then read the cameras from the database again

pedrolamas:
> maybe we can handle the `socket/notifyWebcamsChanged` event to update our state (in case updates come in via another frontend)?

@matmen I have implemented the above, but this won't really do anything if the other frontends don't use Moonraker's new Webcam API's (I checked, and like us, they don't!)

matmen:
True.. It's just something I've noticed (because it had an unhandled event error in the console), but I guess it's good to have it for future support.
I tested this and it works well, but deleting webcams no longer seems to work (they always come back after reloading the page). I assume this is because we write to the store directly and moonraker has the old values cached?

pedrolamas:
Yep, I noticed the delete and have a fix for it ready, just cleaning the code! 😁

src/components/widgets/camera/CameraItem.vue
```diff
@@ -105,12 +105,12 @@ export default class CameraItem extends Vue {
     const config = this.camera
     let transforms = ''
 
-    if (!config.rotate) {
+    if (config.rotation === 0) {
```
matmen, line 108: `rotation` can be `undefined` too (legacy cameras)
```suggestion
    if (!config.rotation) {
```

src/components/settings/cameras/Cameras.vue
```diff
@@ -49,9 +49,19 @@
             text
             x-small
             color=""
+            :disabled="camera.source === 'config'"
             @click.stop="handleRemoveCamera(camera)"
           >
-            <v-icon color="">
+            <v-icon
```
matmen, line 55: Not sure if a title prop would be appropriate here explaining why the config is locked (maybe moving `app.timelapse.tooltip.managed_by_moonraker` to `app.general.tooltip.managed_by_moonraker` and use that?)
LMK if you want me to push this

matmen (changes requested):

src/components/widgets/camera/CameraItem.vue
pedrolamas: I made sure that doesn't happen here: https://github.com/fluidd-core/fluidd/pull/839/files#diff-11df4c839902d11995d754a99da6bcd47e702247233ccecaf07a8b455628b14aR46

src/components/settings/cameras/Cameras.vue
pedrolamas: Good point, I'll change that!

src/components/widgets/camera/CameraItem.vue
matmen: Ah! Odd, one of my cameras still had it as undefined, but I just tried it again and it worked fine :)

matmen:
Can we remove this check considering moonraker allows an arbitrary amount of cameras? https://github.com/fluidd-core/fluidd/blob/b80bdab6d849bd0fcdc51e210c80c43d682a5850/src/components/settings/cameras/Cameras.vue#L13
I think this will need support in the camera card though, so maybe we can handle that in a follow-up PR?

pedrolamas:
> Can we remove this check considering moonraker allows an arbitrary amount of cameras?

Good point, it's gone now!

src/components/settings/cameras/Cameras.vue
pedrolamas: Done!

matmen (approved):
LGTM

PR #841, opened 2022-08-21
chore: Diagnostics panel tweaks

matmen:
Marks the feature as beta and fixes two small mistakes

### DIFF
--- a/src/store/charts/mutations.ts
+++ b/src/store/charts/mutations.ts
@@ -44,7 +44,7 @@ export const mutations: MutationTree<ChartState> = {
     state[payload.type].push(payload.data)
     // console.log('set data', payload.type, payload.data)
     const firstInRange = state[payload.type].findIndex((entry: ChartData) => (Date.now() - entry.date.valueOf()) / 1000 < payload.retention)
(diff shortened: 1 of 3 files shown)


src/store/charts/mutations.ts
```diff
@@ -44,7 +44,7 @@ export const mutations: MutationTree<ChartState> = {
     state[payload.type].push(payload.data)
     // console.log('set data', payload.type, payload.data)
     const firstInRange = state[payload.type].findIndex((entry: ChartData) => (Date.now() - entry.date.valueOf()) / 1000 < payload.retention)
-    if (firstInRange) state[payload.type].splice(0, firstInRange)
+    if (firstInRange > 0) state[payload.type].splice(0, firstInRange)
```
pedrolamas, line 47: `.findIndex()` can return 0 for a match, so just wanted to double-check you want `> 0` here and not `>= 0`

pedrolamas (approved):

src/store/charts/mutations.ts
matmen: This is intentional, if the first element is in range it shouldn't be removed

Issue #464, opened 2022-01-02
Multiple printers setup not available from a different browser


gsdaemon:
Hello

It seems that on a multi-printers setup, the configured printers are saved now only inside localStorage of the browser and, when opening the Fluidd interface from another browser / device, it will only see the default printer which is returned by the apiserver located at the default moonraker address:port (127.0.0.1:7125). (Or maybe I'm doing something wrong)
![image](https://user-images.githubusercontent.com/6863250/147864564-12f6c9ce-ee1b-45e5-a238-1e2d318d5cd4.png)

Would it be possible to save the multi-printers config independently and have them available across multiple clients? Or is there maybe a workaround?

Thanks

shinuza:
The configuration is per device by design, this allows us to have a different layout on mobile and desktop for example. We could look into it but this might take some thinking to figure out which settings should be in global (moonraker) conf and which should be saved locally.

gsdaemon:
Thank you for your reply.
The "per browser" configuration definitely makes sense when it's UI related, but printer instances config imho could be considered more general/global, as they normally have nothing to do with the browser you're opening Fluidd from.

lucaschoeneberg:
Hi, we also use Fluidd for our 3D printers so that our students can observe them from the outside. The current problem is that we always have to give the students the IP addresses of the printer so that the students can get the printer in their browser. So it would be helpful if there were a setting in the fluidd config for the default printer. It is currently written to LocalStorage under: appInstances.

shinuza:
I agree this would be a good addition, but implementing this without an opt-in/opt-out would "break" standalone installation like https://app.fluidd.xyz

Skiddz:
I'm seeing the same issue as the OP, but my additional printers are not "sticking" between Fluidd sessions.  I can add printers and interact with them, but if I shut down the browser (or even just the tabe) and open a new session later, only the default printer is listed and I have to add it again.

PR #843, opened 2022-08-25
feat: Enabling default list of endpoints via config

codegrunt:
Enabling the use of the config.json to allow for a default list of endpoints to be populated on first load of the Fluidd interface.

Signed-off-by: Chris Owen <Chris@ChrisOwen.org>

### DIFF
--- a/public/config.json
+++ b/public/config.json
@@ -3,6 +3,8 @@
     "fluidd.xyz",
     "fluidd.net"
   ],
+  "endpoints": [
+  ],
   "hosted": false,
   "themePresets": [
     {
--- a/server/config.json
+++ b/server/config.json
@@ -3,7 +3,7 @@
     "fluidd.xyz",
     "fluidd.net"
   ],
-  "endpoints":[
+  "endpoints": [
   ],
   "hosted": true,
   "themePresets": [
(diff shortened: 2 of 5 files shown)


codegrunt:
I found this work-item and was looking at enabling this for myself:
https://github.com/fluidd-core/fluidd/issues/464

this primarily was for docker, and the next planned steps would be to create a "refresh: in the settings.  If this doesnt fit your vision just give me a direction and I will see if I can make those changes for you.

pedrolamas:
I like the idea here, but I'm not sure of the necessity of the `sanitizeEndpoint` method - this is an advanced feature, if users are going to use it, they better know what they are doing and set correct URLs!

codegrunt:
not saying it's necessary, and if you do not want it, easy enough to remove.  Mostly built it so that it wouldn't throw errors and stop the initialization of the webpage since it is happening in the bootstrapping of the SPA.

pedrolamas:
> Mostly built it so that it wouldn't throw errors and stop the initialization of the webpage since it is happening in the bootstrapping of the SPA.

That's a very good point and makes complete sense!

pedrolamas (approved):

matmen (approved):
LGTM

PR #846, opened 2022-08-26
docs: add custom theming documentation

matmen:
Not sure if this covers everything yet, LMK if something needs to be changed! :)
Probably would be a good idea to add some kind of README and maybe even an example theme to the themes repo, but I don't have anything ready for that yet.

image


pedrolamas (approved):
LGTM, but would be great if @BastelKlug also pitches in his review! 😁

BastelKlug (approved):
Looks self-explanatory 👍 Nice!

PR #848, opened 2022-08-27
fix(diagnostics): Metrics Explorer styling

matmen:
Before, the production build would apply the `vue-json-viewer` styles _after_ the component styles, rendering the component styles useless (this didn't happen in the dev build). This is the easiest way to solve this without changing any of the bundling logic I think.

| Before | After
| --- | --- |
| image | image |

### DIFF
--- a/src/components/widgets/diagnostics/StateExplorer.vue
+++ b/src/components/widgets/diagnostics/StateExplorer.vue
@@ -44,20 +44,20 @@ export default class StateExplorer extends Mixins(StateMixin) {
 
 <style lang="scss">
 .jv-container>.jv-code {
-  padding: 0;
+  padding: 0 !important;
 }
 
 .jv-container.jv-dark {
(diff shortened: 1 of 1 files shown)


pedrolamas:
I haven't tested this yet, but I have a feeling the more "correct" approach might be to follow this: https://www.npmjs.com/package/vue-json-viewer?activeTab=readme#theming

matmen:
I did try that before, the issue here is that it removes the `jv-light` class from the element, so we would need to include [the corresponding styling](https://github.com/chenfengjw163/vue-json-viewer/blob/master/lib/json-viewer.vue#L209-L367) ourselves, which seems redundant. I'm open to either, but just overriding the couple styles seems easier

pedrolamas:
Good point... given that I think we can go with your proposed approach at least right now!

pedrolamas (approved):

PR #862, opened 2022-09-11
feat: adjustable thumbnail size

### LINKED ISSUE #246: Bigger Job-Layout
(Copied from Discord)
How about a possibility to change the view of the print files?

You can currently see a small preview of the printed part and the details. It would be nice to have the big preview and the details right away. So everyone can choose according to their taste, whether they want a large view of their 3D prints or the already existing small view.

I agree that this creates a lot of "white" space. For some people, this can look tidier too. But this differs from person to person.

Maybe you could manage the change via a small dropdown menu to the left of the search bar. Or you can integrate the option to change the layout in the 3-point menu (right of the searchbar).

You could call it "small preview" and "large preview" - or whatever.

With my modest website programming skills, I imagine that this can be done with the least amount of effort.
But you will be able to judge that better than me. The mockup speaks for itself.

Thanks for your hard work

FluiddJobsMockup

matmen:
Closes #246

image

### DIFF
--- a/src/components/ui/AppBtnCollapseGroup.vue
+++ b/src/components/ui/AppBtnCollapseGroup.vue
@@ -18,7 +18,8 @@
     <template #activator="{ on, attrs }">
       <app-btn
         fab
-        x-small
+        :x-small="size === 'x-small'"
+        :small="size === 'small'"
         text
         color=""
         v-bind="attrs"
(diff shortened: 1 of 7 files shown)


pedrolamas:
Functionally this is perfect, I'm just not sure about the icon... maybe the `resize` one instead?

image

pedrolamas (approved):

matmen:
> I'm just not sure about the icon... maybe the `resize` one instead?

My intent here was to have this serve as a multi-item dropdown, but I think the resize icon would fit better for now.
I did notice that its line width didn't seem to match the others?
image

I think mdiImageSizeSelectLarge looks nice, but I'm fine with either one.
image

pedrolamas:
> I think mdiImageSizeSelectLarge looks nice, but I'm fine with either one.

Lets go with that one, we can move back to the "..." when we have more stuff to add! 🙂

pedrolamas (approved):

PR #864, opened 2022-09-13
feat: quick button "switch light" 

dizz74:
Quick button "light" in the top navigation.
Need select light's pin in settings.
Screenshot_32


src/components/layout/AppBar.vue
```diff
@@ -231,6 +255,21 @@ export default class AppBar extends Mixins(StateMixin, ServicesMixin) {
...
+  get topNavPinLight (): string {
+    return this.$store.state.config.uiSettings.general.topNavPinLight
+  }
+
+  isPinLightON () {
+    const pins : Array<OutputPin> = this.$store.getters['printer/getPins']
+    const lightPin = pins.find(pin => { return pin.name === this.topNavPinLight })
+    if (lightPin !== undefined) return lightPin.value === 1
+    return false
+  }
```
matmen, line 267: This should probably be a getter, and I think it can all be moved into `topNavPinLightOn` getter.

src/components/layout/AppBar.vue
```diff
@@ -254,6 +293,11 @@ export default class AppBar extends Mixins(StateMixin, ServicesMixin) {
     })
   }
 
+  handleLightToggle () {
+    const target = this.isPinLightON() ? 0 : 1
+    this.sendGcode(`SET_PIN PIN=${this.topNavPinLight} VALUE=${target}`, Waits.onSetOutputPin)
```
matmen, line 298: If `isPinLightON` gets refactored to the `topNavPinLightOn` getter, this can become the following:
```suggestion
    this.sendGcode(`SET_PIN PIN=${this.topNavPinLight} VALUE=${+!this.topNavPinLightOn}`, Waits.onSetOutputPin)
```

src/components/layout/AppBar.vue
```diff
@@ -93,7 +93,30 @@
...
+      >
+        <v-tooltip bottom>
+          <template #activator="{ on, attrs }">
+            <app-btn
+              fab
+              small
+              :elevation="0"
+              class="mr-1 bg-transparent"
+              color="transparent"
+              :disabled="!klippyReady"
```
matmen, line 107: I know we currently don't do it for the top nav power toggle, but maybe the button should be disabled when there's a wait on `onSetOutputPin` to indicate a pending state change?

matmen (commented):
Thanks. Functionality wise this LGTM (and doesn't take up too much space in the top nav), but I left some code-related comments below.
Also, I'm not 100% sure about the use case here, as the light pin can already be toggled via the Outputs card, but considering this is a QoL improvement this is fine with me. Showing it as a light icon is also fine with me, even though it can toggle an arbitrary configured pin.

src/components/layout/AppBar.vue
dizz74:
```suggestion
              :disabled="!klippyReady || hasWait($waits.onSetOutputPin)"
```
Good idea. I tried "hasWait" and I don't see disabling of the button because it's too fast, but I think it works.

src/components/layout/AppBar.vue
```diff
@@ -231,6 +255,18 @@ export default class AppBar extends Mixins(StateMixin, ServicesMixin) {
     return this.topNavPowerDevice?.status === 'on'
   }
 
+  get topNavPinLight (): string {
+    return this.$store.state.config.uiSettings.general.topNavPinLight
+  }
+
+
```
matmen, line 262: You will need to remove this extra empty line, I can't push to your `develop` branch from outside GitHub.

matmen (changes requested):

pedrolamas:
I have to admit I am struggling a bit with this one! Functionally, it is working fine and does exactly what it says, but seems a bit weird that for the Power button we are listing the Moonraker entities and there we are showing the Pins directly...

Personally, I think I would prefer for both to use the Moonraker power entities and allow users to configure them there (so this functionality would mimic the one of the existing power button)

ChaosBlades:
Consistency is king but this seems way more straight forward to set up than what the Moonraker docs show. I probably would not even bother if I had to set it up according to moonraker lmao.
https://moonraker.readthedocs.io/en/latest/configuration/#power

The chamber lights have to be hands down the most used macro so I am very much looking forward to this.

pedrolamas:
@dizz74 FYI, I've merged the changes from this PR #970 so that it has a single toggle on the appbar, and users can pick between a Moonraker Power Device or a Klipper Output Pin.

PR #874, opened 2022-09-24
feat(i18n-HU): update Hungarian language

AntoszHUN:
Hungarian language update.
 
  - Update newly added words.


pedrolamas (dismissed):

matmen (dismissed):
LGTM so far, but I'd like to keep this open in case there are more i18n-related changes before the release if that's alright

AntoszHUN:
> LGTM so far, but I'd like to keep this open in case there are more i18n-related changes before the release if that's alright

All right.

AntoszHUN:
Newer translations have been added. Let's say I have to admit that reset_layout or reset_default_layout are completely unnecessary. They all mean the same thing. I don't see the point of using both.

matmen:
Functionality wise they are not the same: the default layout can now be adjusted, so `reset_layout` resets the user-/device-specific layout to match the default layout, and `reset_default_layout` resets the default layout to fluidd's global defaults.

AntoszHUN:
> 



> Functionality wise they are not the same: the default layout can now be adjusted, so `reset_layout` resets the user-/device-specific layout to match the default layout, and `reset_default_layout` resets the default layout to fluidd's global defaults.

I understand. After all, everything can be explained. I never needed it. I don't think many people use it.

AntoszHUN:
I updated the Hungarian language.
  - New words have been added
  - I translated the newly added words

PR #876, opened 2022-09-25
fix: timelapse explorer images not loading with forced_logins

### LINKED ISSUE #849: timelapse snapshot image
### Fluidd Version

v1.19.1-3f99573

### Browser

Chrome, Safari

### Device

Desktop PC, Laptop

### Operating System

macOS, iOS

### What happened

On the timelapse 'tab' the preview image in the 'timelapse status' field doesn't load _when_ user authentication is used  (e.g. `force_logins: true`) instead of running fluidd on a trusted_client. 

The cause seems to be an authentication issue.

Screenshot 2022-08-28 at 09 42 18

I expect an image with the current snapshot to be shown as an authenticated user. I presume an authorization token needs to be passed along from fluidd.

### How to reproduce

Use `force_logins: true` in the moonraker `[authorization]` section and the preview doesn't show.

### Additional information

Here is a moonraker logs example of the authentication failure:

```
moonraker.log:2022-08-28 09:33:58,810 [web.py:log_exception()] - 401 GET /server/files/timelapse_frames/frame000008.jpg (192.168.178.220): Unauthorized
moonraker.log:2022-08-28 09:33:58,813 [app.py:log_request()] - 401 GET /server/files/timelapse_frames/frame000008.jpg (192.168.178.220) [No User] 12.31ms

```

matmen:
This fixes the issue where taken frames would not load when `forced_logins` was turned on in the moonraker config.

This doesn't yet fix the issue where video thumbnails won't load because this will require some rework of the file browser.
In theory this should be relatively simple, though I could not get it working reliably in my previous attempts.
If anyone wants to take a shot at this, feel free to :)

This also introduces a change in `createFileUrl` so that oneshot tokens are only appended to the URL when authorization is actually required.

Closes #849 https://github.com/mainsail-crew/moonraker-timelapse/issues/56


src/components/widgets/timelapse/TimelapseStatusCard.vue
```diff
@@ -113,22 +120,16 @@ export default class StatusCard extends Mixins(StateMixin) {
     this.selectedFrameNumber = value === this.frameCount ? 0 : value
   }
 
-  get previewUrl () {
-    const url = new URL(this.apiUrl ?? document.location.origin)
-
+  async updatePreviewUrl () {
     if (this.lastFrame && this.lastFrame?.file) {
```
pedrolamas, line 124:
```suggestion
    if (this.lastFrame?.file) {
```

src/mixins/files.ts
```diff
@@ -148,11 +148,18 @@ export default class FilesMixin extends Vue {
    * @returns The url for the requested file
    */
   async createFileUrl (filename: string, path: string) {
-    const token = (await authApi.getOneShot()).data.result
-
     const filepath = (path) ? `${path}/${filename}` : `${filename}`
+    let url = `${this.apiUrl}/server/files/${filepath}?date=${Date.now()}`
+
+    const forceLogins = this.$store.getters['server/getConfig'].authorization.force_logins
+    const isTrusted = this.$store.getters['auth/getCurrentUser']?.username === '_TRUSTED_USER_'
```
pedrolamas, line 155: We also need to check for `_API__API_KEY_USER_USER_` here? Seems we do in a few places elsewhere in the code!

matmen: Hmm, I don't immediately see how the frontend could be logged in with the API key user.. In any case, this should be `_API_KEY_USER_` instead of `_API__API_KEY_USER_USER_`, so the existing code is wrong too

pedrolamas: Indeed, I'm not sure either why the API user shows on those other places... shouldn't be a big issue to ignore it here!

pedrolamas:
@matmen is this one ready for review now?

matmen:
Not quite yet, still need to figure out how to asynchronously load thumbnails in the file browser. I'll take another look later, but feel free to take a shot at it if you have some spare time ;)

matmen:
I haven't figured out a good way to handle the async loading of thumbnail URLs yet, so I am just replacing the thumbnails with the file icon. I'll revisit this in a future release.

pedrolamas (approved):

PR #878, opened 2022-09-26
feat: user/device specific layouts

### LINKED ISSUE #370: device specific layout
This is a feature request not a problem.

At the moment if I edit the desktop layout this gets used for every device I view it on, it would be nice to have a "local" view that would be used by just the device I am viewing it on. I use a tablet as a monitor next to my printer and I do not need to see every gadget eg Camera is not much when I am actually next to the printer, however if viewing remotely it would be required. Same for some of the other components of the desktop.

An alternative would be a layout manager where I can create and store layouts and then select which one to use on each device I view it on.

Thank you

### LINKED ISSUE #410: Feature request: User specific layouts
I use fluidd on both on a phone right next to my printer and my desktop. I don't want the camera and some other things visible on the phone. I created users expecting the layout to be specific to a user but updating the layout as `User1` also updated it for `User2`

`User1` and `User2` have independent layouts when they log in. 

Layouts that are per device, maybe the settings could be saved on each machine.

I'm not sure who would benefit from this. I imagine its a weird edge case thing for _some_ people with multiple devices or for multi-user setups.

matmen:
Creates user and device specific layouts on layout change (based on concatenation of `dashboard-{user}-{device-size}`, clones current `dashboard` if no custom one exists yet).

Allows the overriding of the default layout for new users (or other devices), which can also be reset to the initial state.
Not 100% sure about the placement and usage of header buttons here, but it's what was there before:
image

Closes #370
Closes #410

### DIFF
--- a/src/components/layout/AppBar.vue
+++ b/src/components/layout/AppBar.vue
@@ -138,6 +138,22 @@
         @click.stop="handleResetLayout"
         v-html="$t('app.general.btn.reset_layout')"
       />
+      <template v-if="isDashboard">
+        <app-btn
+          small
+          class="mx-2 ml-8"
+          color="primary"
+          @click.stop="handleSetDefaultLayout"
+          v-html="$t('app.general.btn.set_default_layout')"
+        />
+        <app-btn
+          small
+          class="mx-2"
+          color="primary"
+          @click.stop="handleResetDefaultLayout"
+          v-html="$t('app.general.btn.reset_default_layout')"
+        />
+      </template>
     </template>
 
     <user-password-dialog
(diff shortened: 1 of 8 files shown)


src/components/layout/AppBar.vue
```diff
@@ -240,17 +260,45 @@ export default class AppBar extends Mixins(StateMixin, ServicesMixin) {
...
     const pathLayouts = {
       '/diagnostics': 'diagnostics'
-    } as { [key: string]: string }
+    } as Record<string, string>
+
+    const pathLayout = pathLayouts[this.$route.path]
+    let layoutDefaultState
+    if (pathLayout) {
+      // reset to default init state
+      layoutDefaultState = defaultState().layouts[pathLayouts[this.$route.path]]
```
pedrolamas, line 276:
```suggestion
      layoutDefaultState = defaultState().layouts[pathLayout]
```

src/views/Dashboard.vue
```diff
@@ -131,7 +131,7 @@ export default class Dashboard extends Mixins(StateMixin) {
   }
 
   get layout () {
-    return this.$store.getters['layout/getLayout']('dashboard')
+    return this.$store.getters['layout/getLayout'](this.$store.getters['layout/getSpecificLayout'])
```
pedrolamas, line 134:
```suggestion
    const currentLayoutName = this.$store.getters['layout/getSpecificLayout']
    return this.$store.getters['layout/getLayout'](currentLayoutName)
```

src/store/layout/getters.ts
```diff
@@ -53,5 +56,13 @@ export const getters: GetterTree<LayoutState, RootState> = {
         if (config) return { ...config }
       }
     }
+  },
+
+  getSpecificLayout: (state, getters, rootState, rootGetters): string => {
```
pedrolamas, line 61: This returns the layout name, not the layout, so maybe rename to a more appropriate `getSpecificLayoutName` instead?

pedrolamas (changes requested):
Working fine, but I think that getter should be renamed to make clear it returns the layout name and not the actual layout objects!

pedrolamas (approved):

PR #887, opened 2022-10-03
fix: FileSystemBrowser icon size and resolution

pedrolamas:
Before the changes on 0a3f8de1a0e71c01cc2bf54a6ad3985d88369c32, the FileSystemBrowser icon size was 16px, but this was changed to a minimum (default) of 32px.

This caused an UI change, but more importantly, the icons on retina style displays now show with dithering (as the system is scaling a 32px raster image to 64px)

Before the changes on 0a3f8de1a0e71c01cc2bf54a6ad3985d88369c32:

image

After the changes on 0a3f8de1a0e71c01cc2bf54a6ad3985d88369c32 (shows dithering as the 32px raster image is rendered at 200%):

image

After the changes on this PR (32px raster image shown as 200% 16px icon):

image

Also part of the changes on this PR, the 32px icon rendered with no dithering as it is using the large thumbnail:

image

Signed-off-by: Pedro Lamas <pedrolamas@gmail.com>

PR #888, opened 2022-10-03
fix: make FileSystemBrowser icons size consistent

### LINKED ISSUE #885: When increasing thumbnail size also increase Folder Icon size to match

When you increase the Gcode Thumbnail size the Folder Icon stays the same size which makes things look inconsistent.

Increase the Folder Icon to the same size as the Gcode Thumbnail.

Screenshot from 2022-10-03 01-43-41
Don't forget about the up directory icon either.
Screenshot from 2022-10-03 01-46-12

Could re-work folders to be tiled instead of a list (2-3 wide). Make them look more like buttons. If only 1 folder then center it to break it up from the list of gcode files. But that requires more effort than I think is worth it. But I can't code so, you do you :)

### Additional information

_No response_

pedrolamas:
Keep all FileSystemBrowser icons size consistent (so when we increase the size of the thumbnails, we also consistently increase the size of the folders and file icons)

Note: this should be merged **after** #887 so that 16px icons can be shown correctly!

image

Fixes #885 


matmen (approved):
Personally prefer the "compact" view more, but this is fine with me and works well :)

ChaosBlades:
> Personally prefer the "compact" view more, but this is fine with me and works well :)

I put in this request but I agree. I think the proper solution to have the best of both worlds would be to break out the folders into a tile view that is 2 wide while keeping the gcode files in a list view. I mean what do you need the details for anyways? For example the folders file size is 4.0 KB. That is not the size of the combined gcode within and most other details are just -- anyways.

matmen:
I think we're going to move forward with this solution for now, but we might revisit it in the future. I'll go ahead and open a new tracking issue for splitting the file browser into a directory and a file side. I can already see this opening up new problems though, for example because of the limited width in the dashboard. I much prefer the current solution, as it presents all files and folders in a more compact view with more information visible immediately.

matmen (approved):

pedrolamas:
I got to admit, I'm not sure I want to merge this one down... 😅

matmen:
I have to agree here, I much prefer the more condensed view. IMO there's no point in having huge file icons if they don't add any usability.. If we decide to implement #892, this "issue" will be solved anyway (though I'm not sure that implementation makes sense either. It might for the fullscreen jobs view, but IMO it wastes too much space on the dashboard..)

pedrolamas:
Fully agreed. And yes, dashboard show be as is (even the icon size there doesn't make much sense IMHO, not to mention we are showing a min of 32px when in reality they are always half of the selected size...)

Closing this now.

PR #891, opened 2022-10-05
feat: larger thumbnail preview

### LINKED ISSUE #884: Preview Larger Gcode Thumbnail

Even with the newly added adjustable thumbnail size I would still like to 'go further beyond'. Would be nice for large print beds with lots of tiny parts. So you can see what parts specifically are on the plate.

Right-Click on Job > Preview Thumbnail. 
Opens a window in the center of the screen with the full resolution thumbnail.

Could also make the tiny thumbnail a button when pressed does the same thing instead of right-click menu.

Or hovering over the thumbnail generates a tooltip with the full resolution thumbnail.

### Additional information

Should not need to worry about the resolution of the thumbnail. In SuperSlicer at least you can specify the resolution. I have tested with 200x200 for small and 1200x900 for large and it seems to work fine. Just can't utilize it.

matmen:
Not sure about the practicality on this one, considering the default thumbnail size configured in most slicers is still not very large.
As already explained by @ChaosBlades in #884 this can of course be configured, though.
I think this can still be useful considering opening thumbnails in a new tab just downloads them.
Maybe the styling (especially width) of the preview modal still needs to be adjusted to be able to show huge thumbnails?

Closes #884


matmen:
welp, based it off of the wrong branch..

pedrolamas (approved):

ChaosBlades:
SuperSlicer > Printer Settings > Thumbnails > 200x200 for small and 1200x900 for large
Screenshot from 2022-10-25 13-14-53

Works great!

PR #893, opened 2022-10-05
feat: option to beautify file names

### LINKED ISSUE #886: Option to hide '.gcode' and '_' from file names in Jobs Menu

A super duper important feature request.

Settings > General
Add option to 'Hide extensions and underscores from Jobs Menu.

Removes `.gcode` from end of file name and replaces `_` with ` `.

Just remove them by default. I see no reason they should exist other than it requires effort to code them out.

### Additional information

Can you tell I'm OCD?
It just makes the best looking printer WebUI look just a little bit cleaner.

matmen:
Adds an option to display beautified file names in the file explorer (removed extension, replaces underscores with spaces).
Not entirely sure if this should use normal spaces or non-breaking ones (I'd personally go for non-breaking ones as it keeps the previous linebreak style):

Before
image

Normal Spaces
image

Non-Breaking Spaces
image

Closes #886


matmen:
Eh. Not sure if this should instead be a general setting that also beautifies the file names in the gcode preview and file preview?

pedrolamas:
I have to admit, I'm not entirely convinced about this one... if users want "pretty names", they can just write the filename as such!

I personally like to see the extensions, and as we are not showing specific icons or file types, I think we should not hide these either.

matmen:
Not a _huge_ fan of this myself, it just seemed like something simple to implement.
It seems like the side effects (mismatching file names in multiple views, e.g. when renaming files) outweighs the benefit here, though.

ChaosBlades:
> I have to admit, I'm not entirely convinced about this one... if users want "pretty names", they can just write the filename as such!
> 
> I personally like to see the extensions, and as we are not showing specific icons or file types, I think we should not hide these either.

Curious, what other file types would even be found in the file browser other than `.gcode`?

PR #898, opened 2022-10-07
fix: Exclude Object not reacting to mobile inputs

### LINKED ISSUE #897: Allow the exclude/cancel object feature to be clickable on mobile

Just want to start by saying thank you for Fluidd and for the exclude object functionality. It was the main thing I was missing from the OctoPrint days. Just in the past month, you've saved a few hundred grams of filament from the landfill for me alone. 

The problem I'm having is that I almost always access my printers from my phone. I can view the names of the objects and cancel from there, but when I'm printing 6 of the same thing, and one of them fails, there's really no way to know which one I should cancel based on the name, since they all have the same name. I can easily see on the gcode preview which one I need to cancel, but I can't tap on that.

I might be missing something, so please let me know if I am, but I just couldn't figure out how to do that from mobile. 

Allow tapping on the the gcode viewer cancel object when using a mobile browser. 

I've tried to cancel based on filename, but I don't see any way to relate object name to the spot on the bed that failed. 

### Additional information

_No response_

matmen:
Closes #897

### DIFF
--- a/src/components/widgets/exclude-objects/ExcludeObjects.vue
+++ b/src/components/widgets/exclude-objects/ExcludeObjects.vue
@@ -29,6 +29,9 @@
         :d="iconCircle"
         class="hitarea"
         @click="$emit('cancel', name)"
+        @touchstart="touchedElement = name"
+        @touchend="handleTouchEnd(name)"
+        @touchcancel="touchedElement = undefined"
       />
     </svg>
   </g>
@@ -41,6 +44,13 @@ import { Icons } from '@/globals'
 
 @Component({})
 export default class ExcludeObjects extends Mixins(StateMixin) {
+  touchedElement?: string
+
+  handleTouchEnd (name: string) {
+    if (this.touchedElement === name) this.$emit('cancel', name)
+    this.touchedElement = undefined
+  }
+
   get parts () {
     const parts = this.$store.getters['parts/getParts']
     return Object.keys(parts)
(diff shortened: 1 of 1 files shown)


src/components/widgets/exclude-objects/ExcludeObjects.vue
```diff
@@ -29,6 +29,9 @@
         :d="iconCircle"
         class="hitarea"
         @click="$emit('cancel', name)"
+        @touchstart="touchedElement = name"
+        @touchend="handleTouchEnd(name)"
+        @touchcancel="touchedElement = undefined"
```
pedrolamas, line 34: Is it worth also adding `@touchmove="touchedElement = undefined"` so that if moving the canvas it won't raise the cancel dialog?

pedrolamas (approved):

simpat1zq:
Hello. I have no idea if this fix was included in the latest release, but if it was, I still don't think it's working on mobile. I tried from the Samsung Browser, and Chrome Browser on Android 12.

pedrolamas:
Hi @simpat1zq , this was indeed included in the 1.21 release... can you double check on the bottom of the page (just scroll all the way down) that you are actually using this version? Thanks!

simpat1zq:
Hey. So I went to take a screen recording, and tapped in it a bunch of times, and after about 10 taps, it finally worked. I think that it might just be REALLY sensitive to moving the gcode image around instead of detecting the tap. I even zoomed in a lot so the cancel button was huge to make sure I was actually hitting that cancel icon, but I just can't get it to register consistently. But it does eventually register the tap, so I think I'm all set, but if there is a super easy fix to make it register a tap instead of a swipe a little more easily, you may want to consider it. But I'm a happy camper now. 

Thank you so much to you all. The exclude object functionality has already saved a few hundred grams of pla from my local landfill.

PR #906, opened 2022-10-11
feat: adds missing translation keys

pedrolamas:


src/components/ui/AppCardDrag.vue
```diff
@@ -3,7 +3,7 @@
     <v-card-title
       class="card-title card-heading py-1 draggable"
     >
-      Drag Me
+      {{ $t('app.general.label.drag_me') }}
```
matmen, line 6: I don't see this file being imported anywhere - maybe we should just remove it?

src/components/widgets/filesystem/FileSystemContextMenu.vue
```diff
@@ -27,9 +27,7 @@
               <v-list-item-icon>
                 <v-icon>$printer</v-icon>
               </v-list-item-icon>
-              <v-list-item-title>
-                Print
-              </v-list-item-title>
+              <v-list-item-title> {{ $t('app.general.btn.print') }}</v-list-item-title>
```
matmen, line 30:
```suggestion
              <v-list-item-title>{{ $t('app.general.btn.print') }}</v-list-item-title>
```

matmen (changes requested):

src/components/ui/AppCardDrag.vue
pedrolamas: Very good point there, I was actually surprised as I didn't remember seeing "drag me" string before... I'll remove the file and translation!

pedrolamas: Done!

matmen (approved):

PR #911, opened 2022-10-13
fix: disable controls if Klipper not ready

pedrolamas:
Disables the following controls if `klipperReady` is false:

- `Print` and `Preheat` on the file context menu
- All the panels in the "Tune" page

### DIFF
--- a/src/components/widgets/filesystem/FileSystemContextMenu.vue
+++ b/src/components/widgets/filesystem/FileSystemContextMenu.vue
@@ -20,20 +20,26 @@
             <v-list-item
               v-if="canPrint"
               link
+              :disabled="!printerReady"
               @click="$emit('print', file)"
             >
               <v-list-item-icon>
-                <v-icon>$printer</v-icon>
+                <v-icon :disabled="!printerReady">
+                  $printer
+                </v-icon>
               </v-list-item-icon>
               <v-list-item-title>{{ $t('app.general.btn.print') }}</v-list-item-title>
             </v-list-item>
             <v-list-item
               v-if="canPreheat"
               link
+              :disabled="!printerReady"
               @click="$emit('preheat', file)"
             >
               <v-list-item-icon>
-                <v-icon>$fire</v-icon>
+                <v-icon :disabled="!printerReady">
+                  $fire
+                </v-icon>
               </v-list-item-icon>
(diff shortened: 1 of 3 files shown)


src/components/widgets/filesystem/FileSystemContextMenu.vue
```diff
@@ -16,11 +16,9 @@
         no-gutters
       >
         <v-col>
-          <v-list
-            dense
-          >
+          <v-list dense>
             <v-list-item
-              v-if="file.type !== 'directory' && rootProperties.accepts.includes('.' + file.extension) && rootProperties.canPrint"
+              v-if="canPrint"
```
matmen, line 21: Not sure what the best behavior is here, but we can also just disable this if Klippy isn't ready instead of hiding it

src/views/Tune.vue
```diff
@@ -12,7 +15,7 @@
       md="4"
     >
       <bed-mesh-controls
-        v-if="supportsBedMesh && klippyReady"
+        v-if="supportsBedMesh"
         class="mb-2 mb-sm-4"
       />
       <end-stops-card class="mb-2 mb-sm-4" />
```
matmen, line 21: The query button should probably be disabled when Klippy isn't ready, too

src/components/widgets/filesystem/FileSystemContextMenu.vue
pedrolamas: Indeed, only reason I opted for hiding is because we were doing the same for pre-heat! Having said that, it makes sense just to disable, so will change!

src/views/Tune.vue
pedrolamas: No need as we are hiding all cards (hence why I removed it from here)

src/components/widgets/filesystem/FileSystemContextMenu.vue
pedrolamas: Not sure "disabled" state is that visible... what you think?
image

matmen: It looks fine IMO, but up to you ;)

pedrolamas: Updated to match the above!

matmen (approved):

PR #917, opened 2022-10-18
fix: sort the manual probe offsets descending

### LINKED ISSUE #916: "manual probe" order of items
### Fluidd Version

1.20.1

### Browser

Firefox

### Device

Desktop PC

### Operating System

Windows

### What happened

the order of z moves in "manual probe" windows is inconsistent, it is not sorted from big to small (or small to big), which is not very intuitive.
This feature was added in https://github.com/fluidd-core/fluidd/pull/785

showing the items in order from biggest to smallest. As in my nice visualisation in the attached screenshot

or change items to:
1
0.5
0.1
0.05
0.01
0.005
then it would be each half of the previous value

### How to reproduce

just open the "manual probe" screen and see the interesting order :)

### Additional information

fluidd_manualprobe

pedrolamas:
Fixes #916 

### DIFF
--- a/src/components/common/ManualProbeDialog.vue
+++ b/src/components/common/ManualProbeDialog.vue
@@ -173,7 +173,7 @@ export default class ManualProbeDialog extends Mixins(StateMixin, ToolheadMixin)
       1,
       0.1,
       ...this.$store.state.config.uiSettings.general.zAdjustDistances
-    ]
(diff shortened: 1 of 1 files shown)


BastelKlug (approved):
👍

matmen (approved):

PR #919, opened 2022-10-19
chore: dependency updates

pedrolamas:
This is mostly to bring `vue-inline-svg` as I sent them a PR with some new TypeScript types and they have merged and released an update for it, so now we don't need the "vue-inline-svg.d.ts" file anymore!


BastelKlug (approved):
Tested and nothing negative found 👍

matmen (approved):

PR #921, opened 2022-10-19
fix(console): console no longer loses scroll attachment on multiple M118

### LINKED ISSUE #877: Add Forced Auto Scroll to resolve issues with windows loosing auto scroll

All items that feature auto scroll constantly loose scroll and you have to scroll back down to get it caught back up. Either the scroll simply stops or it looses position and and isn't auto scrolling at the actual end of the log. There is also the situation where you are trying to scroll the entire page and your mouse lands on a window that auto scrolls and you scroll up accidentally.

Add a forced auto scroll where even if you scroll up it will immediatly upon the next update scroll back down to the bottom unless you click a check box to disable auto scroll.

Fix the current implementation of auto scroll so it does not loose position

### Additional information

Here is a screenshot of TrueNAS SCALE which has the feature I am proposing. You can scroll up and it will stay unless a new update happens in the log at which point it will scroll back down to the bottom unless you check that box.
Screenshot from 2022-09-25 17-46-45

### LINKED ISSUE #904: Auto-scrolling to bottom of console breaks with multiple M118.
### Fluidd Version

1.7.0-f1b32a9

### Browser

Chrome, Firefox

### Device

Laptop, Mobile device (Phone, Tablet or similar)

### Operating System

Windows, Linux

### What happened

When a macro emits multiple M118 commands in rapid succession (or a plugin prints data with multiple lines) the console scroll no longer sticks to the bottom and no longer follows output.

I expect the console to continue remaining "stuck" to the newest output.

### How to reproduce

Define the macro:
```
[gcode_macro TEST_MULTILINE_M118]
gcode:
    M118 LINE 1
    M118 LINE 2
    INVALID_COMMAND
    M118 LINE 3
    M118 LINE 4
```

Run the command `TEST_MULTILINE_M118` repeatedly in the console until it fills the console view. The next invocation of `TEST_MULTILINE_M118` will stop scrolling partway through the output.

(This was first observed with the automatic z-calibration Klipper plugin, and can be seen there.)

### Additional information

Looking into `src/components/widgets/console/Console.vue`, the issue seems to be the `      this.updateScrollingPaused()` at the end of `scrollToLatest`.

Control flow only reaches that call if scrolling is not already paused, so there is no need to call `this.updateScrollingPaused()` to check if scrolling should be paused or unpaused after scrolling to the bottom. Commenting this one line out fixes the issue and does not appear to cause any problems.

Let me know if you'd like me to submit a pull request for this one-liner.

gauravmm:
Only recheck attachment state if `scrollToLatest` is forced, avoids weird bugs where the console trace detaches on multi-line M118s.

Resolves #877, resolves #904, is more carefully thought through than my proposal in #904.

Signed-off-by: Gaurav Manek gaurav@gauravmanek.com


pedrolamas (approved):

matmen (approved):
LGTM, thanks

PR #936, opened 2022-11-03
Update Hungarian Language

AntoszHUN:
I updated the newly added word.

  - Add a word based on the Japanese language


pedrolamas:
Thank you @AntoszHUN, but in this particular case, it's actually the Japanese translation that is incorrect, so I need to remove that entry from it...

The en.yaml is the base language, so in the future I recommend that you only use that one too.

PR #957, opened 2022-11-21
fix: sort temperature sensors by type and name

### LINKED ISSUE #956: Sorting Sensor names in list of Thermals

Taking my Voron 2.4 to the edge of the possible searching to enlarge the envelope of possibilities, I need several more sensor of which some are temperature sensors. Adding more temperature sensors creates the need for ordering the list of sensors presented in the GUI of fluidd. As can be seen on the attached image, the sensors are perhaps ordered in groups by type (heaters, temperature_fans, sensors), but within group, sensors seems to be a listed ordered by some internal hash value of the string name of the sensor rather than the string name of the sensor it self. Ideally one should be able to set the order of the list of sensors or at least have the list of sensors within a group listed alphabetically in both the "Thermal" window as well as the popup window within "Thermals". As it is now, the "Thermals" list and the popup within the Thermals do not use the same order of listing.
Screen Shot 2022-11-18 at 9 11 38 PM
[klippy(1).log](https://github.com/fluidd-core/fluidd/files/10044528/klippy.1.log)

By default sort each group of thermals alphabetically.

Well when you need a sorted list you need a sorted list. You want a sorted list sorted in a maningful way, when you have more tha 2-3 items on the list in total. Alterntive to sort is to exclude items on list, but that makes no sense here.

### Additional information

Files are attached in previous fields

pedrolamas:
Currently the thermals are grouped by heaters on top, then temperature fan, and finally temperature sensors, but these are unordered inside these groups.

This PR ensures that we still keep the above grouping, and then inside each group we sort by sensor type, and finally name.

image

Fixes #956 

### DIFF
--- a/src/store/printer/getters.ts
+++ b/src/store/printer/getters.ts
@@ -549,6 +549,7 @@ export const getters: GetterTree<PrinterState, RootState> = {
       }
     }
     return pins
+      .sort((a, b) => a.type.localeCompare(b.type) || a.name.localeCompare(b.name))
   },
 
   /**
@@ -601,6 +602,7 @@ export const getters: GetterTree<PrinterState, RootState> = {
     }, {} as Record<string, Sensor>)
 
     return Object.values(sensors)
+      .sort((a, b) => a.type.localeCompare(b.type) || a.name.localeCompare(b.name))
   },
 
   /**


matmen (approved):

pmaersk:
Hi Pedro.

It fixes it mostly. The list is sorted alphabetically groupwise in the main thermals window, but the pop-up window is not. The pop-up should most likely have the same order as the main window with the greyed out entries not shown.

image

pmaersk:
Also in the main window the name is parsed for special characters replaced by spaces and space added in front of capital letters within the name (not the first letter I assume). You may want to parse the name in the pop-up the same way and keep the same group wise order (with the greyed-out entries removed).

PR #958, opened 2022-11-21
fix: properly scale print finish ETA

matmen:
Fixes an issue where the print finish ETA would show a full date in 1970 instead of the proper time


pedrolamas (approved):

pedrolamas:
Nice catch, guess I missed one! 🙂

PR #961, opened 2022-11-22
fix: extrusion estimation spacing

matmen:
Removes spaces leading commas:

Before
image

After
image


pedrolamas (approved):
Got to admit, that was troubling me too... Thanks for fixing it! 🙂

PR #963, opened 2022-11-25
chore: check for circular references on build

pedrolamas:
Ideally, one should not have circular-references between files and though that was never an issue before, when we moved to Vite it became one as it breaks HMR!

I've refactored the code to ensure we don't have any more circular-references, and this PR includes the final bits on that.

This also adds a new circular-reference check that will run on push so that we can ensure these don't make their way back into the codebase.


matmen (approved):

package.json
```diff
@@ -28,6 +28,7 @@
     "test": "npm run test:unit",
     "test:unit": "vitest --environment jsdom",
     "type-check": "vue-tsc --noEmit -p tsconfig.vitest.json --composite false",
+    "circular-check": "tsc --showConfig -p ./tsconfig.app.json > temp.json && madge --ts-config ./temp.json --warning --circular --extensions ts,vue,js ./src",
```
matmen, line 31: `temp.json` isn't deleted after completion, but I assume that's fine as it's only run in CI

pedrolamas: Indeed, I noticed that too, but didn't care much as this is just for CI!

PR #969, opened 2022-11-29
feat: g-code preview alternative controls

pedrolamas:


matmen (approved):
Nice! I tested the UX on mobile and it seems fine to me (not too small).
I did however also notice that #898 didn't actually fix the gcode preview behavior, so I will need to revisit that on the weekend..

PR #982, opened 2022-12-17
feat(i18n-HU): update Hungarian language

AntoszHUN:
I have updated the newly added options.
  - Translation of new menu items
  - Removal of old menu items based on the English version

Please let me know more errors and I will fix them as soon as I can.


src/locales/hu.yaml
```diff
@@ -101,6 +101,7 @@ app:
       add_file: Fájl hozzáadása
       command_palette: Parancs paletta
       download_file: Fájl lekérése
+      keyboard_shortcuts: Billentyűparancs
```
pedrolamas, line 104: I think this one has been removed from the en.yaml file a while ago, so it is safe to remove it from here too.

src/locales/hu.yaml
```diff
@@ -613,6 +619,25 @@ app:
...
+        becsült térfogatáram %{estimatedVolumetricFlow} mm³/s
+      stats_extruded_length: >-
+        A filament extrudálása %{extrudeLength} mm-nél
+        %{extrudeFactor} % folyamot<span class="secondary--text">,
+        biztosít egy</span>-t
+        becsült kimeneti hosszon %{estimatedExtrudedLength} mm
+      stats_max_speed: >-
+        <span class="secondary--text">Feltéve, hogy a</span>
+        %{layerHeight} mm réteg magassága<span class="secondary--text">,
+        a</span> becsült maximális nyomtatási sebesség %{estimatedMaxSpeed} mm/s
```
pedrolamas, line 640: Just wanted to double-check that these are appearing correctly:

image

I found it odd that some of the highlighted areas (marked by the `<span>` tabs above) are showing mid word, so wanted to check with you first

pedrolamas (changes requested):

src/locales/hu.yaml
AntoszHUN: I tried to translate correctly. The location of the words in this context is unusual, because I have not dealt with such a translation part before.

matmen (approved):

PR #986, opened 2022-12-19
feat: allow Fluidd to run fully offline

pedrolamas:
After the first load, Fluidd will be able to run completely offline!

What this means is that even if the host device where Fluidd was installed to goes offline, users will still be able to open Fluidd and connect to Moonraker (assuming it can be reached!)

This is by taking full advantage of the ServiceWorker cache features, so that all files are cached, including the "config.json" file which is the only dynamic file we have (so users can still change values on this file, click "refresh", and Fluidd will use the latest and cache the file for later offline work)

These changes will also improve the loading speed of Fluidd as most of the files required to run it will always be cached in the browser!

On first load, Fluidd will quickly show a flash message indicating that it is ready to run offline:

image

On subsequent releases and updates of Fluidd, when the user loads Fluidd for the first time and cached files are updated, we will show a message on the bottom right corner indicating as such:

image

At this stage the user can click "Reload" or hit F5 and Fluidd will be fully updated!

**Dev Note:** under dev this can be quite tricky to test, so my advice is to do an `npm run build` and then `npm run serve:prod`, load Fluidd, and once you see the indication of "ready for offline", stop the serve and hit F5 to see it hit the cache!

### DIFF
--- a/components.d.ts
+++ b/components.d.ts
@@ -43,6 +43,7 @@ declare module 'vue' {
     FlashMessage: typeof import('./src/components/common/FlashMessage.vue')['default']
     KlippyStatusCard: typeof import('./src/components/common/KlippyStatusCard.vue')['default']
     ManualProbeDialog: typeof import('./src/components/common/ManualProbeDialog.vue')['default']
+    RegisterServiceWorker: typeof import('./src/components/common/RegisterServiceWorker.vue')['default']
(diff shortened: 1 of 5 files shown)


pedrolamas:
@Desuuuu this might interest you as it is a continuation of #985

matmen (approved):

PR #1007, opened 2023-01-10
feat: adds support to create ZIP archives

pedrolamas:
Adds support to Moonraker's file zip endpoint, allowing to create ZIP archives from files and directories.

Requires Moonraker API v1.1 or above, it will not show for other versions.

## File context menu

image

## Bulk actions on multi-select

image

## ZIP Archive example

image

## Context menu on ZIP Archive

image


pedrolamas:
Not sure exactly what happened with this PR, but only now noticed that it was showing the incorrect title (and had an extra, already merged commit)... this PR is to add support to create ZIP archives, as per Moonraker support.

PR #1011, opened 2023-01-15
feat: Crowsnest camera support

matmen:
* Fixes support for opening Crowsnest-served webcams as fullscreen
* Adds auto-detection for crowsnest-served cameras
* Adds reachability check on URL change

Not sure if the new v-switch is necessary here, in theory auto detection of mjpeg-streamer vs crowsnest would work, but IMO setting it manually is cleaner.

### DIFF
--- a/src/components/settings/cameras/CameraConfigDialog.vue
+++ b/src/components/settings/cameras/CameraConfigDialog.vue
@@ -91,6 +91,21 @@
 
     <v-divider />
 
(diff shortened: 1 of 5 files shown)


matmen:
Well, looks like this isn't actually needed - Crowsnest also handles stream/snapshot mode from query params, so a simple configuration change fixes the issue.

PR #1012, opened 2023-01-15
fix: sort files names numerically when prefixed by numbers

### LINKED ISSUE #966: Settings to change Jobs sorting order

The Jobs panel orders files and folders by default by Modified date, and if we sort by name we get the same order as the **ls** command, in my use case (and I think most cases) it's very disorienting, for example I have:
```
~/gcode_files $ ls
10-Calibration  11-Upgrades  1-Cosplay  3-Models  4-Toys  5-Keychains  6-Home  7-Rings  8-Vases  9-Fusion360  Ender3
```

I think we should use the **-v** flag
```
~/gcode_files $ ls -v
1-Cosplay  3-Models  4-Toys  5-Keychains  6-Home  7-Rings  8-Vases  9-Fusion360  10-Calibration  11-Upgrades  Ender3
pi@chonkyraspy:~/gcode_files $ 
```

_No response_

### Additional information

I prints lots of files that have multiple parts and keep my folders organised and synced so an option to change the sort behaviour would be great, I would definitely use name with the **ls -v** order. Thanks!

matmen:
Closes #966

Before
image

After
image

### DIFF
--- a/src/plugins/filters.ts
+++ b/src/plugins/filters.ts
@@ -295,8 +295,13 @@ export const Filters = {
 
         [sortA, sortB] = [sortA, sortB].map(s => (s || '').toString().toLocaleLowerCase())
 
-        if (a.type === 'directory' || b.type === 'directory') {
+        if (a.type === 'directory' || b.type === 'directory' || a.type === 'file' || b.type === 'file') {
           if (a.type === b.type) {
+            if (isNaN(sortA) && isNaN(sortB) && /^\d+/.test(sortA) && /^\d+/.test(sortB)) {
+              // sort file names numerically when prefixed by numbers
+              return (parseInt(sortA.match(/^\d+/)) - parseInt(sortB.match(/^\d+/)))
+            }
+
             return stringCollator.compare(sortA, sortB)
           } else {
             return 0


pedrolamas:
In theory this looks ok, but I noticed that as this change is inside the `for` cycle, so it would not only apply to the filename but also to other fields, including date fields - and I expect that to cause issues!

src/plugins/filters.ts
```diff
@@ -295,8 +295,13 @@ export const Filters = {
 
         [sortA, sortB] = [sortA, sortB].map(s => (s || '').toString().toLocaleLowerCase())
 
-        if (a.type === 'directory' || b.type === 'directory') {
+        if (a.type === 'directory' || b.type === 'directory' || a.type === 'file' || b.type === 'file') {
```
pedrolamas, line 298: Won't this always be true?

matmen:
> In theory this looks ok, but I noticed that as this change is inside the `for` cycle, so it would not only apply to the filename but also to other fields, including date fields - and I expect that to cause issues!

I intentionally didn't limit it to file names only, this way other columns (e.g. Height) can be sorted too. The for loop only iterates over the sorted keys, so that should be fine

pedrolamas:
Ah, but height (and the other number or date columns) are already ordered by numeric value (hence the Number conversion a few lines lower)

Granted, the code is not pretty and sure looks like it needs a good cleanup!

matmen:
> Ah, but height (and the other number or date columns) are already ordered by numeric value (hence the Number conversion a few lines lower)

I don't think the Number parsing is actually used here, isNaN will return false for e.g. `123 mm` - the stringCollator would be used in these cases.

pedrolamas:
That's the thing: the code there is not using "123 mm", it's using 123 (the formatted string is only done on the grid, the source data is `number` type)

PR #1016, opened 2023-01-23
feat(ui): show content of release_info

BastelKlug:
This PR allows the showing of additional information about the distribution.
It also hides elements that are not available

screenshot

screenshot
 

Signed-off-by: Kerim Bilgic <pfusch3d@gmail.com>

### DIFF
--- a/src/components/widgets/system/SystemOverviewCard.vue
+++ b/src/components/widgets/system/SystemOverviewCard.vue
@@ -35,7 +35,7 @@
               <th>{{ $t('app.system_info.label.operating_system') }}</th>
               <td>{{ distribution.name }}</td>
             </tr>
-            <tr v-if="distribution.release_info.name">
+            <tr v-if="distribution.release_info?.name">
               <th>{{ $t('app.system_info.label.distribution_name') }}</th>
               <td>
                 {{ distribution.release_info.name }} {{ distribution.release_info.version_id }}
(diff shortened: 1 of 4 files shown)


src/components/widgets/system/SystemOverviewCard.vue
```diff
@@ -7,42 +7,48 @@
...
             </tr>
-            <tr>
+            <tr v-if="cpuInfo.bits && cpuInfo.processor && cpuInfo.cpu_count">
               <th>{{ $t('app.system_info.label.processor_desc') }}</th>
               <td>{{ cpuInfo.bits }} {{ cpuInfo.processor }} with {{ cpuInfo.cpu_count }} cores</td>
             </tr>
-            <tr>
-              <th>{{ $t('app.system_info.label.distribution_name') }}</th>
+            <tr v-if="distribution.name">
+              <th>{{ $t('app.system_info.label.operating_system') }}</th>
```
pedrolamas, line 35: Everything on the PR is perfect, only this label change here is giving me some doubts... I'm not sure what is the correct denomination here, if "Operating System" or "Distribution", but that comes from my lack of knowledge of Linux!

BastelKlug: I think operating system is correct because freedesktop.org  also calls it like this.

https://www.freedesktop.org/software/systemd/man/os-release.html

image

pedrolamas: Good enough for me!! 😁

pedrolamas (approved):

PR #1025, opened 2023-02-03
feat: Add support for gcode viewing on plotter-style devices

kieraneglin:
Previously, the gcode viewer did not work if no extrusion moves were defined. This broke the viewer in cases where Klipper was being used on 2d plotters without a Z or extruder.

### DIFF
--- a/src/store/gcodePreview/getters.ts
+++ b/src/store/gcodePreview/getters.ts
@@ -48,7 +48,7 @@ export const getters: GetterTree<GcodePreviewState, RootState> = {
       ? (a: number, b: number) => Number.isNaN(a) || a < b
       : (a: number, b: number) => a !== b
 
-    moves.forEach((move, index) => {
+    moves.forEach((move: Move, index: number) => {
       if (move.z !== undefined && z !== move.z) {
         z = move.z
         zStart = index
--- a/src/store/gcodePreview/getters.ts
+++ b/src/store/gcodePreview/getters.ts
@@ -48,21 +48,31 @@ export const getters: GetterTree<GcodePreviewState, RootState> = {
       ? (a: number, b: number) => Number.isNaN(a) || a < b
       : (a: number, b: number) => a !== b
 
-    for (let index = 0; index < moves.length; index++) {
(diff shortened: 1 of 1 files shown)


src/store/gcodePreview/getters.ts
```diff
@@ -48,21 +48,31 @@ export const getters: GetterTree<GcodePreviewState, RootState> = {
...
         zLast = z
 
         output.push({
           z,
           move: zStart,
-          filePosition: moves[index].filePosition
+          filePosition: move.filePosition
         })
       }
+    })
```
kieraneglin, line 66: This is largely just noise. I refactored the `for` loop to a `forEach` but functionality remains unchanged

pedrolamas: Can't remember what it was specifically, but I think I've had issues doing that in the past... 🤔

pedrolamas:
Thanks for sending this PR, @kieraneglin.

I saw that you provided a quick example gcode for this on Discord, but can I ask that you upload a sample file so we can run some tests?

kieraneglin:
@pedrolamas sure thing! I've attached a file that creates a small spiral using only XY moves as well as a reference for its expected output (courtesy of https://sandify.org/). Just rename the .txt to .gcode 🤙 

Screen Shot 2023-02-03 at 7 54 07 AM
[small_spiral_testing.txt](https://github.com/fluidd-core/fluidd/files/10580464/small_spiral_testing.txt)

pedrolamas:
Thanks for providing that @kieraneglin, it was quite useful to test - and even allowed me to find a non-related bug on the app-slider (for which I now have a fix!)

pedrolamas (approved):

pedrolamas:
Thank you @kieraneglin , can you please add a footer to the main PR comment with a "Signed-off-by" line, following the [CONTRIBUTING](https://github.com/fluidd-core/fluidd/blob/develop/CONTRIBUTING.md) guidelines (_you are missing the email address on yours!_)

Thank you!

kieraneglin:
@pedrolamas updated!

matmen (approved):

PR #1026, opened 2023-02-03
feat(i18n-zh): updates Chinese translations

Ercsion:
updates Chinese translations
Signed-off-by: Ercsion <bhs5612@gmail.com>

### DIFF
--- a/src/locales/zh-HK.yaml
+++ b/src/locales/zh-HK.yaml
@@ -1,6 +1,7 @@
 app:
   bedmesh:
     label:
+      active: active
       box_scale: 顯示比例
       flat_surface: 顯示參考平面
       mesh_matrix: 網格矩陣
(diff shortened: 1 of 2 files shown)


src/locales/zh-HK.yaml
```diff
@@ -89,74 +93,99 @@ app:
...
+      create_zip_archive: 壓縮
+      delete: 刪除
       download: 下載
       edit: 編輯
       exit_layout: 退出修改模式
       extrude: 擠出
+      filter: 過濾
+      forgot_password: 忘記密碼？
       heaters_off: 停止加熱
+      job_queue: ''
```
pedrolamas, line 144: There are 11 spots on this file with empty strings.

These need to either be translated or the key removed from the file (thus using the English fallback) so that we don't show empty labels!

image

pedrolamas (changes requested):
Thank you @Ercsion, can you please add a footer to the main PR comment with a "Signed-off-by" line, following the [CONTRIBUTING](https://github.com/fluidd-core/fluidd/blob/develop/CONTRIBUTING.md) guidelines.

Thank you!

pedrolamas (approved):

matmen (approved):

PR #1038, opened 2023-02-13
feat: updated tool extrusion settings to respect printer config

### LINKED ISSUE #1032: Unable to set tool extrude length > 50mm
### Fluidd Version

1.23.1

### Browser

Firefox

### Device

Desktop PC

### Operating System

Windows

### What happened

In the system settings it's possible to change the default extrude length to a 100mm.

Capture

But this screws up the tool interface on the dashboard and it seems like it won't allow higher than 50 there?

Capture2

Being able to use > 50mm extrude length

### How to reproduce

Try to set to > 50 extrude length in tool interface on dashboard.

### Additional information

_No response_

kieraneglin:
Fixes #1032

Updates the `ToolheadSettings` component to respect an extruder's `max_extrude_only_velocity` and `max_extrude_only_distance` when trying to update the default extrusion distance/speed values. Also adds a last-resort speed/distance check on extrusion/retraction commands

### DIFF
--- a/src/components/widgets/toolhead/ExtruderMoves.vue
+++ b/src/components/widgets/toolhead/ExtruderMoves.vue
@@ -81,14 +81,6 @@ import ToolheadMixin from '@/mixins/toolhead'
 export default class ExtruderMoves extends Mixins(StateMixin, ToolheadMixin) {
   valid = true
 
-  get maxExtrudeSpeed () {
-    return this.activeExtruder?.max_extrude_only_velocity || 500
-  }
-
-  get maxExtrudeLength () {
-    return this.activeExtruder?.max_extrude_only_distance || 50
-  }
-
   get extrudeSpeed () {
     const extrudeSpeed = this.$store.state.config.uiSettings.toolhead.extrudeSpeed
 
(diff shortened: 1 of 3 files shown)


src/components/widgets/toolhead/ExtruderMoves.vue
```diff
@@ -123,16 +115,22 @@ export default class ExtruderMoves extends Mixins(StateMixin, ToolheadMixin) {
 
   sendRetractGcode (amount: number, rate: number, wait?: string) {
     if (this.valid) {
+      const safeAmount = Math.min(amount, this.maxExtrudeLength)
+      const safeRate = Math.min(rate, this.maxExtrudeSpeed)
```
kieraneglin, line 119: This logic isn't strictly necessary as `this.valid` does a pretty good job of ensuring correctness before issuing an extrude/retract command. However, I so like this as a guardrail since it allows us to be confident that an extruder command won't exceed machine limits even if there was a logic bug in this form.

Now that I'm typing it out, I'm wondering if Klipper would even respect a command that would exceed its specified limits. I should test that.

At any rate, if Klipper handles this or if it's generally unwanted, I'll remove this portion. Just let me know!

kieraneglin:
I didn't know whether to title this as a feature or a bug fix or something else, so I defaulted to `feat`

matmen (dismissed):
lgtm!

src/components/widgets/toolhead/ExtruderMoves.vue
pedrolamas: I can confirm that Klipper will indeed complain if the value is above the `max_extrude_only_distance` 

image

pedrolamas: That should not be necessary (to be honest, even the `valid` check is probably overkill here), so I've gone ahead and reverted that small bit before doing the actual merge

pedrolamas (approved):

PR #1029, opened 2023-02-04
feat: moved gcode progress bar out of dialog

kieraneglin:
Loading a file into the gcode viewer opens a dialog that removes control until the file is parsed. This has some potentially nasty effects like delaying access to the emergency stop button, but it's also a minor and recurring annoyance. Based on that, I've moved the parsing progress information into the gcode card itself, lessening the time that control is removed from the user.

This PR also updates the behaviour of the file download dialog. Previously, it would not show if the Jobs panel wasn't present - this has been changed so now the file download dialog is independent of the filesystem components.

However, I am not a designer and I could really use some help in improving the design. Any contributions to this end would be much appreciated!

Signed-off-by: Kieran Eglin <kieran.eglin@gmail.com>

https://user-images.githubusercontent.com/569917/217333120-a21c691d-bbc2-4649-bcbb-5c57e868b5e7.mov

https://user-images.githubusercontent.com/569917/217333128-24c2e048-4094-49f5-99ff-91a94b594283.mov

https://user-images.githubusercontent.com/569917/217333132-55c23443-7920-4e07-9fb3-2d465f5c9381.mov

https://user-images.githubusercontent.com/569917/217333137-08358bfc-ae2c-4c44-8743-787a9ac85c5b.mov

https://user-images.githubusercontent.com/569917/217333141-b1603731-74ab-4658-881c-464664f42885.mov

PR #1040, opened 2023-02-13
feat: display gcode progress in overlay rather than dialog

kieraneglin:
re-attempt at #1029

Loading a file into the gcode viewer opens a dialog that removes control until the file is parsed. This has some potentially nasty effects like delaying access to the emergency stop button, but it's also a minor and recurring annoyance. Based on that, I've moved the parsing progress information into the gcode card itself, lessening the time that control is removed from the user.

Eventually I'd like to provide a mechanism to optionally display file download progress as an overlay (or some other mechanism that retains user control), but that's not captured in this PR.

Screen Shot 2023-02-13 at 3 23 27 PM
Screen Shot 2023-02-13 at 3 23 52 PM


kieraneglin:
@pedrolamas do you think there's interest in this PR?

pedrolamas:
I do see value in this PR, only I'm not entirely convinced of the approach.... I need to think a bit longer on it!

Having said that, we may be looking at this from the wrong side: the real problem is that the Emergency Stop button should ALWAYS be accessible, which is not the case when these dialogs are open!

kieraneglin:
@pedrolamas thank you for the feedback! The emergency stop issue is definitely pressing, but for me the main aspect of it is the annoyance of opening my web interface for a quick change or status update and having the download + parsing dialog holds that up. I'll be the first to admit that it's a minor thing, but it's a fairly frequent annoyance. 

Approach aside, my ideal goal would be having the file download progress/controls and the parsing progress contained entirely within the g-code viewer card so that it's not interfering with the use of the rest of the dashboard. 

Again, just let me know if that's not something you're interested in adding!

pedrolamas:
Implemented in e28a958f77198b94947720bae3135a42b4e46180

PR #1041, opened 2023-02-16
feat: allow ordering of gcode macros

### LINKED ISSUE #1027: Allow custom ordering of Macro buttons

Improving navigation of large macro groups  

It would be useful to be able to change and customise the order of macro buttons within a group rather than always presenting them in alpha sequence. This would make it clearer and easier to navigate big groups and keep related items next to each other.  It doesn't need to be an elaborate / drag and drop style control just an extra hidden sort seq field override on each macro to allow placement and sub grouping within the macro group.

Prefixing macro aliases with numerics or deliberately renaming them to affect sort order.

### Additional information

_No response_

kieraneglin:
Resolves #1027

https://user-images.githubusercontent.com/569917/219500268-7acc13dd-8fc6-4acc-a79b-b670dea40e85.mov


src/store/macros/getters.ts
```diff
@@ -40,37 +40,31 @@ export const getters: GetterTree<MacrosState, RootState> = {
...
         return macro
       })
-      .sort((a, b) => a.name.localeCompare(b.name))
 
     return macros
   },
 
   // Gets visible macros, transformed. Should include the macro's config.
   // Is only used on the dashboard. Grouped by category.
   getVisibleMacros: (state, getters) => {
```
kieraneglin, line 49: This needed to be updated to utilize `getMacrosByCategory` so that macros would be sorted per-category, rather than being sorted overall then split into categories which lead to unexpected behaviour. Since I needed to update it anyway, I thought I'd refactor it to DRY things up somewhat

pedrolamas: Good point, this was actually rather stupid considering we will then do other operations on this!

pedrolamas (approved):

matmen (approved):

Issue #908, opened 2022-10-12
Perform sanity checks on UserInput field in Fluidd to prevent failed prints due to Klipper crashing.


netweaver1970:
### Is your feature request related to a problem? Please describe

I was live adjusting the PressureAdvance setting in FLuidd, made a mistake (entered 25 instead of 0.025, when overwriting 0.035) and the clearly impossible value was accepted and -obviously- sent to Klipper. Then Klipper tries to do something with it and crashes, as it doesn't make sense at all. Moral of the story, a 6 hr print wasted due to this:
![image](https://user-images.githubusercontent.com/18028618/195374136-0f64228d-12dd-4dae-9b77-b474acea32d0.png)

### Describe the solution you'd like

a sensible value range check on all (at least the ones that can immediately kill a print) at input time, so the user (and printer) is protected against bad input.

### Describe alternatives you've considered

don't touch anything in FLuidd after starting a  print job. But that's not what's Fluidd is made for.

pedrolamas:
Thanks for reporting this @netweaver1970.

I get your point and I do think there should be some sanity check on these values, but I wonder if these shouldn't actually be made centrally in Klipper instead of having all the interfaces do it...

netweaver1970:
Sure, one can do it on multiple levels. I agree that the lowest level (aka Klipper) should have its own input validation but knowing how busy Kevin is and me not being able to make a proper PR for the validation(s), with the added risk that it won't get reviewed/merged anyhow, I think the likeliest way forward is to do it one level up, at the GUI side. 

Of course, it wouldn't be optimal, as the Fluidd enhancement wouldn't help Mainsail users but one could argue that's their problem then :)

slavedas:
@pedrolamas thank you!  It is appreciated.
@netweaver1970 you definitely aren't the only one and I agree that bounds checking would help as well - even if it's user defined values that we ourselves are responsible for setting.  I'd love it if during a print if the temp is set to below min-temp either in the GCODE or by the user, it would pause the print and throw up a warning instead of allowing it to run to unrecoverable failure.

pedrolamas:
@slavedas took a while, but we've just released 1.23.2 with an improved handling of input on those controls

> I'd like to comment that the CAUSE of this error is that when you have 0.035 and delete the 35 to make 0.0 the UI autocorrects this value to 0

This scenario should not happen anymore, as no format of the input will happen until the user presses the <kbd>Enter</kbd> key!

[...]

PR #1042, opened 2023-02-17
refactor: input control improvements

pedrolamas:
This is a big PR with quite some changes, specifically:

- Adds new `AppTextField` - a control that auto-selects text on focus, and provides a "submit" event on enter
- Renames controls with integrated label to `AppNamed*`
  - Renamed `AppSelect` to `AppNamedSelect`
  - Renamed `AppSlider` to `AppNamedSlider`
  - Renamed `AppSwitch` to `AppNamedSwitch`
- Improves `AppNamesSelect` considerably, to better handle the situation reported [here](https://github.com/fluidd-core/fluidd/issues/908#issuecomment-1305756153), also adds new "submit" event that is raised on <kbd>ENTER</kbd> or manual slider change
- Simplifies Boolean attributes (so `:disabled="true"` is now just `disabled` and for false we can just omit the attribute)
- Fixes a bug for temperature fans with min temp 0 would get undefined instead

A major change is that all dashboard AppNamedSlider controls now require pressing <kbd>ENTER</kbd> to accept the value change (only affects direct text input, slider is unchanged)


matmen (approved):
nice, lgtm!

Issue #1044, opened 2023-02-21
Slider input lockout not present on all tablets


kieraneglin:
### Fluidd Version

Latest develop

### Browser

Safari, Other (Please tell us below)

### Device

Mobile device (Phone, Tablet or similar)

### Operating System

Android, iOS

### What happened

Slider input locks aren't present on tablets that are larger than the Vuetify mobile breakpoint 

### What did you expect to happen

Slider input locks should be present on tablets. Possible solutions:

- add checks for touch screens - possibly in tandem with a larger breakpoint cutoff to not show locks on touchscreen laptops. ([relevant SO question](https://stackoverflow.com/questions/56324813/how-to-detect-touch-device-in-2019))
- change the cutoff to include larger devices. This has bigger ramifications as there are a lot of independent definitions of `isMobile` in the codebase that would all need their tablet behavior considered. 
  - tangential: it may be worth refactoring `isMobile` into a helper method that's then imported and called by the views. DRY it up and all that

### How to reproduce

Open Fluidd UI on a normal tablet (an iPad, for example)

kieraneglin:
@pedrolamas if there is consensus on how to handle this, I can open a pull request with a fix!

pedrolamas:
I don't use tablets, so I never came across this... having said that, we do have a better `is-mobile` method that checks for browser agent: https://github.com/fluidd-core/fluidd/blob/develop/src/util/is-mobile.ts

At the moment, we only use it to ensure we don't show Monaco editor on mobile devices as it is not supported and most likely wouldn't work.

My guess is that this method would be a better fit for purpose than using the existing Vuetify check!

PR #1045, opened 2023-02-21
chore: refactor isMobile to use mixin

kieraneglin:
Works toward (but does not resolve) #1044

This replaces the often-redefined `isMobile` component getter with a mixin. Since there was previously two mechanisms for detecting mobile (more commonly through viewport size but sometimes through user agent), I've also renamed `isMobile` to better reflect what we're actually checking for.

After this I'll make the change to fully resolve the issue above, but I didn't want to clutter up this PR.

### DIFF
--- a/src/components/widgets/thermals/ThermalChart.vue
+++ b/src/components/widgets/thermals/ThermalChart.vue
@@ -16,12 +16,13 @@
 </template>
 
 <script lang='ts'>
-import { Vue, Component, Watch, Prop, Ref } from 'vue-property-decorator'
+import { Component, Watch, Prop, Ref, Mixins } from 'vue-property-decorator'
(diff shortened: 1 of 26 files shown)


src/components/widgets/thermals/ThermalChart.vue
```diff
@@ -292,7 +292,7 @@ export default class ThermalChart extends Vue {
 
   createSeries (label: string, key: string) {
     // Grab the color
-    const color = Vue.$colorset.next(getKlipperType(key), key)
```
kieraneglin, line 295: This is an unrelated change, but the TS compiler wouldn't let me continue without fixing it

pedrolamas (approved):

matmen (approved):
LGTM, thanks!

PR #1048, opened 2023-02-27
feat(i18n-UK): update Ukrainian translations

Entraft:
feat: Fixed some typos and added missing translations.

Signed-off-by: Entraft pasinc23@gmail.com

### DIFF
--- a/src/locales/uk.yaml
+++ b/src/locales/uk.yaml
@@ -1,13 +1,14 @@
 app:
   bedmesh:
     label:
-      box_scale: Масштаб коробки
+      active: активно
+      box_scale: Масштаб координат
       flat_surface: Показати базову площину
       mesh_matrix: Сітчаста матриця
       probed_matrix: Зондована матриця
-      profile_name: Ім'я профілю
-      remove_profile: 'Видалити профіль %{name} '
-      scale: Масштаб кольорів
+      profile_name: Імʼя профілю
+      remove_profile: Видалити профіль %{name}
+      scale: Масштаб площини
       wireframe: Каркасна сітка
     msg:
       hint: >-
@@ -30,6 +31,7 @@ app:
       'on': Діаграма вимк.
       power: Потужність
       target: Ціль
+      rate_of_change: Зміни
     tooltip:
       help: >-
         Утримуйте Shift для збільшення.<br />Клікніть на елемент, щоб змінити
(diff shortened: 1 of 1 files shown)


src/locales/uk.yaml
```diff
@@ -74,92 +78,141 @@ app:
...
+      root_disabled: Root недоступний. Перевірте файли логів.
+      url:
+        klipper_config: https://www.klipper3d.org/Config_Reference.html#%{hash}
+        moonraker_config: https://moonraker.readthedocs.io/en/latest/configuration/#%{hash}
+        moonraker_telegram_bot_config: >-
+          https://github.com/nlef/moonraker-telegram-bot/wiki/Sample-config#%{hash}
+    url:
+      klipper_config: https://www.klipper3d.org/Config_Reference.html
+      moonraker_telegram_bot_config: https://github.com/nlef/moonraker-telegram-bot/wiki/Sample-config
+      moonraker_config: https://moonraker.readthedocs.io/en/latest/configuration/
```
pedrolamas, line 120: This bit doesn't look correct, there should not be any `url` under `tooltip` as it currently shows

pedrolamas (changes requested):

src/locales/uk.yaml
Entraft: I added them because in other languages this is done in such way (through a plugin for VS code i18n Ally)

Entraft: Thank you! Done.

src/locales/uk.yaml
```diff
@@ -74,92 +78,136 @@ app:
...
       rename_dir: Перейменувати папку
       rename_file: Перейменувати файл
       upload_file: Завантаження файлу | Завантаження файлів
     tooltip:
       low_on_space: Мало місця на диску
+      root_disabled: Root недоступний. Перевірте файли логів.
+    url:
+      klipper_config: https://www.klipper3d.org/Config_Reference.html
+      moonraker_telegram_bot_config: https://github.com/nlef/moonraker-telegram-bot/wiki/Sample-config
+      moonraker_config: https://moonraker.readthedocs.io/en/latest/configuration/
```
pedrolamas, line 115: These are the used by Code Lens for deep linking, the hash params and the rest of the format needs to be kept as is (only thing to change would be the destination language, but I believe none of these components actually have ukranian support)

```suggestion
      klipper_config: 'https://www.klipper3d.org/Config_Reference.html#%{hash}'
      moonraker_config: 'https://moonraker.readthedocs.io/en/latest/configuration/#%{hash}'
      moonraker_telegram_bot_config: 'https://github.com/nlef/moonraker-telegram-bot/wiki/Sample-config#%{hash}'
```

pedrolamas (approved):
This is perfect, thank you @Entraft

pedrolamas (approved):

PR #1079, opened 2023-04-09
feat: adds load/unload filament support

pedrolamas:
If there is an `UNLOAD_FILAMENT` macro, that will be shown in the Tools menu; if not, if there is an `M702` macro, that will be shown in the Tools menu (as "M702 (Unload filament)")

Similarly, if there is a `LOAD_FILAMENT` macro, that will be shown in the Tools menu; if not, if there is an `M701` macro, that will be shown in the Tools menu (as "M701 (Load filament)")

(Priority will always be given to `UNLOAD_FILAMENT` and `LOAD_FILAMENT` macros)

image

### DIFF
--- a/src/components/widgets/toolhead/ToolheadCard.vue
+++ b/src/components/widgets/toolhead/ToolheadCard.vue
@@ -101,23 +101,29 @@
             </app-btn>
           </template>
           <v-list dense>
-            <v-list-item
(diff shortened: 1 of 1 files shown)


pedrolamas:
I'm not very fond of the icons, but if someone has a better suggestion, I am all ears!

PR #1097, opened 2023-05-07
i18n: improves delete confirmation message

### LINKED ISSUE #1080: "Are you sure? This will delete all files and folders." for job delete
### Fluidd Version

1.23.4

### Browser

Firefox

### Device

Laptop

### Operating System

Linux

### What happened

When I click on a file in the jobs list and select remove the dialog that comes up is "Are you sure? This will delete all files and folders."

I would suggest that this makes it sound like a bulk operation that's going to wipe a lot of data instead of just removing that one thing you asked it to remove.  I would suggest changing that message to confirming that the selected job will be removed.

### How to reproduce

Click an item in jobs and select remove.

### Additional information

_No response_

pedrolamas:
Improves "delete item/items" confirmation message

## Single item

image

## Multiple items

image

Fixes #1080

### DIFF
--- a/src/components/widgets/filesystem/FileSystem.vue
+++ b/src/components/widgets/filesystem/FileSystem.vue
@@ -764,7 +764,9 @@ export default class FileSystem extends Mixins(StateMixin, FilesMixin, ServicesM
       destinationPath = arr.join('/')
     }
 
-    const items = (Array.isArray(source)) ? source.filter(item => (item.name !== '..')) : [source]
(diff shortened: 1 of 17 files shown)


matmen (approved):
LGTM - opened https://github.com/pedrolamas/fluidd/pull/8 for the German translation

matmen (approved):

thijstriemstra:
ha, was about to file a bug for this. thanks for the fix!

PR #1099, opened 2023-05-10
feat: save and restore editor view state

### LINKED ISSUE #1092: Add an Option to open setting files folded

When I open a .cfg file in the Editor, i almost always instinctively hit "fold all" to navigate the file and get a overview. In case anybody doesn't know this very usefull feature, it gets you from this: 
image

to this: 
image

I would like a little Option in the File Editor section to open cfg files with all sections folded.

Alternatively an even better approach would be to store the "fold state" between multiple openings of the same file, but that is probably considerably more effort.

### Additional information

_No response_

pedrolamas:
This PR adds a new Monaco Editor option to "Save and restore view state" (defaults to enabled)

With this option enabled, any file that the user opens will have the current view state (scroll position, selected areas, folds/unfolded regions, ...) stored in the browser local storage when the user closes the editor, and restored when the user re-opens the file.

https://github.com/fluidd-core/fluidd/assets/85504/872b5f70-4632-47bc-ace3-6ac30b9f85a6

To ensure there are no collisions between multi-instances, roots, paths, and filenames, this is the process to generate the state key:

- Create the "get file" Moonraker API URL (so that will be `<moonraker_base_url>/server/files/<root>/<path>/<filename>`)
- Generate an MD5 of this url and encode in Base64 (makes a smaller string)
- Prepend "monaco."

Fixes #1092

### DIFF
--- a/src/components/settings/FileEditorSettings.vue
+++ b/src/components/settings/FileEditorSettings.vue
@@ -38,6 +38,19 @@
       <v-divider />
 
       <app-setting
+        :title="$t('app.setting.label.save_and_restore_view_state')"
+      >
+        <v-switch
+          v-model="restoreViewState"
+          hide-details
+          class="mb-5"
+          @click.native.stop
+        />
+      </app-setting>
+
+      <v-divider />
+
+      <app-setting
         :title="$t('app.setting.label.show_code_lens')"
       >
         <v-switch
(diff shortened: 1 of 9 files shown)


matmen (approved):
Nice, LGTM!
Opened https://github.com/pedrolamas/fluidd/pull/9 for the German translation

PR #1100, opened 2023-05-11
Macro buttons: send only modified parameters

Piezoid:
Reduces console clutter and fix issues when the default parameters are invalid.

### DIFF
--- a/src/components/widgets/macros/MacroBtn.vue
+++ b/src/components/widgets/macros/MacroBtn.vue
@@ -113,8 +113,10 @@ export default class MacroBtn extends Mixins(StateMixin) {
       if (['m117', 'm118'].includes(this.macro.name)) {
         s += ` ${this.params.message.value}`
       } else {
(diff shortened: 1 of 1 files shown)


pedrolamas:
Thanks for submitting this PR @Piezoid, however I am not entirely sure we can commit it without breaking existing macros!

The reason is that there is a major difference between sending an empty value on a parameter and completely omitting said parameter!

Take this example macro:

```ini
[gcode_macro TEST]
gcode:
    {% set A = params.A %}
    {% set B = params.B %}
    M118 (A == "") = { A == '' }
    M118 (B == "") = { B == '' }
    M118 A is null = { A == null }
    M118 B is null = { B == null }
```

This macro has two parameters `A` and `B`, and will output both values, and if any is empty string or null.

If you run `TEST A=` you will get this output:

image

As you can see from the above, setting an empty value will just be assumed as an empty string, but omitting the parameter will set it to `null`!

PR #1101, opened 2023-05-11
fix: more selective keyboard events for history navigation

Piezoid:
While copying text from a different virtual desktop to fluidd's console, I was annoyed that the window manager global shortcuts (Meta+Up/Down keys) triggered history navigation in the text input, erasing the current input. This is because keyup events are not captured by the windows manager, unlike keydown events.

I also added .exact filters, enabling navigation in multi-line/wrapped inputs using the Ctrl modifier, and selection using the Shift modifier.


pedrolamas (approved):
Very good catch!! Thanks for submitting this PR @Piezoid

pedrolamas (changes requested):
Can you please add a footer to the main PR comment with a "Signed-off-by" line, following the [CONTRIBUTING](https://github.com/fluidd-core/fluidd/blob/develop/CONTRIBUTING.md) guidelines.

Thank you!

pedrolamas (approved):

PR #1114, opened 2023-06-23
feat: basic Nevermore Controller support

SanaaHamel:
Add support for [Nevermore Controller](https://github.com/SanaaHamel/nevermore-controller) sensors. Sadly due to how Fluidd reads data, I can't fake it by pretending to be a bme280.

Signed-off-by: Sanaa Hamel <sanaahamel@outlook.com>


pedrolamas:
Thanks @SanaaHamel for this PR.

I will keep this in mind and see if there is something we can improve on this area to make it more "dynamic", but I must admit I am not comfortable adding support for ad-hoc sensors that are not supported by mainline Klipper.

SanaaHamel:
I'd prefer a more flexible solution as well, but don't really have the time to dedicate to it. Mainsail has the same issue, but I can manage it there using shenanigans with sensor's qualified object name. Fluidd is more pedantic (doesn't use the full object name when selecting sensor-like objects, and gives up if the sensor object prefix doesn't match the declared sensor kind).

Ideally I'd like to be able to declare compound sensor sources (temp, pressure, etc), as well as plot non-temperature data. I expect that would take some changes in Klipper, however.

pedrolamas (approved):

pedrolamas:
We have been revisiting some of these feature requests as we have been adding support to non-mainline Klipper modules for a while.

Given this, I have reviewed and approved the changes on this Pull Request.

matmen (approved):
LGTM, thanks!

pedrolamas:
@SanaaHamel any chance I can get a full JSON blob of the `nevermoresensor` status reported by Klipper?

I might be looking into adding some more functionally to the bit this PR already added! 😉

SanaaHamel:
> @SanaaHamel any chance I can get a full JSON blob of the `nevermoresensor` status reported by Klipper?
> 
> I might be looking into adding some more functionally to the bit this PR already added! 😉

```python
class NevermoreSensor:
  ...
  DISPLAY_FIELDS = {"temperature", "humidity", "pressure", "gas", "gas_raw"}
  ...
```
These are the fields provided. Same interp as the main nevermore obj, except `pressure` is rounded, and `gas_raw` is an int.

PR #1117, opened 2023-07-04
feat(i18n-PL): update Polish translations

Daviteusz:


src/locales/pl.yaml
```diff
@@ -124,7 +124,7 @@ app:
       heaters_busy: Drukarka jest obecnie zajęta. Wyłączenie grzałek może spowodować
         nieudany wydruk.
       flow: Przepływ
-      pressure_advance: Natężenie docisku
+      pressure_advance: Pressure Advance
```
matmen, line 127: Is this intentional? Seems like it was translated before

Daviteusz: This is a technical translation that should remain in the original so that the documentation can be found. The documentation is not currently available in Polish.

PR #1119, opened 2023-07-07
feat: spoolman support

matmen:
Adds support for [Spoolman](https://github.com/Donkie/Spoolman) (https://github.com/Arksine/moonraker/pull/651)

done;
- [x] get spoolman state from moonraker / spoolman instance
- [x] check if spoolman is supported and configured
- [x] spool selection dialog on print start
  - [x] metadata filters (search)
  - [x] default to currently selected spool
- [x] Warning if remaining filament on spool might not be enough
- [x] Warning if filament types don't match
- [x] "Change Spool" option during print/idle
- [x] support selecting spools by QR code
  - [ ] [depends on standardized format for spool QR codes, see Donkie/Spoolman/issues/10]
    - [x] support for `<origin>/spools/show/<id>` codes (see https://github.com/Donkie/Spoolman/issues/10#issuecomment-1635689582)
  - [ ] device camera (out of scope for now, requires secure context (HTTPS))
  - [x] printer camera
    - [x] camera selection on multi cam setup
    - [x] auto selection of spool
    - [x] MJPEG (Adaptive + Stream)
    - [x] HLS Stream
    - [x] WebRTC
    - [x] IP Camera
    - [ ] HTTP page (unsupported)
- [x] docs
- [x] test mobile support
- [x] settings
  - [x] setting to disable automatically showing the modal
  - [x] setting to automatically open QR code detection (default camera?)
- [x] move QR detection to separate thread
- [x] display currently active spool in dashboard

todo:
- [ ] let user select spool on print start through external API call
  - [ ] IMO not really needed with the Spoolman dashboard card, adds unnecessary complexity

### Screenshots
Spool Selection modal on print start
image

Warning when trying to start a print without a spool selection
image

Not enough filament warning
image

Mismatched filament material warning
image

Spoolman dashboard card
image
image

Filtering spools
image

#### QR code scanning
image

Camera Selection on multi cam setup
image

Warning on unsupported code
image

Error on inexistant spool
image

Auto-closes when scanning a valid code (compatible pattern + ID exists in database), selects spool in modal
https://github.com/fluidd-core/fluidd/assets/25269274/30ba1f17-95dd-483c-a1c5-1ac067d9d7ba


Donkie:
Neat! Let me know if you need any help integrating with Spoolman :)

matmen:
@Donkie I think having something like a spool list update event would be nice to have so changes can be reflected in the frontend in real time? Either sending the full spool list or just a subset of modified/created/deleted spools.
Currently this would either require explicit fetching of the list (in this case when creating the spool selection dialog) or manually reloading the page, which isn't the best UX wise imo

Donkie:
> @Donkie I think having something like a spool list update event would be nice to have so changes can be reflected in the frontend in real time? Either sending the full spool list or just a subset of modified/created/deleted spools.
> 
> Currently this would either require explicit fetching of the list (in this case when creating the spool selection dialog) or manually reloading the page, which isn't the best UX wise imo

Sure I could add that to my To-Do list

matmen:
I've added support for QR codes containing the string `/spool/show/<id>` for now.
In theory, this string alone could be used to identify spools, in addition to that it can be prefixed with the spoolman URL (e.g. `http://docker:7912/spool/show/1`) so they can be scanned with a mobile device and lead to the management page directly.
The only caveat I see here is that, if the spoolman base URL changes, the QR codes will need to be re-printed (though QR code scanning in Fluidd would still work).

Happy to add other patterns if the need for those exist, or if there is a decision for a standardized QR code content for Spoolman.

matmen:
@Donkie Feel free to review too or let me know if I'm missing anything 😉

Donkie:
I've been thinking about what to put in the QR code, and I don't see the point of making it a URL. I think something simple like `S:123` would do just fine. Keeping it short means we can use smaller QR codes but still have a good chance of scanning it successfully.

Sure, if it were a URL, people could use regular phone QR code scanners and see spool details, but I feel like that's not a huge deal. If you've got the spool right in front of you, you can already see some of those details on the label. We'll also have visible text on the label with spool ID and material (see the suggestion here: https://github.com/Donkie/Spoolman/issues/54#issuecomment-1646890837). Plus, being dependent on a fixed spoolman URL doesn't sound great.

By the way, I'm also planning on adding a QR scanner camera app to the spoolman interface. It will then be able to understand the `S:123` code and take you right to the spool show page.

What do you think about these arguments? @matmen

pedrolamas:
Just started to test & review this (and just installed spoolman on my docker stack), and am seeing we need to add some more `null` checks!

image

pedrolamas:
Most of these props seem to be optional, so we might want to mark as such so that TypeScript flags the places we need to perform null/undefined checks!

image

Donkie:
> Most of these props seem to be optional, so we might want to mark as such so that TypeScript flags the places we need to perform null/undefined checks!
> 
> image

Yup, you can find the ones I use here, which have the correct data types: https://github.com/Donkie/Spoolman/blob/master/client/src/pages/spools/model.tsx (and equivalent path for filaments and vendors)

matmen:
@Donkie Good points, I was just assuming that it would be easier to open a spool and view/edit it in the web UI with a full-URL QR code. Scanning the codes in spoolman directly would be a good alternative I think.
I assume you haven't decided on a specific format for the code contents yet and `S:123` is just an example?

Donkie:
> @Donkie Good points, I was just assuming that it would be easier to open a spool and view/edit it in the web UI with a full-URL QR code. Scanning the codes in spoolman directly would be a good alternative I think. I assume you haven't decided on a specific format for the code contents yet and `S:123` is just an example?

I have not settled on anything yet, but IMO `S:123` is as good format as any, do you have any other suggestions?

pedrolamas:
> I have not settled on anything yet, but IMO `S:123` is as good format as any, do you have any other suggestions?

When you saw your messages yesterday, I was going to suggest creating a custom app moniker/proto/intent, something like `spoolman://<data>`

Here's an example for WhatsApp, try to scan this and open it:

image

On Android (unsure if it works on iPhone) it works as WhatsApp registered the "whatsapp" intent, so it will open a new message in WhatsApp for the specified number!

matmen:
> > Most of these props seem to be optional, so we might want to mark as such so that TypeScript flags the places we need to perform null/undefined checks!
> > image
> 
> Yup, you can find the ones I use here, which have the correct data types: https://github.com/Donkie/Spoolman/blob/master/client/src/pages/spools/model.tsx (and equivalent path for filaments and vendors)

My bad, I should've checked the docs haha, I just assumed every filament had a material, vendor, etc. Fixed now ;)

Donkie:
@pedrolamas that's a really good idea. Support for web seems to be on the horizon, with chrome already supporting it. https://developer.mozilla.org/en-US/docs/Web/Manifest/protocol_handlers

pedrolamas:
@Donkie well, I've got some better news then: PWA apps already support [protocol handling](https://developer.mozilla.org/en-US/docs/Web/Manifest/protocol_handlers) (and it works on most modern browsers)!!

Fluidd is currently a full PWA app, which means we could register it as handler for something like "fluidd:" and it would launch on such links (assuming the user installed Fluidd as PWA, so one could take advantage of that)

If you manage to get Spoolman registering on the computer as a PWA app, you could then register and handle such protocol!

Nevertheless, my vote goes to use a "spoolman://" protocol as a future proof solution!

pedrolamas:
Quick list of issues I have found:

## Startup errors

image

I believe these are because of the `mount` changes in `CameraItem.vue`, as the subcomponent is not ready yet (thus `streamingElement` is undefined at that point)

## Full dialog scroll

Animation2

Though not a big problem, ideally we want to use set `scrollable` prop in the `v-dialog` so that the header and footer are always visible (check `AppDialog.vue` for example)

src/components/widgets/toolhead/ToolheadCard.vue
```diff
@@ -154,8 +154,10 @@ type Tool = {
   name: string,
   label?: string,
   disabled?: boolean,
+  enabledOnPrint?: boolean,
   wait?: string,
   icon?: string,
+  callback?: () => void,
 }
```
pedrolamas, line 161: Are these changes required? I can't seem to find this in use anywhere...

matmen: Oops, that's a leftover from when the Change Spool button was in the Tools submenu, I'll remove those later

matmen:
@pedrolamas Thanks for the feedback - should be fixed now 😉

pedrolamas (dismissed):

pedrolamas (approved):

NonaSuomy:
"device camera (out of scope for now, requires secure context (HTTPS))"

Make the browser device camera work with non https local sources.

`chrome://flags/#unsafely-treat-insecure-origin-as-secure`

Enable the `Insecure origins treated as secure`

Add the address for your Spoolman/Fluidd server IP:Port that you want to ignore the secure origin policy for. Remember to include the port number too.

image

Save and restart Chrome.

Click enable on camera request when pops up.

NonaSuomy:
Instead of failing from a nonexisting QR code lookup can you make it check if it has the right properties say a new spool that has a spoolman sticker already on it from the factory or friend and then ask if you want to add it to the spoolman db?

matmen:
> Instead of failing from a nonexisting QR code lookup can you make it check if it has the right properties? Say a new spool that has a Spoolman sticker already on it with all the fields embedded from the factory or friend and then ask if you want to add it to the Spoolman DB.

I don't think there's a point to have this in fluidd - we're just acting as the frontend for spool selection, not for the data management itself. All other data management actions (creating/editing vendors, filaments, spools) already happens in the Spoolman UI..

matmen:
> "device camera (out of scope for now, requires secure context (HTTPS))"
> 
> Make the browser device camera work with non https local sources.
> 
> `chrome://flags/#unsafely-treat-insecure-origin-as-secure`
> 
> Enable the `Insecure origins treated as secure`
> 
> Add the address for your Spoolman/Fluidd server IP:Port that you want to ignore the secure origin policy for. Remember to include the port number too.
> 
> image
> 
> Save and restart Chrome.
> 
> Click enable on camera request when pops up.

Thanks, I'll take a look at this in the following couple days and add it to https://github.com/fluidd-core/fluidd/pull/1149.

NonaSuomy:
> > Instead of failing from a nonexisting QR code lookup can you make it check if it has the right properties? Say a new spool that has a Spoolman sticker already on it with all the fields embedded from the factory or friend and then ask if you want to add it to the Spoolman DB.
> 
> I don't think there's a point to have this in fluidd - we're just acting as the frontend for spool selection, not for the data management itself. All other data management actions (creating/editing vendors, filaments, spools) already happens in the Spoolman UI..

That's fine I've requested this as an addition in spoolman I'm just saying support the ability to tell spoolman that the spool is not in the database that was scanned how ever you are checking that so that spoolman can then add it from your scan.

Right now the QR-Code only has `web+spoolman:s-1` encoded hopefully eventually there will be all the basic elements in the QR-Code to add a full DB entry.  So in this case you would just have to fully post the QR-Code data to spoolman to add an entry.

goodwin:
Are there any plans to make it working with Tool Changer support? For ex., using it with ERCF or other MMU

pedrolamas:
@goodwin currently there is no specific support for any type of MMU, but you might be able to achieve that by using macros: https://moonraker.readthedocs.io/en/latest/configuration/#setting-the-active-spool-from-klipper

matmen:
@goodwin FYI toolchanger support has been implemented via https://github.com/fluidd-core/fluidd/pull/1324 and released in v1.28: https://docs.fluidd.xyz/features/spoolman#toolchanger-support

PR #1124, opened 2023-07-21
Tools macros - support tool objects.

viesturz:
Detect tools based on tool objects. To support extension based tool configs.

WIP - looking for initial feedback, follwup from https://github.com/fluidd-core/fluidd/pull/1123
The code is very crude, but the main idea is that extension based configs are like this:
```
[tool T0]
 some config
[tool T1]
 more config
```

And the ative tool is detected via tool status, not macro variables.

EG: https://github.com/TypQxQ/Klipper_ToolChanger and forks or https://github.com/viesturz/klipper/tree/toolchanger.


pedrolamas:
Implemented via #1197

PR #1149, opened 2023-08-13
feat(spoolman): QR code scanning support

matmen:
As described in https://github.com/fluidd-core/fluidd/pull/1119

Adds support for spoolman QR codes (https://github.com/Donkie/Spoolman/issues/10).
Adds an option to automatically commit a spool selection when a valid QR code is scanned (no need to manually confirm the selection)
Adds support for using the device's camera as a scan source (https://github.com/fluidd-core/fluidd/pull/1119#issuecomment-1676159359)
Closes https://github.com/fluidd-core/fluidd/issues/1151

### DIFF
--- a/src/components/widgets/spoolman/QRReader.vue
+++ b/src/components/widgets/spoolman/QRReader.vue
@@ -35,7 +35,7 @@ import BrowserMixin from '@/mixins/browser'
 })
 export default class QRReader extends Mixins(StateMixin, BrowserMixin) {
   dataPatterns = [
-    /web\+spoomnan:s-(\d+)/,
+    /web\+spoolman:s-(\d+)/,
     /\/spool\/show\/(\d+)\/?/
   ]
 
--- a/src/components/widgets/spoolman/QRReader.vue
+++ b/src/components/widgets/spoolman/QRReader.vue
@@ -16,6 +16,7 @@
       <CameraItem
         :camera="camera"
         :embedded="true"
+        crossorigin="anonymous"
         @frame="handlePrinterCameraFrame"
       />
     </v-card-text>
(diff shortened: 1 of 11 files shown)


src/components/widgets/spoolman/QRReader.vue
```diff
@@ -34,7 +34,11 @@ import BrowserMixin from '@/mixins/browser'
   components: { CameraItem }
 })
 export default class QRReader extends Mixins(StateMixin, BrowserMixin) {
-  dataPatterns = [/\/spool\/show\/(\d+)\/?/]
+  dataPatterns = [
+    /web\+spoomnan:s-(\d+)/,
```
pedrolamas, line 38:
```suggestion
    /web\+spoolman:s-(\d+)/,
```

pedrolamas (commented):
One typo, once that is fixed this is good to go!

pedrolamas (approved):

matmen:
@NonaSuomy Feel free to test these changes, I just added support for using the device's camera as a scan source

pedrolamas:
@matmen To fix #1150, I made `crossorigin` an optional property, so you will need to add `crossorigin="anonymous"` to [this element](https://github.com/fluidd-core/fluidd/pull/1149/files#diff-cd6975ab9d4451c59e4630210657f1df8a4f930f6a901edd948f187b20d06490R16)

Having said that we now know that this will still fail if the streaming server is not setting the `Access-Control-Allow-Origin "*"` response header... in that case the DOMException that you are already catching will have `ex.name === "SecurityError"`

pedrolamas:
And while we're at it, we might want to add #1151 here?

pedrolamas (approved):

PR #1177, opened 2023-09-18
feat: Adds toolhead bars style controls

pedrolamas:
The default toolhead control in Fluidd is the "cross" style control, but some users do prefer the more direct "bars" style control (as it requires a single click to move an axis).

This PR adds such controls.

## Default state

image

## "FORCE_MOVE" enabled state

image

## Settings when using "Bars" style

image

## Settings when using "Cross" style

image

Users can now choose between "cross" and "bars" style controls, specifying the step increments in case of the latter.


pedrolamas:
While working on this, I am considering removing the "FORCE_MOVE" from the "Tool" card and moving this functionality to a new "Steppers" card under the "Tune" tab (but would allow said card to be added to the Dashboard in case users still wanted that)

@matmen @kerbilg any thoughts on this?

PR #1182, opened 2023-09-20
feat: Added "calibrate" button to BedMeshCard

BillyNate:
Added a "calibrate" button to the Bed Mesh title menu.
Just so we no longer need to switch to the Tune tab, but start a bed mesh calibration from the Home tab.


pedrolamas (approved):
Thanks @BillyNate, looks like a nice addition! 🙂

matmen (approved):
LGTM, thanks!

PR #1209, opened 2023-10-18
feat: add show hidden option to outputs panel

TomW1605:
feat: add show hidden option to outputs panel

adds an option to the outputs panel that lets you show any outputs that have been hidden by putting `_` at the start of their name.

i would like to add something that makes it so you can tell the "hidden" ones apart but that seems a bit more complex.

i ran this through the [existing test](https://docs.fluidd.xyz/development#run-unit-tests) and it passed everything but havnt had a chance to look at writing any new ones for this option

Signed-off-by: Thomas White <TomW1605@users.noreply.github.com>


TomW1605:
dont think im going to get around to adding the thing to display hidden ones differently but the _ at the start of the name should be enough. i also cant work out what tests are needed for this and cant actually find any existing tests for the ui to see what sort of stuff is being tested.

so im marking this as ready. if someone can point me at some existing ui tests or can give me guidance on what should be tested here i will work on that but i will do a new PR for the different look thing

TomW1605:
is there anything else that needs to be done to this before it can be merged?

pedrolamas:
Apologies, I just haven't had the time yet to review this, but will try to get that done before next release.

TomW1605:
no problem, just wanted to make sure there wasnt something i needed to do with it

src/components/widgets/outputs/OutputsCard.vue
```diff
@@ -6,6 +6,43 @@
...
+        </template>
+
+        <v-list dense>
+          <v-list-item @click="showHidden = !showHidden">
+            <v-list-item-action class="my-0">
+              <v-checkbox :input-value="showHidden" />
+            </v-list-item-action>
+            <v-list-item-content>
+              <v-list-item-title>
+                Show Hidden
```
pedrolamas, line 39: Strings like this need to be localizable, so it should move to "en.yaml" and consumed with `$t(app....)` so it can be translated

src/components/widgets/outputs/OutputsCard.vue
```diff
@@ -6,6 +6,43 @@
...
+            x-small
+            text
+            v-bind="attrs"
+            class="ms-1 my-1"
+            v-on="on"
+          >
+            <v-icon>
+              $cog
+            </v-icon>
+          </v-btn>
```
pedrolamas, line 29: Layout has been simplified

```suggestion
          <v-btn
            icon
            v-bind="attrs"
            v-on="on"
          >
            <v-icon dense>
              $cog
            </v-icon>
          </v-btn>
```

src/store/config/types.ts
```diff
@@ -105,6 +105,7 @@ export interface GeneralConfig {
   enableDiagnostics: boolean;
   thumbnailSize: number;
   colorPickerValueRange: ColorPickerValueRange;
+  showHidden: boolean;
```
pedrolamas, line 108: This should not be called `showHidden` as it is just for outputs, so instead I would suggest `showHiddenOutputs`

src/store/printer/getters.ts
```diff
@@ -648,7 +648,7 @@ export const getters: GetterTree<PrinterState, RootState> = {
 
       if (
         supportedTypes.includes(type) &&
-        (!filterByPrefix.includes(type) || !name.startsWith('_'))
+        ((!filterByPrefix.includes(type) || !name.startsWith('_')) || showHidden)
```
pedrolamas, line 651: Simplify

```suggestion
        (showHidden || !filterByPrefix.includes(type) || !name.startsWith('_'))
```

pedrolamas (commented):
I've finally managed to review this, in terms of code I have left a few things that need to change.

Functionality wise, this does make me a bit skeptical of its usage... Fluidd, Mainsail, and KlipperScreen have been moving towards the convention of hiding any and all elements that are prefixed with "_"

Users are aware of this, and the point is to allow to show/hide elements by using the configs, so in what scenario would this be used?

That also takes me to the other part which is that if we do this for Outputs, we probably should do the same for the rest, and that might not be as trivial as it was with Outputs here!

src/components/widgets/outputs/OutputsCard.vue
TomW1605: that seems resnable

src/store/config/types.ts
TomW1605: agree, i didnt think of that

src/store/printer/getters.ts
TomW1605: yep, that works better. will also change the name as mentioned above

pedrolamas: The name here is fine, as this is the `getOutputs` method, so in this specific context it is obvious we are referring to Outputs!

TomW1605:
> I've finally managed to review this, in terms of code I have left a few things that need to change.

thanks for taking the time to look this over. they all seem resnable, i will make the changes and test them this weekend

> Functionality wise, this does make me a bit skeptical of its usage... Fluidd, Mainsail, and KlipperScreen have been moving towards the convention of hiding any and all elements that are prefixed with "_"
> 
> Users are aware of this, and the point is to allow to show/hide elements by using the configs, so in what scenario would this be used?

i think there is a use case for this. mainly that it seems common to have outputs that you dont directly control 99% of the time that clutter up the interface so you hide them. this makes it easy to directly control these outputs without having to change the config files, something you can’t do at all while printing. it would also allow you to view non controllable outputs such as temp controlled fans that may be hidden much of the time as they are not normally something you would need to see.

> That also takes me to the other part which is that if we do this for Outputs, we probably should do the same for the rest, and that might not be as trivial as it was with Outputs here!

i agree it would probably be useful for other things, though the main other things i can think of that use the _ hide system are temp sensors and macros. macros already have a way to hide and show them though fluidd without having to edit config files and that can be changed while printing and temp sensors can be hidden from the graph though adding this option to that widget as well would probably be useful. in the case of the temp widget there is already a "settings" dropdown with selectable options so UI wise adding this option would work fine. i have not looked at the code side to know how practical it is there but i dont think that should stop this being merged. happy to take a look at that too but i feel it should be a seprate PR.

src/components/widgets/outputs/OutputsCard.vue
pedrolamas: Lets also make this "Show hidden outputs" to make it more descriptive and distinct from any other.

TomW1605: agree

TomW1605: ok, got the other things done. where do you recomend putting this in the translation files?

pedrolamas:
I just gave this a quick test in my simulated environment, the outputs card is not showing, so something is broken!

TomW1605:
do you get any errors or anything? it seems to be working fine for me (both before and after the update merge i just did)

pedrolamas:
No errors, just the Outputs card missing... and it is indeed because of the `get hasOutputs` on Dashboard.vue: it is missing `(false)` on each of the store getters in use!

What I have no idea is how it is working for you!

TomW1605:
ok, got it fixed, and worked out why it was working for me.

i forgot that the remote dev stuff in webstorm dosnt auto push to remote when you do a git pull. so my testing was running a version without the check in Dashboard.vue

TomW1605:
other than putting in the translation is there anything more that needs doing with this on my side or is it just waiting on having time? no problem if it is time, i understand your busy and this is a fairly now priority

TomW1605:
ok, @pedrolamas, i have added the text to the translation file in the best place i could work out. is there anything else that needs doing here?

PR #1239, opened 2023-11-17
docs: Add information about the translation platform

Daviteusz:
Weblate operational for 3 months now, there is no point in hiding it ;)


pedrolamas:
Thanks @Daviteusz, to be honest, the only reason we haven't pushed it even more is because we still haven't had Libre hosting approval!

Daviteusz:
@pedrolamas The chance that this project will not be approved is rather zero. Since it works, users should know about it :)

pedrolamas (approved):

PR #1245, opened 2023-11-24
feat: adds WebRTC (go2rtc) support

vajonam:
feat: webrtc go2rtc support

- initial support for go2rtc's WebRTC streams, low latency streams
- works without need an html page. allows pop-out. works on 
- Also supports MSE using mode in the url `https://go2rtc.domain.com/webrtc/stream.html?src=trident&mode=mse`

Signed-off-by: vajonam <vajonam@github>


src/components/settings/cameras/CameraConfigDialog.vue
```diff
@@ -85,7 +85,8 @@
             { text: $t('app.setting.camera_type_options.hlsstream'), value: 'hlsstream' },
             { text: $t('app.setting.camera_type_options.webrtc_camera_streamer'), value: 'webrtc-camerastreamer' },
             { text: $t('app.setting.camera_type_options.video'), value: 'ipstream' },
-            { text: $t('app.setting.camera_type_options.iframe'), value: 'iframe' }
+            { text: $t('app.setting.camera_type_options.iframe'), value: 'iframe' },
+            { text: $t('app.setting.camera_type_options.webrtc_gortc'), value: 'webrtc-gortc' }
```
pedrolamas, line 89: I know that Mainsail PR has not been merged yet, but we should attempt to use the same type here on both UIs.

```suggestion
            { text: $t('app.setting.camera_type_options.webrtc_gortc'), value: 'webrtc-go2rtc' }
```

src/store/cameras/types.ts
```diff
@@ -33,7 +33,7 @@ export interface CameraConfig extends CameraConfigWithoutId {
   id: string;
 }
 
-export type CameraService = 'mjpegstreamer' | 'mjpegstreamer-adaptive' | 'ipstream' | 'iframe' | 'hlsstream' | 'webrtc-camerastreamer' | 'device'
+export type CameraService = 'mjpegstreamer' | 'mjpegstreamer-adaptive' | 'ipstream' | 'iframe' | 'hlsstream' | 'webrtc-camerastreamer' | 'device' | 'webrtc-gortc'
```
pedrolamas, line 36:
```suggestion
export type CameraService = 'mjpegstreamer' | 'mjpegstreamer-adaptive' | 'ipstream' | 'iframe' | 'hlsstream' | 'webrtc-camerastreamer' | 'device' | 'webrtc-go2rtc'
```

src/components/widgets/camera/services/WebrtcGortcCamera.vue
```diff
@@ -0,0 +1,98 @@
+<template>
+  <video
+    ref="streamingElement"
+    autoplay
+    playsinline
+    controls
```
pedrolamas, line 6: We don't have `controls` on the other players, not sure we should have on this one either.

```suggestion
```

src/components/widgets/camera/services/WebrtcGortcCamera.vue
```diff
@@ -0,0 +1,98 @@
...
+  // webrtc player methods
+  // adapted from https://github.com/AlexxIT/go2rtc/blob/master/www/webrtc.html
+  // also adapted from https://github.com/mainsail-crew/mainsail/pull/1651
+
+  get url () {
+    const urlSearch = new URL(this.buildAbsoluteUrl(this.camera.urlStream || '')).search.toString()
+    const url = new URL('api/ws' + urlSearch, this.buildAbsoluteUrl(this.camera.urlStream || ''))
+    url.searchParams.set('media', 'video+audio')
+    // change protocol to ws
+    if (url.protocol === 'https') { url.protocol = 'wss' + ':' } else { url.protocol = 'ws' + ':' }
```
pedrolamas, line 35:
```suggestion
    url.protocol = url.protocol === 'https'
      ? 'wss:'
      : 'ws:'
```

vajonam: When I use a 2 in the name. Then the component name mapping fails. It maps with spaces in there. That's why I had to lose the 2.

src/components/settings/cameras/CameraConfigDialog.vue
vajonam: if I use the go2rtc, this doesn't map correctly. 

   ``` if (this.camera.service) {
      const componentName = `${this.$filters.startCase(this.camera.service).replace(' ', '')}Camera`

      if (componentName in CameraComponents) {
        return CameraComponents[componentName]
      }```

src/components/widgets/camera/services/WebrtcGortcCamera.vue
vajonam: controls is nice to have to allow us to pop out the player or cast it.. think its a value add in this case.

src/components/settings/cameras/CameraConfigDialog.vue
vajonam: image

See here the startCase mangles up the name adding spaces. What do you suggest here?

vajonam: I can replace the `' ',''`  again after it converts it to startCase?

src/components/widgets/camera/services/WebrtcGortcCamera.vue
pedrolamas: In that case, only enable it in full screen mode.

Having it on the card can easily interfere with the controls we add.

vajonam: also allows for audio to be unmuted, nice to have a feature that you just listen for a print to start and then mute.  but I am okay as long.

src/components/widgets/camera/services/WebrtcGo2RtcCamera.vue
```diff
@@ -0,0 +1,100 @@
...
+  get url () {
+    const urlSearch = new URL(this.buildAbsoluteUrl(this.camera.urlStream || '')).search.toString()
+    const url = new URL('api/ws' + urlSearch, this.buildAbsoluteUrl(this.camera.urlStream || ''))
+    url.searchParams.set('media', 'video+audio')
+    // change protocol to ws
+    url.protocol = url.protocol === 'https'
+      ? 'wss:'
+      : 'ws:'
+    return url.toString()
+  }
```
pedrolamas, line 39: I am struggling to see what the point of this bit is here... the whole objective is that users supply the stream URL, so in this case I think should be "ws://<host>:<port>/api/ws?src=<camera>&media=video+audio" so that this method would not be required.

vajonam: go2rtc doesn't obviously expose the wss/ws urls, it exposes only the http url links, so from a user perspective its easier to get those.

vajonam: example this is the streams page for a given stream.

image

pedrolamas: I've commented on the original Mainsail PR on this topic so we can get to a common ground here.

vajonam: updated your if statement missing colon! broke https support!

pedrolamas: That's what I get for coding in GitHub comments! 😅

vajonam: and me for reviewing it in!

pedrolamas (approved):

pedrolamas (changes requested):
Thank you @vajonam, can you please add a footer to the main PR comment with a "Signed-off-by" line, following the [CONTRIBUTING](https://github.com/fluidd-core/fluidd/blob/develop/CONTRIBUTING.md) guidelines.

vajonam:
Done.

pedrolamas (approved):

PR #1247, opened 2023-11-26
fix: ui cleanup

vajonam:
fix: ui cleanup

- center the toolhead controls, looks a bit neater.

image


vajonam:
@pedrolamas please close this if its not longer valid (I dont think this is)

PR #1252, opened 2023-11-29
feat: improves status panel layout

### LINKED ISSUE #1211: padding on print thumbnail making time switch between 1 and 2 lines
### Fluidd Version

1.26.0

### Browser

Chrome

### Device

Desktop PC

### Operating System

Windows

### What happened

When using a 4 column layout on my 27" 4k monitor the extra padding on the print thumbnail in the status widget means there is not enough space for the time estimates. But only just not enough, so every time the seconds changes from 1 to 2 digits it switches between 1 and 2 lines and pushes the whole column up and down and is very annoying. This padding is added by the button class. When the padding is removed there is enough room for the whole time on a single line

video of it changing as the time changes
https://github.com/fluidd-core/fluidd/assets/17092573/4f27b64e-f6d3-4a8d-aebe-90c89b5ce2e3

the extra padding because its a button
image

with the padding removed there is enough room for the time
image

no have padding taking up extra space. i dont think the padding on the button itself is nesaserry here because there is enough space around it from the layout of the status widget

### How to reproduce

make a column layout and change the window size so that the status widget just wide enough to show the thumbnail but not wide enough to show the full time. i get that this is pretty spesific but i dont think my particular setup would be that uncommon. it just takes a bit of messing with things to reproduce the issues without it

### Additional information

discord link: https://discord.com/channels/762863525818662974/762864050652053534/1165204358300962857

i have also had a look at fixing this by simply removing the padding and it works fine
https://github.com/TomW1605/fluidd/commit/e22472161eacb966ae72c7f4e98d51c98aa7cf0b

### LINKED ISSUE #1251: consider a differenet layout for cluttered printing status
### Fluidd Version

develop

### Browser

Chrome

### Device

Desktop PC, Laptop

### Operating System

Linux

### What happened

When printing this is what the status panel looks like.

image

I am not sure this is the best use of space along with circular progress bar

Consider changing the layout. Maybe we can get creative and put the thumbnail inside the circle for progress bar and then have the progress bar indicator on the outside border? or just even stack the thumbnail and progress bar.

image
 

### How to reproduce

Print in 3 panel mode. on largish monitor. 

### Additional information

None.

vajonam:
one possible alternate layout for the status panel.

## Before

image

## After

image

Fixes #1251
Fixes #1211 


src/components/widgets/status/StatusTab.vue
```diff
@@ -18,16 +18,35 @@
...
+              height="100%"
+              class="print-thumb"
+              @click="handleViewThumbnail"
+            >
+              <img
+                class="print-thumb"
+                :src="thumbnail"
+              >
+            </v-btn>
+          </div>
```
matmen, line 46: Enlarging the thumbnail doesn't work for me, this might need to be something like:

```suggestion
          <div
            class="progress-wrapper"
            style="cursor: pointer"
            @click="handleViewThumbnail"
          >
            <v-progress-circular
              :rotate="-90"
              :size="90"
              :width="7"
              :value="estimates.progress"
              color="primary"
              class="progress-circle"
            />
            <div class="thumbnail-background" />

            <div class="fill-height print-thumb">
              <img
                class="print-thumb"
                :src="thumbnail"
              >
            </div>
```

src/components/widgets/status/StatusTab.vue
```diff
@@ -435,4 +435,52 @@ export default class StatusTab extends Mixins(StateMixin, FilesMixin) {
...
+  opacity: 0.8; /* Adjust opacity as needed */
+}
+
+.percentage-wrapper {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-43%,200%);
+
+}
```
matmen, line 484: The indentation here is off - and please add a blank line before this section and remove the one after

matmen (changes requested):

matmen:
Thanks for your PR, I think the changes look nice. I've left a few comments in my review above

src/components/widgets/status/StatusTab.vue
vajonam: Not sure why the lint didn't catch that. will do.

src/components/widgets/status/StatusTab.vue
vajonam: @matmen  you can commit this to branch and I can test it out.

vajonam:
@matmen  that pushes the percentage wrapper, not sure how to address that?

src/components/widgets/status/StatusTab.vue
```diff
@@ -18,25 +17,41 @@
...
+              <img
+                class="print-thumb"
+                :src="thumbnail"
+              >
+            </div>
+            <div class="percentage-wrapper">
+              <span class="primary--text">{{ estimates.progress }}%</span>
+            </div>
+          </div>
         </v-col>
```
matmen, line 47: Sorry, looks like I forgot a line in my selection
```suggestion
                :src="thumbnail"
              >
            </div>
          </div>
          <div class="percentage-wrapper">
            <span class="primary--text">{{ estimates.progress }}%</span>
          </div>
        </v-col>
```

src/components/widgets/status/StatusTab.vue
```diff
@@ -2,9 +2,8 @@
   <div>
     <!-- Only show the linear progress for mdAndDown -->
     <v-progress-linear
-      v-if="
-        progressVisible &&
-          $vuetify.breakpoint.mdAndDown
+      v-if="progressVisible &&
+        $vuetify.breakpoint.mdAndDown
```
matmen, line 6: LGTM functionality wise now, but I'd prefer it if unrelated lines weren't changed - looks like these were auto-formatted? Applies to https://github.com/fluidd-core/fluidd/pull/1252/files#diff-8014f06344747bd0e9f2beb41bbca3f9938e86a5228dbdb110ece4fb4acc3860R52-R139 too

vajonam: oops was trying to fix the other .css formatting will roll it back.

vajonam:
should be sorted now!

matmen (approved):
LGTM, thanks!

src/components/widgets/status/StatusTab.vue
```diff
@@ -422,17 +420,71 @@ export default class StatusTab extends Mixins(StateMixin, FilesMixin) {
...
+  border-radius: 50%;
+  opacity: 0.8;
+  /* Adjust opacity as needed */
+}
+
+.percentage-wrapper {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-43%, 200%);
```
vajonam, line 488: @matmen  not to happy with this absolute positioning.. probably a better way to do this. Am sure @pedrolamas  when we gets to looking into this might have some other ideas. if I find a better way to do this without using numbers I will update it

vajonam: think i got it. added some rows. and center justify the content. looks clearer and less fiddly.

vajonam:
@matmen can you test this latest commit, because it makes up much less room, I am proposing we used in mdAndUp breakpoints.

matmen (dismissed):
LGTM!

pedrolamas:
I've made some tweaks and simplified the styles, will perform some more tests to ensure I haven't broken anything before merging down.

pedrolamas (dismissed):

src/components/widgets/status/StatusTab.vue
```diff
@@ -405,4 +395,18 @@ export default class StatusTab extends Mixins(StateMixin, FilesMixin, ToolheadMi
...
+  .progress-button {
+    width: 90px !important;
+    height: 90px !important;
+    border-radius: 50%;
+    overflow: hidden;
+  }
+
+  .progress-button-image {
+    max-width: 71px;
+    max-height: 71px;
```
matmen, line 408: With a 71px image, I've noticed the thumbnail is shifted 1px to the right and bottom, so I'd imagine this will need to be an even number:
image
image

matmen (commented):
I've tested the changes and they seem fine on my end, but the positioning of the image isn't quite centered now:

src/components/widgets/status/StatusTab.vue
pedrolamas: Very nice catch!! Took me a bit to repro this as I normally use a 4K 200% scale monitor and it worked fine there, only when I moved to the 1080p 100% scale one I could see the problem!

rursache:
please add a setting for this, i want to use the "old" layout. my thumbnails contain information in each corner which the new layout cuts off

vajonam:
can you post a screenshot of the old layout with your thumb info? just curious to see.

rursache:
> can you post a screenshot of the old layout with your thumb info? just curious to see.

here is an example of thumbnail elegoo cura is generating for each sliced model. if you put it inside a circle, you lose all the useful details.

knight_bust-300x300

how it looks in fluidd now:

Screenshot 2024-01-12 151353

how it looked before:

Screenshot 2024-01-12 151438 aaa

i think its obvious which one was better (at least for me and elegoo users). 

please add a setting toggle for this layout change. thanks!

vajonam:
I disagree there is no saved space, in the cases where there is longer filename it helps with the layout. However, I understand your issue. click on the thumbnail, will bring up full thumb. I get the issue.

pedrolamas:
@rursache I personally wasn't aware of this possibility though I can perfectly see your point.

I've discussed this with @vajonam and we are going to add an option to pick between the "classic" layout (the existing one where progress and thumbnail is separate) and the new "compact" layout (that matches the changes on this PR), that way users can choose which one is better for their purposes.

rursache:
@pedrolamas & @vajonam thank you for your effort and consideration!

rursache:
@pedrolamas was the option added? where can i find it and switch to the old layout? thanks

yw4z:
296250408-71128d49-9f5d-4302-87c4-8f7d22e1d1dd-2
hi just wanted to share ideas
• moved percentage to title bar
• used progress bar at full length below title bar and this allows to using gcode preview bigger
• added exclude as button. button type is different before

pedrolamas:
> @pedrolamas was the option added? where can i find it and switch to the old layout? thanks

Just created the PR!

rursache:
> 296250408-71128d49-9f5d-4302-87c4-8f7d22e1d1dd-2 hi just wanted to share ideas • moved percentage to title bar • used progress bar at full length below title bar and this allows to using gcode preview bigger • added exclude as button. button type is different before

looks great!

> > @pedrolamas was the option added? where can i find it and switch to the old layout? thanks
> 
> Just created the PR!

amazing!

PR #1260, opened 2023-12-06
feat: adds minimum_cruise_ratio support

pedrolamas:
Adds support for Klipper's new `minimum_cruise_ratio` (keeping `accel_to_decel` support as legacy) as specified here: https://github.com/Klipper3d/klipper/pull/6418


src/components/widgets/limits/PrinterLimits.vue
```diff
@@ -74,17 +74,33 @@
...
+            :value="minimumCruiseRatio"
+            :reset-value="defaultMinimumCruiseRatio"
+            :min="0"
+            :max="0.99"
+            :step="0.01"
+            :disabled="!klippyReady"
+            overridable
+            :locked="isMobileViewport"
+            :loading="hasWait($waits.onSetMinimumCruiseRatio)"
+            @submit="setMinimumCruiseRatio"
```
matmen, line 88: I feel like displaying the ratio in % is more intuitive, what do you think?

```suggestion
            :value="minimumCruiseRatio * 100"
            :reset-value="defaultMinimumCruiseRatio * 100"
            :min="0"
            :max="99"
            :step="1"
            suffix="%"
            :disabled="!klippyReady"
            overridable
            :locked="isMobileViewport"
            :loading="hasWait($waits.onSetMinimumCruiseRatio)"
            @submit="val => setMinimumCruiseRatio(val / 100)"
```

matmen (commented):
LGTM overall, just one small comment:

src/components/widgets/limits/PrinterLimits.vue
pedrolamas: I did consider it, but then I looked at the PWM output pins we have, and we set it from 0.00 to 1.00...

There is also the fact that ratio is supposed to be between 0 and 1, whereas percentage would be 0 and 100!

matmen (dismissed):

PR #1276, opened 2023-12-14
fix: corrects original source attributions

vajonam:
fix: add attribution and inspiration for the source circle control was adapted from. This was already in the original PR, making it more obvious by moving out the top. 


meteyou:
Many thanks for adding the links to Mainsail. I personally think the "heavily adapted" is just "dirty wording". I would consider "adapted for fluidd" to be "correct" in this case.

just my 2 cents...

PR #1278, opened 2023-12-17
feat: add feature to filter moonraker backup files

### LINKED ISSUE #1277: Filter moonraker backup files

There's a 'Filter Klipper backup files' option but this doesn't filter moonraker's backup files, e.g. `moonraker.conf.backup`

Would be nice to be able to filter out any backup files: moonraker, klipper etc. 

_No response_

### Additional information

fluidd 1.27

vajonam:
fixes #1277

- add filter for moonraker backup files.

### DIFF
--- a/src/components/widgets/filesystem/FileSystem.vue
+++ b/src/components/widgets/filesystem/FileSystem.vue
@@ -394,7 +394,7 @@ export default class FileSystem extends Mixins(StateMixin, FilesMixin, ServicesM
             }
             break
           case 'moonraker_backup_files':
-            if (file.type === 'file' && file.filename.match(/^.moonraker.conf.bkp$/)) { return false }
+            if (file.type === 'file' && file.filename === 'moonraker.conf.bkp') { return false }
             break
           case 'klipper_backup_files':
             if (file.type === 'file' && file.filename.match(/^printer-\d{8}_\d{6}\.cfg$/)) {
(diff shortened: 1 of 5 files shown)


src/components/widgets/filesystem/FileSystem.vue
```diff
@@ -393,7 +393,9 @@ export default class FileSystem extends Mixins(StateMixin, FilesMixin, ServicesM
               return false
             }
             break
-
+          case 'moonraker_backup_files':
+            if (file.type === 'file' && file.filename.match(/^.moonraker.conf.bkp$/)) { return false }
```
pedrolamas, line 397: You are using a regular expression here, so "." has to be escaped like "\." as is on the klipper backup files expression just below.

Nevertheless, we do not need a regular expression here, so we can just compare the strings.

```suggestion
            if (file.type === 'file' && file.filename === 'moonraker.conf.bkp') { return false }
```

vajonam: the backup file stats with a "." so cleaned up your update.

vajonam: the file starts with `.` so the hidden file filter will catch this, but I like to see hidden files but hide backup of moonraker, so I like this PR, but will leave it upto you.

pedrolamas: Good one, I missed the starting "."

PR #1279, opened 2023-12-18
feat: new SVG assets for circle toolhead controls

pedrolamas:
Replaces the current SVG assets that have been ported from Mainsail with news ones designed from scratch.

Some relevant changes:

- all assets are "static" (as in they are designed to be used "in situ" instead of having a global definitions and then applying runtime transformations to the asset)
- all MDI icons have been resized to have same dimensions
- new "xy center" icon (based on existing "focus" MDI icon)
- new "stepper motors off" icon (tries to resemble a stepper motor when looked from above)
- the 4 corner buttons have been reoriented to be aligned with the user perspective (thus resembling more the original Pronterface interface)

image

image


vajonam:
looks good!

matmen (approved):

PR #1289, opened 2023-12-23
feat(i18n-ko): Update Kor translations

crwusiz:


pedrolamas:
Hi @crwusiz thank you for sending this Pull Request.

I have a small problem though, as I now have a colliding translation to merge that was provided a few hours before [here](https://github.com/fluidd-core/fluidd/pull/1288/files) by @SangJin521

I do see some differences in terms such as the translation for "base" or "actual", so I am unsure which is "more" correct here...

Can I ask you both for help to get this to a common ground?

For what is worth, the current advice is for translations to be provided via [weblate](https://hosted.weblate.org/projects/fluidd/fluidd/ko/), as that will make it easier to merge.

pedrolamas:
Replaced by #1288

PR #1301, opened 2024-01-07
feat(spoolman): live update support

matmen:
Adds a WebSocket connection to listen for Spoolman spool object changes

### DIFF
--- a/src/components/widgets/spoolman/SpoolmanCard.vue
+++ b/src/components/widgets/spoolman/SpoolmanCard.vue
@@ -132,10 +132,6 @@ export default class SpoolmanCard extends Mixins(StateMixin) {
     this.$store.commit('spoolman/setDialogState', { show: true })
   }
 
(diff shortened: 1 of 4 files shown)


matmen:
@Donkie While implementing this, I've noticed there's no spool updates being fired when the spools filament object is updated.
Do you think it makes sense to fire update events cascadingly so a single WS connection suffices or does it make more sense to subscribe to the root endpoint and handle the state updates in the frontend (i.e. finding all spools with the associated filament ID and overwriting the vendor)?

Donkie:
Hmm, that's not something I've thought of. I feel like it would fire a lot of events unnecessarily if I changed the backend to cascadely send events to all spools if a filament changes. If it's easy for you I think it'd be better if you also subscribe to the filament of the spool if you're interested in changes in it.

pedrolamas (approved):

PR #1305, opened 2024-01-10
feat(spoolman): remember spool selection dialog sort key/order

### LINKED ISSUE #1181: Setting for default Spoolman sort order

The popup for selecting a spool will default to "Last Used" when this is rarely important in my opinion.
I can change it to "Location" (which is the only thing I personally care to sort by) but any page refresh will reset this.

Having a new option in the Spoolman Settings for the default sort order so when ever the window is opened that is how it is sorted. This would be the best option I feel as it'll allow you to change it temporarily if needed but the next time you open it it'll be back to how you prefer.

Secondarily would be to just have it remember how you left it, similar to how other tabs work. Though I don't feel this is as good of an option as the previous since you may only want to sort it another way (say by name) for the one time and want it to go back the next time you have the options come up.

### Additional information

_No response_

matmen:
Closes #1181

### DIFF
--- a/src/components/widgets/spoolman/SpoolSelectionDialog.vue
+++ b/src/components/widgets/spoolman/SpoolSelectionDialog.vue
@@ -93,14 +93,14 @@
             :custom-filter="filterResults"
             :no-data-text="$t('app.file_system.msg.not_found')"
             :no-results-text="$t('app.file_system.msg.not_found')"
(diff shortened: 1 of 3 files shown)


pedrolamas (approved):
I've moved the default sort to the default initial state so it is more consistent with the rest.

The rest is all good!

PR #1316, opened 2024-01-20
feat(spoolman): handle v2 responses, improve connection error handling

matmen:
Adds support for the v2 responses and events proposed in https://github.com/Arksine/moonraker/pull/792

Emits errors via toast notifications (on initial load) and shows a disconnected/out-of-sync state more clearly:
image

### DIFF
--- a/src/api/socketActions.ts
+++ b/src/api/socketActions.ts
@@ -734,7 +734,8 @@ export const SocketActions = {
       'server.spoolman.proxy', {
         params: {
           request_method: 'GET',
-          path: '/v1/spool'
+          path: '/v1/spool',
+          use_v2_response: true
         },
         dispatch: 'spoolman/onAvailableSpools'
       }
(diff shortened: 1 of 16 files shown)


pedrolamas (approved):

thijstriemstra:
Shouldn't "it" be a 1.28 version, 1.27.2 sounds like a minor minor update to 1.27.1 @pedrolamas

pedrolamas:
> Shouldn't "it" be a 1.28 version? 1.27.2 sounds like a minor minor update to 1.27.1 @pedrolamas

Indeed!!

When we release an updated version, we normally create a new patch Milestone just after it so we can start assigning stuff for the next release, and when we see it is getting bigger than a patch release, we just change it to the next minor version - which is what I am about to do now! 🙂

thijstriemstra:
thanks for the update. It's been quite a while since last release, any ETA for the next @pedrolamas?

pedrolamas:
KungFuPandaLiGIF

Issue #1317, opened 2024-01-21
Unable to download logs and config files when using auth/authz with moonraker


samwiseg0:
### Fluidd Version

fluidd: v1.27.1-2056453

### Browser

Chrome, Safari

### Device

Desktop PC

### Operating System

macOS

### What happened

When trying to download log/config files from fluidd while using authz with moonraker. Files cannot be downloaded. This has been an issue for some time but I have not had time to file this bug. 

In safari, a dummy file is downloaded with error:
```{"error": {"code": 401, "message": "Unauthorized", "traceback": "Traceback (most recent call last):\n\n  File \"/home/pi/moonraker-env/lib/python3.9/site-packages/tornado/web.py\", line 1767, in _execute\n    result = self.prepare()\n\n  File \"/home/pi/moonraker/moonraker/app.py\", line 578, in prepare\n    self.current_user = auth.check_authorized(self.request, self.endpoint)\n\n  File \"/home/pi/moonraker/moonraker/components/authorization.py\", line 807, in check_authorized\n    raise HTTPError(401, \"Unauthorized\")\n\ntornado.web.HTTPError: HTTP 401: Unauthorized (Unauthorized)\n"}}```

In chrome an error will popup. 

After adding a client as trusted there is no issue downloading files. This could be a moonraker issue but I am unsure as the request is coming from fluidd?

### What did you expect to happen

The log file to be able to be downloaded while having authz configured in moonraker. 

### How to reproduce

Configure moonraker to prompt non trusted clients to auth. 

Navigate to any log/config file in the printer data dir. Try and download.

matmen:
I can't reproduce this in either Firefox or Chrome on Windows. Can you share your moonraker.log, fluidd-access.log and your moonraker.conf (either by turning forced logins off or using scp)?

samwiseg0:
Sure here is all the logs. 
[unable_to_download_files.zip](https://github.com/fluidd-core/fluidd/files/14002687/unable_to_download_files.zip)

samwiseg0:
Just for clarity. I do not have `force_logins` set to true. I just omitted my client from the trusted clients list.

matmen:
Ahh, thanks for the clarification! I can reproduce it now, I'll take a look at it

samwiseg0:
> Ahh, thanks for the clarification! I can reproduce it now, I'll take a look at it

Awesome! Thank you!

PR #1318, opened 2024-01-21
fix: don't assume user is trusted when force_logins is turned off

matmen:
Fixes https://github.com/fluidd-core/fluidd/issues/1317.

I've tested theme loading and couldn't reproduce the issue described in https://github.com/fluidd-core/fluidd/pull/1065, so I don't think there's any side effects. Not sure why the `!forceLogins` check was required in the first place in https://github.com/fluidd-core/fluidd/commit/e5d177c854dc806c87c10c6568b04dfe74163b8a, @pedrolamas any chance you remember?

### DIFF
--- a/src/mixins/files.ts
+++ b/src/mixins/files.ts
@@ -12,9 +12,7 @@ export default class FilesMixin extends Vue {
   }
 
   get isTrustedUser (): boolean {
-    const forceLogins = this.$store.getters['server/getConfig'].authorization.force_logins
-
-    return forceLogins === false || this.$store.getters['auth/getCurrentUser']?.username === '_TRUSTED_USER_'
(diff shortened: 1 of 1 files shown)


pedrolamas:
> I've tested theme loading and couldn't reproduce the issue described in #1065, so I don't think there's any side effects. Not sure why the `!forceLogins` check was required in the first place in [e5d177c](https://github.com/fluidd-core/fluidd/commit/e5d177c854dc806c87c10c6568b04dfe74163b8a), @pedrolamas any chance you remember?

I think I just didn't think of the scenario we are talking here (unauthorized client) and thus assumed `forced_logins` needed to be `true`...

I've tested this PR and seems to work fine, so I think we are good to merge in.

pedrolamas (approved):

PR #1319, opened 2024-01-22
i18n: adds translation keys for System Utilization labels

jernejp21:
As you can see on image below, **load**, **awake time** and **memory used** are not translated. They are fixed string in the code. Changed fixed string with translatable code.

slika


pedrolamas (approved):

matmen (approved):

pedrolamas:
Thanks @jernejp21

Issue #1269, opened 2023-12-10
Spoolman integration: multiple extruders/spools


fweber3:
### Is your feature request related to a problem? Please describe

The combination of Fluidd/Moonraker/Klipper/Spoolman works really fine for me (Thanks everybody!).

However they currently only work with printers with a single extruder. I would like to use multi-extruder printers with the same ease.

### Describe the solution you'd like

1. I would like to see the current spool selection widget extended, so that I can select at least 2 spools _at the same time_, for the same print, one for each extruder. (This is my current personal need, although 10+ -extruder machines obviously exist). I would select spools A and B just like the current single selection, physically load both spools into the printer, and start the print.

2a. I already have gcode macros "T0"/"T1" for performing the extruder switch. These macros would be extended by a command "Now use filament A", "Now use filament B" .... referencing the previously /selected/ spools (and notably /not/ the backend's spool ID). 

2b. For a test: a command to change spools is already available (at the Moonraker level, IIRC) but it requires an actual, hardcoded spool ID. It worked beautifully from an "envisioned" workflow perspective, but hardcoded spool-IDs in gcode will nevertheless not be my way forward ;-)

### Describe alternatives you've considered

I had considered using the slicer (currently PrusaSlicer) for my spool management. However:
- I do not wish to duplicate spool mangement between spoolman /and/ the slicer. This is not feasible for >30 colors and several spools each.
- I usually pre-slice each model once, than print it many times with any/all of the 30 filaments(colors). Pre-Slicing lots of models to >30 files each just so that these include spool information is ... also not feasible. Even worse, the actual spool IDs sliced into the file keep changing quickly.
- In my workflow, slicing always implies a good check of the result. Slicing right before every print (to get a known spool) instead of pre-slicing would dramatically multiply the time spent on checks.

Therefore, the whole "Use the slicer" line of thinking seems a nogo for me due to matters of scale.

### Additional information

I am not really sure how my request would integrate with QR-code scanning. Along with the spool presented, the system would suddenly need information about which extruder the spool should belong to - which is a per-print attribute, not a per-spool attribute. Perhaps showing an additional "extruder number" QR-code /before/ the spool is an option? On single-extruder machines, the whole complication is superfluous and must be suppressed.

matmen:
The solution I'm thinking of looks like this:

- The `T0`..`Tn` macros will need to have an additional `variable_spool_id` (or similar, needs to be standardized across the frontends), defaulting to `None`
[...]

PR #1324, opened 2024-01-25
feat(spoolman): multi-tool support

matmen:
Closes https://github.com/fluidd-core/fluidd/issues/1269

Adds a new tool selection dropdown to the spoolman card when one or more `gcode_macro`s with a `spool_id` variable are detected. Emits a `SET_GCODE_VARIABLE` command to update the tools associated spool.

screenshot

[ Docs](https://github.com/fluidd-core/fluidd/assets/25269274/4b64343a-d8a4-4ad9-a8ba-ac8834f75bbe)

When a tool has a spool ID set, the tools "color dot" will be replaced by a filament spool icon in the corresponding filaments color (like in the spool selection modal).
Persisting settings available using Klipper's `[save_variables]` module.


pedrolamas:
> When a tool has a spool ID set, the tools "color dot" will be replaced by a filament spool icon in the corresponding filaments color (like in the spool selection modal).

The only issue I see with this is if the spool color is gray, it won't show up if Fluidd is using Dark theme (same for white spool in Light theme), hence why I was using that "dot" thing!

matmen:
The same issue exists in the spool selection dialog, I think this should be an easy fix by just adding an outline to the spool icon

pedrolamas:
> ...adding an outline to the spool icon

not sure if that is possible, but if you manage to do it, I am **very** interested! 😁

matmen:
I think this should work:

image
image
image
image

pedrolamas:
> I think this should work:

Sure, I personally can see it fine, though I am not sure that we can claim that it is "accessible" to every seeing user!

Also, a 16-bit screen or lower might be problematic... but granted, nowadays these are extremely rare! 😁

matmen:
Yeah, that was my thought too, but I couldn't really figure out how to make it accessible _and_ have it look decent. Even the current solution is kinda meh (especially on smaller screens), I think ideally we'd want a light or dark background color on the spool icon depending on the lightness of the filament color - or maybe even a multi-tone icon?

fweber3:
> Yeah, that was my thought too, but I couldn't really figure out how to make it accessible _and_ have it look decent. Even the current solution is kinda meh (especially on smaller screens), I think ideally we'd want a light or dark background color on the spool icon depending on the lightness of the filament color - or maybe even a multi-tone icon?

I'm not a big designer, but here's two triggers outside of your box:
How about using a contrasting color for accessibility BUT make it dashed/dotted (lines) or striped/polkadots (areas)? How about a strikethrough? ("big red cross", or something less in-your-face).

pedrolamas:
I am not sure what is the best course of action here in terms of layout (I too am not a designer), but I am happy to review this in terms of functionality and we can revisit this topic in a later change!

pedrolamas:
I admit I am troubled with the whole `SET_GCODE_VARIABLE`...

I honestly don't think we should be doing this from Fluidd side because if the user has a KlipperScreen or also uses Mainsail, it will not work until they too make the same changes...

In my view, this should either be done on Moonraker side, or somehow directly from Klipper side.

matmen:
> Any particular reason for that `macro.name.toUpperCase()`?

That's just so the macro name is displayed in upper case in the modals, in theory we can do this with CSS instead.
Either way should be fine I think, as currently only the "change spool" dropdown and spool selection modal consume the `targetableMacros` result.

> I admit I am troubled with the whole `SET_GCODE_VARIABLE`...
> 
> I honestly don't think we should be doing this from Fluidd side because if the user has a KlipperScreen or also uses Mainsail, it will not work until they too make the same changes...
> 
> In my view, this should either be done on Moonraker side, or somehow directly from Klipper side.

Well, having an API to set gcode macro variables directly would be optimal, but AFAIK there isn't one right now. I think the `spool_id` variable name should be fine across frontends, and if the functionality to set gcode macro variables via Moonraker gets added in the future this can still be refactored.

pedrolamas (approved):
Not fully convinced with the usage of the variables to store state (as that will not be reflected in other frontends), but I think we can go with this and revisit later if it comes to a common solution!

PR #1327, opened 2024-01-27
refactor(cameras): use cached url object for adaptive MJPEG sources

matmen:
Resolves an error when destroying an adaptive MJPEG camera feed:
image

Profiling showed this also improves `handleRefresh` performance by ~15% (1.97ms vs 1.67ms).

### DIFF
--- a/src/components/widgets/camera/services/MjpegstreamerAdaptiveCamera.vue
+++ b/src/components/widgets/camera/services/MjpegstreamerAdaptiveCamera.vue
@@ -24,6 +24,7 @@ export default class MjpegstreamerAdaptiveCamera extends Mixins(CameraMixin) {
   requestTime = 0
   timeSmoothing = 0.6
   requestTimeSmoothing = 0.1
+  url: URL | undefined
 
   handleImageLoad () {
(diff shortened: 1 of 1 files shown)


pedrolamas (dismissed):
Nice find! I made a minor change though, to just get perform a check on the `cameraImageSource` instead of adding a new `url` property. This way we only need to worry about a sole property to reset.

matmen:
@pedrolamas Moving the URL instantiation outside of the handleRefresh method was intentional, no point in creating a new object 30 times a second 😉 Maybe removing the `cameraImageSource` altogether and replacing it with a `url` object is the way to go here?

pedrolamas:
> Moving the URL instantiation outside of the handleRefresh method was intentional, no point in creating a new object 30 times a second 😉

Ah, fair point on that! This type of code has been in for ages, but indeed I see no reason why we shouldn't change it to a more "optimized" approach.

> @pedrolamas Moving the URL instantiation outside of the handleRefresh method was intentional, no point in creating a new object 30 times a second 😉 Maybe removing the `cameraImageSource` altogether and replacing it with a `url` object is the way to go here?

I tried it and doesn't work without other changes because the URL instance is always the same so Vue doesn't raise a change notification.

I'll refactor back to 2 properties!

pedrolamas (approved):

PR #1328, opened 2024-01-28
feat: improve support for adaptive bedmeshes

matmen:
Improves support for Klipper's new native adaptive bed meshes (https://github.com/Klipper3d/klipper/pull/6461 https://github.com/Klipper3d/klipper/pull/6473)

Shows a "virtual" bed mesh profile in the list and disables the "delete profile" button, temporary (/adaptive) profiles cannot be deleted.
image

### DIFF
--- a/src/store/mesh/getters.ts
+++ b/src/store/mesh/getters.ts
@@ -1,5 +1,12 @@
 import type { GetterTree } from 'vuex'
-import type { BedMeshProfile, MeshState, AppMeshes, KlipperBedMesh, KlipperBedMeshProfile, LegacyKlipperBedMeshProfile } from './types'
+import type {
+  MeshState,
+  AppMeshes,
+  KlipperBedMesh,
+  KlipperBedMeshProfile,
+  LegacyKlipperBedMeshProfile,
+  BedMeshProfileListEntry
+} from './types'
 import type { RootState } from '../types'
 import { transformMesh } from '@/util/transform-mesh'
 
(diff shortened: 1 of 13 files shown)


src/store/mesh/getters.ts
```diff
@@ -61,24 +68,42 @@ export const getters: GetterTree<MeshState, RootState> = {
       profiles.push({
         name,
         active: name === bedMesh.profile_name,
-        min,
-        max,
```
pedrolamas, line 65: Shouldn't `min` and `max` be kept here?

pedrolamas:
So far I only performed a quick test on this, but I'm not sure I understand why we need the new `BedMeshProfileListEntry` without `min` and `max` from `BedMeshProfile`... any specific reason for this new type?

src/store/mesh/getters.ts
matmen: It's not shown in the bed mesh profile list, but I think there's no harm in keeping it. I'll add those back later

pedrolamas: That's a very good point: I did a quick check, and I can't find it in use anywhere, so I see no reason why we can't just drop those properties!

pedrolamas:
I have now managed to catch up on this and even check the code changes on Klipper side.

Functionality wise, I have tested these changes in Fluidd and they worked perfectly, no issues at all!

From what I can see, Klipper makes no real distinction between a regular mesh and an adaptive mesh (except this naming convention that brings nothing special), and that is what is troubling me right now as I fail to see what increased value there is in indicating to the user that this is an adaptive mesh, which we are basing only on the mesh name!

matmen:
Adaptive bed meshes arent reported in the bed mesh profile list, so they don't show up in the list. This can be confusing to users, as there's an active bed mesh loaded, but the list doesn't show an active one (or no meshes at all if there's none defined in the config).

src/store/mesh/getters.ts
```diff
@@ -61,10 +68,20 @@ export const getters: GetterTree<MeshState, RootState> = {
...
-        range: Math.abs(min - max),
-        ...profile
+        adaptive: false,
+        range: Math.abs(min - max)
+      })
+    }
+
+    if (!(bedMesh.profile_name in klipperProfiles)) {
```
pedrolamas, line 76: We need to make sure that there is an active mesh set (should be enough to check for `profile_name`)

```suggestion
    if (bedMesh.profile_name && !(bedMesh.profile_name in klipperProfiles)) {
```

Without that check, this will happen:

image

matmen: Nice catch, thanks!

pedrolamas (approved):

PR #1330, opened 2024-01-29
fix: file path on upload

### LINKED ISSUE #1325: Missing slash in file editor
### Fluidd Version

1.27.1

### Browser

Firefox

### Device

Desktop PC

### Operating System

Windows

### What happened

I have a `macros` directory with a `print-end.cfg` file. When I edit the file, it shows a path without a slash: 

afbeelding

This doesn't seem to be happening to other sub-directories..

show `/macros/print-end.cfg`

### How to reproduce

create macros folder with file

### Additional information

_No response_

pedrolamas:
Main issue here is that some helper methods sometimes returned a "\" terminated path, which would then lead to these weird differences.

I made sure that these methods do not return a terminated "/"

Fixes #1325


matmen (approved):
LGTM

PR #1338, opened 2024-01-31
feat(spoolman): allow selecting table columns in spool selection dialog

### LINKED ISSUE #1336: Spoolman: Select fields shown in Fluidd

Is there any way to select the options from Spoolman that are shown in Fluidd?
eg i do not need the fields like charge, comment or location shown in Fluidd.
But I do not find the option to select the fields that are shown.
Anybody who can help me?
Greetings

no solution

_No response_

### Additional information

_No response_

matmen:
Closes #1336

image


src/store/config/state.ts
```diff
@@ -116,6 +116,14 @@ export const defaultState = (): ConfigState => {
...
           { value: 'time_added', visible: true },
           { value: 'time_in_queue', visible: false }
+        ],
+        spoolman: [
+          { value: 'filament_name', visible: true },
+          { value: 'id', visible: true },
+          { value: 'material', visible: true },
+          { value: 'location', visible: true },
+          { value: 'comment', visible: true },
+          { value: 'last_used', visible: true }
```
matmen, line 126: Not sure these are necessary, I think they're just for hidden-by-default columns?

pedrolamas: Indeed, all we need here is the hidden columns, so I would revert the changes on this file.

pedrolamas: I stand corrected: the entry IS required, but only needs the columns with `visible:false`, the rest is not required.

Given this, using `[]` should be ok!

matmen:
>... but honestly I would leave it as you currently have it on this PR.

Oops, only saw that now. If either way works I'd prefer the empty array tho, that way it's less changes across files when more columns are added.

matmen:
Oh weird, I didn't get a notification for that. I'll check it out right now!

pedrolamas (approved):

PR #1340, opened 2024-01-31
fix(thermals): size columns dynamically, disallow line breaks

matmen:
Before
image

After
image


pedrolamas (approved):
Nice one, never did like those 1% values there!

PR #1346, opened 2024-02-07
feat: adds print in progress status layout setting

pedrolamas:
This is a follow up on [this request](https://github.com/fluidd-core/fluidd/pull/1252#issuecomment-1889149738), specifically it adds a new setting to allow selecting the Print in Progress layout:

## New Print in Progress layout setting

image

## Default

image

## Compact

image

### DIFF
--- a/src/components/settings/GeneralSettings.vue
+++ b/src/components/settings/GeneralSettings.vue
@@ -170,6 +170,20 @@
       <v-divider />
 
       <app-setting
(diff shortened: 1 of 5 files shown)


thijstriemstra:
looks like the thumbnail needs to be (scaled) a bit smaller in order to fit in that circle?

pedrolamas:
We opted for clipping in the "compact" layout, but it is still clickable and that way you can see the full thumbnail image.

matmen (approved):

PR #1355, opened 2024-02-11
fix: don't override `gcode_macro` variables with stale values

### LINKED ISSUE #1269: Spoolman integration: multiple extruders/spools

matmen:
> When a macro is edited via the UI it saves a copy of it in the database, which is then used to override some default settings. It looks like this also overrides the macros variables, so they will never change as far as the UI is concerned.

Fixes #1269 not showing up

### DIFF
--- a/src/store/macros/getters.ts
+++ b/src/store/macros/getters.ts
@@ -25,8 +25,8 @@ export const getters: GetterTree<MacrosState, RootState> = {
           disabledWhilePrinting: false,
           color: '',
           categoryId: '0',
-          variables,
           ...stored,
+          variables,
           ...{ config }
         }
 


pedrolamas:
Indeed, this will fix this specific problem, but I have a feeling we are storing more than we should on the database (probably because of some of my refactorings...)

pedrolamas (approved):

matmen:
Yep, I agree - we should only be storing the values that can actually be changed through the UI..

PR #1356, opened 2024-02-11
fix: remove redundant data from stored macros

matmen:
Follow-up on https://github.com/fluidd-core/fluidd/pull/1355

### DIFF
--- a/src/store/macros/getters.ts
+++ b/src/store/macros/getters.ts
@@ -2,6 +2,14 @@ import type { GetterTree } from 'vuex'
 import type { Macro, MacrosState } from './types'
 import type { RootState } from '../types'
 
+export const MACRO_DEFAULTS = {
+  alias: '',
+  visible: true,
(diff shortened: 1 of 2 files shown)


src/store/macros/getters.ts
```diff
@@ -2,6 +2,14 @@ import type { GetterTree } from 'vuex'
 import type { Macro, MacrosState } from './types'
 import type { RootState } from '../types'
 
+export const MACRO_DEFAULTS = {
+  alias: '',
+  visible: true,
+  disabledWhilePrinting: false,
+  color: '',
+  categoryId: '0'
```
pedrolamas, line 10: Added missing order field

```suggestion
  categoryId: '0',
  order: undefined
```

pedrolamas (changes requested):
We need to also include the `order` field, without that it will not store the order changes!

pedrolamas (approved):

PR #1374, opened 2024-02-25
feat: allow percentage-based inputs in color picker

### LINKED ISSUE #1368: RGB 0-1 colorpicker support
I'm currently using a site like https://rgbcolorpicker.com/0-1 to convert the RGB values from fluidd colorpicker to valid 'klipper rgb' values.

Would be great if the colorpicker can provide rgb 0-1 values that I can use in the klipper RGB led config.

### Additional information

Current (1.28.1) implementation:

afbeelding

matmen:
Closes #1368

image image


src/components/ui/AppColorPicker.vue
```diff
@@ -327,6 +347,16 @@ export default class AppColorPicker extends Vue {
...
+
+  convertValueRange (value: number, direction: 'in' | 'out') {
+    if (this.valueRange === 'absolute') return value
+
+    let factor = 1
+    if (this.valueRange === 'percentage') factor = 255
+    if (direction === 'out') factor = 1 / factor
+
+    return Math.round(value * factor * 1000) / 1000
+  }
```
matmen, line 359: Not sure all this is really necessary, I've implemented it like this just in case there'll be more value ranges in the future, but I personally can't think of any (that would work with the number input, anyway). Using a boolean `usesPercentageValueRange` property instead would work fine too, and might save a tiny bit of memory

matmen:
Also, I'm not sure if there should be a setting (either server-wide or saved in the browser) to set a default value range. @thijstriemstra Would you mainly be using the `0..1` range and would remembering the last selection be useful in your case?

thijstriemstra:
I would refer to the klipper docs on `[neopixel]`(and `[led]`): https://www.klipper3d.org/Config_Reference.html#neopixel

```
#initial_RED: 0.0
#initial_GREEN: 0.0
#initial_BLUE: 0.0
#initial_WHITE: 0.0
#   Sets the initial LED color. Each value should be between 0.0 and
#   1.0. The default for each color is 0.
```

matmen:
@thijstriemstra I'm just asking if you want an option to remember the last choice (either `0..255` or `0..1`) so you don't have to switch to `0..1` every time you open the color picker (if you primarily use the `0..1` option) :D

thijstriemstra:
> you don't have to switch to 0..1 every time

oh, yes, good idea.

src/components/ui/AppColorPicker.vue
```diff
@@ -74,8 +74,27 @@
           />
         </v-layout>
 
+        <div class="text-right mt-4 mr-1">
+          <app-btn-group>
```
pedrolamas, line 78: Might make sense to use `<v-btn-toggle />` here instead?

src/components/ui/AppColorPicker.vue
```diff
@@ -74,8 +74,29 @@
...
+            >
+              0..255
+            </v-btn>
+            <v-btn
+              x-small
+              value="percentage"
+            >
+              0..1
+            </v-btn>
+          </v-btn-toggle>
```
pedrolamas, line 95: Works fine, but I wonder if we should make it match the style used around the rest of the app?

image

For this I replaced `v-btn` with `app-btn` and removed the `color="primary"` from the top

```suggestion
          <v-btn-toggle
            v-model="valueRange"
            mandatory
          >
            <app-btn
              x-small
              value="absolute"
            >
              0..255
            </app-btn>
            <app-btn
              x-small
              value="percentage"
            >
              0..1
            </app-btn>
          </v-btn-toggle>
```

pedrolamas (dismissed):

src/components/ui/AppColorPicker.vue
matmen: yep, let's keep it consistent! thanks

PR #1375, opened 2024-02-26
fix: make color picker use v-model and propsync

### LINKED ISSUE #1372: Fluidd resends `SET_LED` commands if led color is changed by macro or console command
### Fluidd Version

v1.28.1

### Browser

Edge

### Device

Desktop PC

### Operating System

Windows

### What happened

Invoking `SET_LED` from console will cause a 2nd `SET_LED` to be sent by Fluidd

Invoking `SET_LED` from console should just update Fluidd without sending any other commands.

### How to reproduce

- Open the led color picker and then close it again
- Send a `SET_LED` command from the console
- Observe the 2nd `SET_LED` command appearing

### Additional information

Reported by `fboc` (@fbeauKmi) in Fluidd Discord.

[klippy_vnothing.log](https://github.com/fluidd-core/fluidd/files/14393790/klippy_vnothing.log)

pedrolamas:
This PR refactors the `AppIroColorPicker` and `AppColorPicker` to use v-model and propsync where possible, thus making the control easier to consume, and fixing the reported issue.

Note: following up on what we already do with our custom input controls, changing the value on the text input boxes will not take effect unless <kbd>Enter</kbd> is pressed.

Fixes #1372

### DIFF
--- a/src/components/widgets/diagnostics/config/MetricsConfigStep.vue
+++ b/src/components/widgets/diagnostics/config/MetricsConfigStep.vue
@@ -75,7 +75,7 @@
 
               <app-setting :title="$t('app.setting.label.line_color')">
                 <app-color-picker
-                  :primary="metric.style.lineColor"
+                  :value="metric.style.lineColor"
                   :title="metric.name"
                   dot
                   @change="handleColorChange('lineColor', metric, $event)"
@@ -99,7 +99,7 @@
 
               <app-setting :title="$t('app.setting.label.fill_color')">
                 <app-color-picker
-                  :primary="metric.style.fillColor ?? metric.style.lineColor"
+                  :value="metric.style.fillColor ?? metric.style.lineColor"
                   :title="metric.name"
                   dot
                   @change="handleColorChange('fillColor', metric, $event)"
--- a/src/components/widgets/diagnostics/config/MetricsConfigStep.vue
+++ b/src/components/widgets/diagnostics/config/MetricsConfigStep.vue
@@ -78,7 +78,7 @@
                   :value="metric.style.lineColor"
                   :title="metric.name"
                   dot
-                  @change="handleColorChange('lineColor', metric, $event)"
+                  @input="metric.style.lineColor = $event"
                 />
               </app-setting>
 
@@ -102,7 +102,7 @@
                   :value="metric.style.fillColor ?? metric.style.lineColor"
                   :title="metric.name"
                   dot
-                  @change="handleColorChange('fillColor', metric, $event)"
+                  @input="metric.style.fillColor = $event"
                 />
               </app-setting>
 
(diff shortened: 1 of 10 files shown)


pedrolamas:
@matmen I know you also have an incoming PR with work on this area, I made it easy to integrate your changes in this one (just by using the new `currentRed`, `currentGreen`, `currentBlue` and `currentWhite` props)

matmen (changes requested):
I've opened [a PR](https://github.com/pedrolamas/fluidd/pull/12) on your repo that fixes the color settings not applying in the metrics settings, otherwise these changes LGTM

src/components/widgets/diagnostics/config/MetricsConfigStep.vue
```diff
@@ -99,10 +99,10 @@
 
               <app-setting :title="$t('app.setting.label.fill_color')">
                 <app-color-picker
-                  :primary="metric.style.fillColor ?? metric.style.lineColor"
+                  :value="metric.style.fillColor ?? metric.style.lineColor"
                   :title="metric.name"
                   dot
-                  @change="handleColorChange('fillColor', metric, $event)"
+                  @input="metric.style.fillColor = $event"
```
matmen, line 105: The fill color selection doesn't update when the color is changed - triggering an update for the whole style prop seems to fix this:
```suggestion
                  @input="metric.style = { ...metric.style, fillColor: $event }"
```

matmen (changes requested):

src/components/widgets/diagnostics/config/MetricsConfigStep.vue
pedrolamas: From my tests, I don't think this is necessary - this is why I changed the `fillColor` default from `undefined` to `null` as the previous is not reactive and the latter is!

matmen: That fixes it for new metrics, but the old ones have `undefined` stored in the moonraker DB ;)

matmen: well, no, of course they don't, as undefined isn't a JSON type.. i'll take another look at it :D

matmen: Ah yeah, the `fillColor` key is just missing in the stored data. Not sure if it makes sense to write an extra data migration step for this though, I think just updating the whole `style` prop is fine here

pedrolamas: I've added a migration step to make sure that any existing `fillColor = undefined` gets set to `null` instead, so that should solve the problem for existing data!

matmen (approved):

PR #1378, opened 2024-03-02
feat: add finish time selection from slicer or file

### LINKED ISSUE #1373: Finish Time (ETA) Selection, Slicer / File

Sometimes the "File" ETA is not precise. 

Currently, the Finish Time is estimated from the "File" time. Sometimes, this is somewhat off the real printing duration, where "Slicer" estimated time is more precise. It would be great if we could have the option to choose between File and Slicer for ETA. 

A simple radio button or pull-down selection menu on Settings would be sufficient. 

### Additional information

_No response_

MadeBaruna:
added options on general setting with file or slicer for finish time and progress 
close #1373 

file:

slicer:

### DIFF
--- a/src/components/settings/GeneralSettings.vue
+++ b/src/components/settings/GeneralSettings.vue
@@ -66,9 +66,7 @@
 
       <v-divider />
 
-      <app-setting
-        :title="$t('app.setting.label.confirm_on_estop')"
-      >
+      <app-setting :title="$t('app.setting.label.confirm_on_estop')">
         <v-switch
           v-model="confirmOnEstop"
           hide-details
@@ -79,9 +77,7 @@
 
       <v-divider />
 
-      <app-setting
-        :title="$t('app.setting.label.show_upload_and_print')"
-      >
+      <app-setting :title="$t('app.setting.label.show_upload_and_print')">
         <v-switch
           v-model="showUploadAndPrint"
           hide-details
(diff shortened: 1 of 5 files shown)


pedrolamas:
Hi @MadeBaruna, thank you for your work on this Pull Request, this was an absolute stellar job! 🙂

I made a couple of changes to it, but the most important is that I added the possibility to select more than one data source for the Print Progress and Print ETA (so when more than one is selected, we will display an average)

pedrolamas:
Here's how it currently looks on the settings side:

image

pedrolamas (approved):

matmen (approved):
Nice work, LGTM! Thanks

yw4z:
@pedrolamas what about adding an average of both as option

pedrolamas:
> what about adding an average of both as option

@yw4z just tick both checkboxes and it will calculate an average! 🙂

yw4z:
@pedrolamas nice to see you already added but using a radio button with separate options would be better to show there is an average option. also i will share a detailed design mockup soon for print status section :) which section is best to share at discussion / issues

pedrolamas:
@yw4z it literally says so on the description, not sure what else we could do here to improve it... 🤔

image

yw4z:
@pedrolamas people like me don't read much i guess :)

PR #1383, opened 2024-03-04
feat: adds support for machine peripherals

### LINKED ISSUE #1290: Simplify finding MCU IDs

With preconfigured OS images and pre-flashed boards becoming more and more common in printer kits, the only reason to SSH into the host is to run `ls /dev/serial/by-id/* ` It works fine, but it would be cool if this wasn't needed, and would simplify the process for newbies.

I was thinking that when editing .cfg files, an option in the right click menu that lists possible connected MCUs (which can be inserted when clicked on) would be a reasonably elegant solution - see poorly made mockup image
This could be set to only appear on lines starting with `serial: ` to reduce clutter.
Screenshot 2023-12-24 at 10 13 11 pm

_No response_

### Additional information

_No response_

pedrolamas:
Adds support for the new Moonraker Machine Peripherals (Devices)

CAN bus tab will appear if one or more CAN interfaces exist -> this bit has not been thoroughly tested as I do not have a CAN network myself that I can use for testing purposes.

## Shortcut on system info card

image

## Shortcut on file editor

image

## On open

image

## After refresh, no devices found

image

## After refresh, serial devices found

image

## After refresh, usb devices found

image

Resolves #1290


pedrolamas:
I am still iterating on the UI of this (trying expander panels, tabs, etc.), but this is what I am thinking in terms of dialog with expander panels:

# All collapsed

image

# Serial expanded

image

# USB expanded

image

thijstriemstra (commented):
NIce! I would not add the usb id in the title since some device names can be quite long, and it's also an important piece of info I'd rather see in the table, easier to find and copy/paste.

pedrolamas:
I honestly am a bit on a lost with how best to show this dialog and that it would look good in both desktop and mobile... for copy-paste I will probably just add some buttons on the values, but the whole look/organization of it is what I am more on a loss.

The expanders (as shown above) look a bit weird when they are all collapsed... Mainsail went for tabs, I am not fully on board with that too...

thijstriemstra:
> The expanders (as shown above) look a bit weird when they are all collapsed..

what about using tabs instead?

thijstriemstra:
> some device names can be quite long, 

e.g. a BTT U2C v2.1 shows up as `OpenMoko, Inc. Geschwister Schneider CAN adapter`

```
Bus 001 Device 005: ID 1d50:606f OpenMoko, Inc. Geschwister Schneider CAN adapter
```

pedrolamas:
Tabs seem to be ok'ish...

# On open

image

# After refresh, no devices found

image

# After refresh, devices found

image

src/components/common/PeripheralsDialog.vue
```diff
@@ -0,0 +1,428 @@
...
+      <v-tabs-items
+        v-model="tab"
+        touchless
+      >
+        <v-tab-item
+          v-for="peripheralGroup in peripheralGroups"
+          :key="peripheralGroup.type"
+        >
+          <template v-if="peripheralGroup.count == null">
+            Click on the refresh button to search for devices
```
matmen, line 64: I think this should be extracted into the language definitions

src/components/common/PeripheralsDialog.vue
```diff
@@ -0,0 +1,428 @@
...
+          </template>
+
+          <template v-else-if="peripheralGroup.count == 0">
+            No devices found
```
matmen, line 68: See above

src/components/common/PeripheralsDialog.vue
```diff
@@ -0,0 +1,428 @@
...
+
+          <template v-else-if="peripheralGroup.type === 'usb'">
+            <v-row
+              v-for="(device, index) in peripherals.usb_devices"
+              :key="index"
+            >
+              <v-col>
+                <v-card outlined>
+                  <v-card-title>{{ device.product ?? 'Unknown' }} ({{ device.vendor_id }}:{{ device.product_id }})</v-card-title>
+                  <v-card-subtitle>{{ device.description ?? 'Unknown' }}</v-card-subtitle>
```
matmen, line 158: Not sure if the `'Unknown'` should be extracted too

src/components/common/PeripheralsDialog.vue
matmen: I'm also wondering if we should just trigger the device search automatically when a tab is selected and there's no data for it yet, that would save the user a couple clicks

matmen (changes requested):
Functionality wise this LGTM, just a couple i18n related things:

src/components/common/PeripheralsDialog.vue
pedrolamas: > I think this should be extracted into the language definitions

Nice catch, that is now fixed!

> I'm also wondering if we should just trigger the device search automatically when a tab is selected and there's no data for it yet, that would save the user a couple clicks

I gave this a try, and though it works it is so fast that causes a weird flash and then a resize of the dialog. There is also the danger for CAN as from what I read, this should better be left as a manual process (single node ok, multi-node is bad!).

For now, I will leave this to the user to click "refresh".

src/components/common/PeripheralsDialog.vue
pedrolamas: Done!

matmen (approved):

PR #1389, opened 2024-03-13
feat: layout improvements for small screen

pedrolamas:
Improves layout in small screens:

- dialogs will be shown in full-screen.
- setting entries will break in 2 lines (title + content)

This will fix the problem with large dialogs (one can now scroll without affecting the background) and dismiss without a button (a close button is now shown on all dialogs).

## Settings page

Before | After
--- | ---
image | image

## Add User dialog

Before | After
--- | ---
image | image

## Bed Screws Adjust dialog

Before | After
--- | ---
image | image

## Add Camera dialog

Before | After
--- | ---
image | image

## Spool Selection dialog

Before | After
--- | ---
image | image

Before | After
--- | ---
image | image

### DIFF
--- a/src/components/common/AddInstanceDialog.vue
+++ b/src/components/common/AddInstanceDialog.vue
@@ -4,17 +4,11 @@
     max-width="320"
     :save-button-disabled="!verified"
     :valid.sync="valid"
+    :title="$t('app.general.title.add_printer')"
+    :help-tooltip="$t('app.endpoint.tooltip.endpoint_examples')"
     persistent
     @save="addInstance"
   >
-    <template #title>
-      <span class="focus--text">{{ $t('app.general.title.add_printer') }}</span>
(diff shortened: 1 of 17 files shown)


matmen (commented):
The spool selection and diagnostics card config dialogs both have a button in the bottom left that conflicts with the e-stop button, but I'm not sure what the best option would be to resolve that

Screen Shot 2024-03-13 at 18 22 47
Screen Shot 2024-03-13 at 18 22 55

pedrolamas:
Good catch... not sure yet how to fix this, but will get this back to draft and we can iterate on this matter.

pedrolamas:
I've now fixed the buttons alignment!

Basically, it will align them to the right on mobile view, and keep left on non-mobile as it is now!

image

image

matmen (approved):

PR #1397, opened 2024-03-20
fix: use Klipper can_extrude by default

CromFr:
Current conditions disabled extrusion controls on Fluidd if the klipper min_extrude_temp protection is set to a sub-zero temp.

For some context, I'm working on a ceramic printer, that does not have any heater nor thermistor on the hotend. The quick and easy way to disable any temp-related security on klipper is to set min_extrude_temp to -273.15, so that any temp detected by a connected (or in my case disconnected) thermistor won't prevent the extrusion.

While this works as expected with klipper (sending `G1 E10` does extrude), the extrusion controls on fluidd are disabled with this cheeky message: 
image

Note: I haven't looked at how min_extrude_temp is fetched by fluidd, maybe a value of 0 has a special meaning (i.e. a default value when undefined / could not be fetched?)

### DIFF
--- a/src/mixins/toolhead.ts
+++ b/src/mixins/toolhead.ts
@@ -20,8 +20,13 @@ export default class ToolheadMixin extends Vue {
     const activeExtruder = this.activeExtruder
 
     return (
-      activeExtruder !== undefined &&
-      activeExtruder.temperature >= activeExtruder.min_extrude_temp
+      activeExtruder?.can_extrude ??
+      (
+        activeExtruder !== undefined &&
+        activeExtruder.temperature >= 0 &&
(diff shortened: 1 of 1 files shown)


pedrolamas (approved):

pedrolamas:
Hi @CromFr, thank you for opening this Pull Request.

I've made a few changes as to use the `can_extrude` property that is already provided by Klipper as the defualt "source of truth" so that it will match whatever Klipper says, and fallback to current behavior if that property is not available (only for 2+ year old Klipper deployments)

CromFr:
That's a much better solution indeed !

PR #1402, opened 2024-03-29
feat: Allow extruder moves when force move is enabled

CromFr:
The FORCE_MOVE klipper gcode can be used for extruder motors, however the extrude/retract buttons on the web interface are disabled if the temperature is too low. This patch allows controlling the extruder motors when "force move" is enabled on the web interface (with red buttons like the axis buttons).

I haven't tested with multiple extruders, but I expect `this.$store.state.printer.printer.toolhead.extruder` to contain the name of the extruder that is controlled by the component

### DIFF
--- a/src/components/widgets/toolhead/ExtruderMoves.vue
+++ b/src/components/widgets/toolhead/ExtruderMoves.vue
@@ -29,9 +29,10 @@
       </v-col>
       <v-col cols="6">
         <app-btn
-          :disabled="!klippyReady || !extruderReady || !valid"
+          :color="(forceMove) ? 'error' : 'primary'"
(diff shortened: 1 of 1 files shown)


pedrolamas:
Hi @CromFr thank you for submitting this Pull Request.

I haven't validated this yet, but I wanted to say I currently have a WIP (work in progress) branch with a few changes for `Force Move`, as to replace the current usage with a bar control for each stepper.

Reason being that the current implementation only works for the default "cross" is the only one with Force Move, but I want to make this available for every type of toolhead control.

CromFr:
That's great news ! Feel free to close this PR if needed :)

PR #1404, opened 2024-04-02
feat: adds keyboard shortcuts

pedrolamas:
Adds support for keyboard shortcuts!

## Keyboard Shortcuts dialog (shown with <kbd>?</kbd>)

image

image

## Button tooltips (shown only if Keyboard Shortcuts are enabled)

image

image

image

## Settings entry

image


src/components/common/KeyboardShortcutsDialog.vue
```diff
@@ -0,0 +1,125 @@
...
+              <td><kbd>{{ keyboardShortcuts.timelapse }}</kbd></td>
+            </tr>
+            <tr>
+              <th>{{ $t('app.general.title.tune') }}</th>
+              <td><kbd>{{ keyboardShortcuts.tune }}</kbd></td>
+            </tr>
+            <tr>
+              <th>{{ $t('app.general.title.diagnostics') }}</th>
+              <td><kbd>{{ keyboardShortcuts.diagnostics }}</kbd></td>
+            </tr>
```
matmen, line 45: Should the history, timelapse and diagnostics ones be hidden when the component isn't configured or enabled?

matmen (approved):
LGTM, just one small question:

src/components/common/KeyboardShortcutsDialog.vue
pedrolamas: I actually first implemented like you describe... then for some odd reason I removed that!

I think it does makes sense to filter out what is not there, so I'll put the code back in!

pedrolamas: Done! 😁

matmen:
Both the pause and cancel shortcuts conflict with some firefox shortcuts (new private window and element picker respectively), and while <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>C</kbd> seems to work fine when printing (only opens the element picker when the printer is idle), <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd> always opens a new private window for me

pedrolamas:
I have a feeling I am opening a can of worms here... I'm going to check the keyboard shortcuts that Gmail uses and try to use those (should be safer, I guess...)

matmen:
> I've set `Pause` to Shift+p and Cancel to Shift+c, seems to work fine in Chrome/Edge and Firefox...

With just Shift+P and Shift+P I now can't type any upper case commands containing a C or P in the console while a print is running ;) I assume the event handler is just missing the `eventTargetIsContentEditable` call tho 😁

> I now got myself thinking... arrows to control the toolhead? 😁
> 
> Like Shift+arrow key and that would move the toolhead in the indicated direction by 10mm?

I guess that could work - maybe it could even use the configured move distances using the number keys, but that'd be a lot of combinations to remember haha

pedrolamas:
> With just Shift+P and Shift+P I now can't type any upper case commands containing a C or P in the console while a print is running ;) I assume the event handler is just missing the `eventTargetIsContentEditable` call tho 😁

Oops... fixed now! 🙂

matmen (approved):
LGTM

PR #1406, opened 2024-04-04
feat(FileSystem): remember last sort order

pedrolamas:
This will store the last user selected file sort order (per root) and restore it on reload of the page.

I also did a small refactor of the initial set of `state.uiSettings` to make it a bit more dynamic (we had already forgotten to add a few properties there, this should ensure that is now automatic)


matmen (approved):
LGTM

Issue #1408, opened 2024-04-08
Spoolman integration erroneously reports missing metadata


matmen:
Reported by @claudermilk in https://github.com/Arksine/moonraker/issues/838 (and various other users on Discord):

> ### What happened
> 
> Sometimes when starting a print job, after selecting a spool for the Spoolman integration the error message "The amount of filament required to print the selected file is unknown. Do you wish to continue?" is displayed.
> ### Client
> 
> Fluidd
> ### Browser
> 
> Chrome
> ### How to reproduce
> 
> After a print job is completed, start another one. It's intermittent, so no telling when it throws the error.
> ### Additional information
> 
> Issuing a firmware_restart clears the issue. I have seen this on two different printer installations; one just got a complete software and firmware re-install. I have seen this since the Spoolman integration was introduced. Running on RPi Debian Lite on both a RPi 3A+ and 4. Spoolman is a standalone server currently running the current 0.16 (was 0.13--same behavior).
> 
> The relevant log error line is: [common.py:build_error()] - JSON-RPC Request Error - Requested Method: server.files.metadata, Code: -32601, Message: Metadata not available for <filename.gcode>

So far I've only seen this happen when a file is moved/renamed/deleted, but the reprint tab hasn't updated its contents yet - but then it would also error when trying to start a print.

claudermilk:
Conveniently, one of the printers immediately did this. 
[moonraker.log](https://github.com/fluidd-core/fluidd/files/14908827/moonraker.log)

matmen:
Thanks. From the logs I can see fluidd requesting the metadata first and then the file being uploaded after (~20s later), but that doesn't make much sense. I assume the file was already on your host and you re-uploaded it after fluidd showed the error message?
Either way, the request looks fine to me, assuming the file isn't located in some sub-directory. Can I ask if you tried starting the print job via the Jobs page or via the reprint tab on the dashboard?

claudermilk:
It will do it whether I start from reprint or the Jobs panel. It seems more frequent with reprint. I have seen this with files in the "root" for gcodes and subdirectories within that. I am not re-uploading the file, I am using one already there.

 I've seen this on dozens of files sliced from both Super Slicer and Orca, many times from a file that just printed successfully; like the job finishes, I clear the plate, and hit print, then kaboom. So I know the file is fine, and it doesn't seem to have any connection with the specific slicer or version.

Arksine:
[...]

PR #1411, opened 2024-04-09
fix(spoolman): improve sanity check logic

matmen:
Functional Changes:
- Fixes file metadata retrieval  
  (we might want to move this logic into `SocketActions.serverFilesMetadata` to ensure a consistent behavior?)
- Fixes the printer state checks (sanity checks now only run when starting a print or changing the spool while an active print job is running)

Re-ordered some code for a clearer execution path

Closes https://github.com/fluidd-core/fluidd/issues/1408

### DIFF
--- a/src/components/widgets/spoolman/SpoolSelectionDialog.vue
+++ b/src/components/widgets/spoolman/SpoolSelectionDialog.vue
@@ -230,7 +230,15 @@ export default class SpoolSelectionDialog extends Mixins(StateMixin, BrowserMixi
 
       if (this.currentFileName) {
         // prefetch file metadata
-        SocketActions.serverFilesMetadata(this.currentFileName)
+        if (!this.currentFile && this.currentFileName.includes('/')) {
+          // if the file is in a subdirectory and isn't cached
+          // we need to populate the cache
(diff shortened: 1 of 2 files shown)


src/components/widgets/spoolman/SpoolSelectionDialog.vue
```diff
@@ -322,10 +330,15 @@ export default class SpoolSelectionDialog extends Mixins(StateMixin, BrowserMixi
...
 
-  get currentFile () {
-    const splitFilepath = this.currentFileName.split('/')
-    const filename = splitFilepath.pop()
+  splitFilePath (path: string): [string, string] {
+    const splitFilepath = path.split('/')
+    const filename = splitFilepath.pop()!
     const filepath = splitFilepath.join('/')
+    return [filename, filepath]
+  }
```
pedrolamas, line 338: I believe we can use the internal `getFilePaths` instead of this?

matmen: good point, should be fixed :)

pedrolamas (approved):

PR #1414, opened 2024-04-13
feat: print progress estimation improvements

pedrolamas:
- Restores the filament estimation calculation code
- Allows selection of absolute file position progress (`virtual_sdcard.progress` from Klipper)
- If relative file position progress selected and not available, fallback to `virtual_sdcard.progress`

### DIFF
--- a/src/store/printer/getters.ts
+++ b/src/store/printer/getters.ts
@@ -95,9 +95,9 @@ export const getters: GetterTree<PrinterState, RootState> = {
     }
   },
 
-  getFilePrintProgress: (state): number => {
+  getFileRelativePrintProgress: (state): number => {
     const { gcode_start_byte, gcode_end_byte, path, filename } = state.printer.current_file ?? {}
-    const { file_position } = state.printer.virtual_sdcard ?? {}
+    const { file_position, progress } = state.printer.virtual_sdcard ?? {}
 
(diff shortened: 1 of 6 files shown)


src/store/printer/getters.ts
```diff
@@ -111,25 +111,48 @@ export const getters: GetterTree<PrinterState, RootState> = {
...
     return state.printer.display_status.progress || 0
   },
 
+  getFilamentPrintProgress: (state) => {
+    const { filament_used, filename: statsFilename } = state.printer.print_stats ?? {}
+    const { filament_total, path, filename } = state.printer.current_file ?? {}
+
+    const fullFilename = path ? `${path}/${filename}` : filename
+
+    if (filament_used && filament_total && fullFilename === statsFilename) {
```
matmen, line 131: I think the `filament_used` check here should be a nullish check instead of a falsy check. Right now, when starting a print, the progress starts out at 5% for me (fallback to virtual_sdcard.progress), and then jumps to 1% as soon as filament is extruded.
Otherwise LGTM!

matmen (changes requested):

src/store/printer/getters.ts
pedrolamas: Ah, very nice find, I've added the check!

matmen (approved):

PR #1418, opened 2024-04-18
feat: show all fan and pin values as percentages

pedrolamas:
Ideally, limited scale values in the front-end should be presented as a percentage instead of the Klipper value - that is for config and console purposes.

The changes in this pull request are as follows:

- non-controllable (read-only) fans and output pins will show "Off" when value is 0, otherwise presents the value on a scale of 0% to 100%.
- controllable pwm output-pins will always present the value on a scale of 0% to 100%.

## Before

image

## After

image

### Partial test config used for the above:

```ini
[heater_fan heater_fan]
pin: PD4
off_below: 0.2
shutdown_speed: 0.0
max_power: 0.6
fan_speed: 0.85

[output_pin test_pwm_pin]
pin: PD7
pwm: True
value: 0.6
scale: 0.8
cycle_time: 0.1
```

### DIFF
--- a/src/components/ui/AppNamedSwitch.vue
+++ b/src/components/ui/AppNamedSwitch.vue
@@ -12,6 +12,8 @@
       v-model="inputValue"
       class="mt-0"
       :disabled="disabled || loading"
+      :true-value="trueValue"
+      :false-value="falseValue"
       hide-details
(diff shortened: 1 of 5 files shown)


matmen (approved):
LGTM

PR #1427, opened 2024-04-29
feat: backup and restore settings

### LINKED ISSUE #1113: Export the configuration file of Fluidd for easy recovery when reinstalling

Just a while ago, I reinstalled the klipper family bucket including Fluidd due to some update compatibility issues. I tried backing up Fluidd configuration files before reinstalling (for example, dashboard layouts, temperature presets, macro categories, etc.), but I couldn't find an option to do this or where to store the configuration files. I had to reconfigure Fluidd after reinstalling, which is actually a rather tedious job, and I may even forget some of the previous settings.

Add a configuration file export option in Fluidd's settings interface to facilitate recovery after reinstallation. Or, friends who know it can tell me what the storage path of the configuration file is, and it is also possible.

_No response_

### Additional information

_No response_

pedrolamas:
Add a new Backup and Restore buttons to easily export and import Fluidd settings across multiple printers.

As a first implementation, this will backup/restore ALL Fluidd settings (there is no selections of specific parts of the settings).

image

Resolve #1113

### DIFF
--- a/src/components/layout/AppSettingsNav.vue
+++ b/src/components/layout/AppSettingsNav.vue
@@ -44,7 +44,7 @@ export default class AppSettingsNav extends Vue {
       { name: this.$tc('app.setting.title.camera', 2), hash: '#camera', visible: true },
       { name: this.$t('app.setting.title.tool'), hash: '#toolhead', visible: true },
       { name: this.$t('app.setting.title.thermal_presets'), hash: '#presets', visible: true },
-      { name: this.$t('app.setting.title.gcode_preview'), icon: '$cubeScan', hash: '#gcodePreview', visible: true },
+      { name: this.$t('app.setting.title.gcode_preview'), hash: '#gcodePreview', visible: true },
       { name: this.$t('app.general.title.timelapse'), hash: '#timelapse', visible: this.supportsTimelapse },
       { name: this.$t('app.spoolman.title.spoolman'), hash: '#spoolman', visible: this.supportsSpoolman },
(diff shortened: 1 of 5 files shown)


matmen (approved):
LGTM functionality wise, but what do you think of adding an additional key to the export (e.g. `"fluidd": "v1.2.3"`) and using it to check if the selected backup file actually is a fluidd config export before restoring it?

pedrolamas:
I like the idea of adding the version to it, that makes complete sense to me!

Confirming if it is a Fluidd backup or not is tricky, as this is just a JSON file... granted we could have some schema to check against but that makes versioning very tricky!

matmen:
Well, using a `fluidd` key we could at least check if that's there - not a high chance a random json file will have one ;) No need to validate the entire schema IMO

pedrolamas:
> Well, using a `fluidd` key we could at least check if that's there - not a high chance a random json file will have one ;) No need to validate the entire schema IMO

Agreed! 🙂

pedrolamas:
Following the comments, I made output filename is now "backup_fluidd_<version>_<instancename>.json"

The JSON blob is now using this format:

```json
{
  "meta": {
    "app": "Fluidd",
    "version": "1.2.3",
    "type": "some string identifier"
  },
  "data": {}
```

We will verify that `meta.app === 'Fluidd'` and that `meta.type === "<identifier>"` so that we can match the expected format!

Should be enough to validate the expected format and that version will allow us flexibility in the future if we want to!

matmen (approved):
LGTM

Escrich:
Good Job!, great job!

PR #1445, opened 2024-06-02
feat: show print progress when status panel is collapsed

### LINKED ISSUE #1432: Put progress bar in the folded up version of print status

When the print job status is folded, no way to see progress

Put progress bar in the folded up version of print status

Instead of a progress bar just put a % or time left

### Additional information

_No response_

matmen:
Adds a linear print progress bar to the status panel when it's collapsed, so the current progress is always visible

**Before**
grafik

**After**
grafik

---

Closes #1432


pedrolamas:
I had previously taken a quick stab at this and stopped when I noticed that it would break the corners style... seems you are hitting the same issue here:

image

matmen:
Yep! I fixed that locally (should have been commit https://github.com/fluidd-core/fluidd/commit/8c651f46dded0064a76e04bd193f2885ddd037dd), but looks like I pushed the wrong file. I'll push the correct one when I get home and let you know 😉

matmen:
@pedrolamas should be good to go :)
image

matmen:
well, looks like there still is some kind of bug related to the v-if in the slot template. for whatever reason the `hasCollapsedContentSlot` getter only gets called on component initialization, but not when the isCollapsed state changes, resulting in unexpected behavior. changing the getter to a function fixes this, but also makes vue re-evaluate the expression every second or so, not sure if that's just because of the component re-rendering tho..

pedrolamas (approved):
LGTM!

PR #1448, opened 2024-06-11
feat: show warning for Trinamic OTPW flag

### LINKED ISSUE #1105: TMC Drivers: Report OTPW
Most (or all?) TMC drivers report OTPW (over-temperature pre warning) conditions. This happens before actual over-temperature occurs & the driver shut down.
This information can be queried through TMC registers. All drivers I'm aware of report the `otpw` flag in `DRV_STATUS`, which can be read through `DUMP_TMC STEPPER={stepper} REGISTER=DRV_STATUS`. The result of this will look something like this if an overtemperature pre-warning is raised:
```
DRV_STATUS: 84004008 sg_result=8 stealth=1 otpw=1(OvertempWarning!) stst=1
```

Currently, users are unaware of these warnings, unless the pre-warning turns into an overtemperature condition, and their printer shuts down.

### Proposed solution

It'd be great if this information could be shown in the UI.  
I think it'd probably fit best in the notifications, where e.g a Raspberry Pi overtemperature warning would be shown.  

In certain cases, it might be desirable to ignore these warnings - for example, if the user made sure that their drivers don't heat up past the overtemperature pre-warning.

pedrolamas:
Shows a notification message when any `TMCxxxx` driver reports an `otpw` flag (over-temperature pre warning)

image

The link on the notification is currently pointing to https://www.klipper3d.org/TMC_Drivers.html#tmc-reports-error-ot1overtemperror

Resolves #1105

### DIFF
--- a/src/store/helpers.ts
+++ b/src/store/helpers.ts
@@ -1,6 +1,35 @@
 import type { Commit, Dispatch } from 'vuex'
 import type { RootState } from './types'
 import { SocketActions } from '@/api/socketActions'
+import type { AppPushNotification } from './notifications/types'
+
+export const handleTrinamicDriversChange = (payload: any, state: RootState, dispatch: Dispatch) => {
+  for (const item in payload) {
+    const [type, nameFromSplit] = item.split(' ', 2)
+
+    if (
+      /^tmc\d{4}$/.exec(type) &&
+      payload[item]?.drv_status?.otpw != null &&
+      state.printer.printer?.[item]?.drv_status?.otpw == null
+    ) {
+      const name = nameFromSplit ?? item
+
+      const notification: AppPushNotification = {
+        id: `${item}-otpw`,
+        title: `Stepper driver '${name}' is over-heating`,
+        description: 'This may lead to a failed print',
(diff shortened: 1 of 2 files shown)


src/store/helpers.ts
```diff
@@ -1,6 +1,35 @@
...
+    if (
+      /^tmc\d{4}$/.exec(type) &&
+      payload[item]?.drv_status?.otpw != null &&
+      state.printer.printer?.[item]?.drv_status?.otpw == null
+    ) {
+      const name = nameFromSplit ?? item
+
+      const notification: AppPushNotification = {
+        id: `${item}-otpw`,
+        title: `TMC '${name}' is over-heating`,
```
matmen, line 19: I feel like this should be something like `Stepper driver 'name' is over-heating`, chances are end users don't know what TMC means

matmen (approved):
LGTM functionality wise, just one small concern:

src/store/helpers.ts
pedrolamas: Good point, I updated the description to your suggestion!

NyxCode:
Awesome stuff, greatly appreciate the work you put into fluidd!

NyxCode:
image

might have even saved my 5160s!

PR #1470, opened 2024-07-26
feat: improve fan and outputs panel layout

HelgeKeck:
feat: My feature.
This PR improves tha panel layout for the fans and outputs, it also groups the devices and makes the collapseable

Some description.

Would be nice to make the collapsed state persistent. some help would be appreciated

Signed-off-by: Helge Keck helgekeck@hotmail.com


HelgeKeck:
image

thijstriemstra:
I'm not sure about "controllable fans", because all fans are controllable right. Not sure yet what to suggest either.

HelgeKeck:
> I'm not sure about the "controllable fans" label, because all fans are controllable right. Not sure yet what to suggest either.

its probably not the correct word i choosed. but i think having them grouped by fans that the suer can control and other fans is nebeficial, especially if you have mroe than one fan
image

ningpj:
"Adjustable fans" perhaps

PR #1471, opened 2024-08-01
feat: load default template at startup if available in .fluidd-theme folder

HelgeKeck:
feat: Loading a default template at startup if available

So i want to load a default template json file at the first app start, same files that come from the backup and restore feature, 
This PR isnt really working, but i hope to get some help here.

Problem at is that the getCustomThemeFile doesnt return any root files files for any reason and i cant find out why.

Signed-off-by: Helge Keck helgekeck@hotmail.com


HelgeKeck:
so, i have indeed now a working version. definitly needs some more love from someone who understands that better.

it loads a default.json from the .fluidd-theme folder at the first app start. atm its just a backup and restore type json file

but it works

PR #1484, opened 2024-08-15
fix: null check cards on layout init

### LINKED ISSUE #1473: Web interface not visible
### Fluidd Version

v1.30.1

### Browser

Chrome

### Device

Laptop

### Operating System

Windows

### What happened

If I try to go to fluidd in my browser, it looks like it is fully loaded, with the logo at the top of the tab, etc. but the browser page is blank, with no UI visible. I even tried reinstalling Fluidd, and it did not help

I expect to be able to see and be able to interact with the Fluidd web interface

### How to reproduce

I think I changed the trusted_clients option in the moonraker.conf to my laptops IP address, but it did work after that. I also installed Sonar. Then it randomly stopped working. When I tried to reinstall it, it still did not work. Even removing Sonar did not help.

### Additional information

_No response_

pedrolamas:
We have had a few reports from users that Fluidd stops working after editing the dashboard layout.

It seems that somehow the layout data is getting incorrectly recorded in the database with `null` values for some cards.

image

Though I don't yet know the reason for this happening, the proposed fix here is to filter these `null` values out before we attempt to load the layout.

Fixes #1473

### DIFF
--- a/src/store/layout/mutations.ts
+++ b/src/store/layout/mutations.ts
@@ -32,13 +32,17 @@ export const mutations: MutationTree<LayoutState> = {
       // migrate existing layouts
       const migratableLayoutKeys = ['dashboard']
 
-      for (const layoutKey of Object.keys(payload.layouts)) {
+      for (const [layoutKey, currentLayout] of Object.entries(payload.layouts)) {
+        for (const [containerKey, components] of Object.entries(currentLayout)) {
+          currentLayout[containerKey] = components
+            .filter(card => card != null)
+        }
+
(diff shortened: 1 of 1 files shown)


matmen (approved):

pedrolamas:
Initial fix didn't solve the problem as the `null` value filtering needs to be done a bit earlier, so I moved the code above.

matmen (approved):
I won't have access to a machine running fluidd for the next couple of days, so I can't test it functionality wise, but the code LGTM

PR #1494, opened 2024-09-15
Update moonraker_conf.md

PeanutJ19:


pedrolamas:
Hi @PeanutJ19, not sure what you were trying to do here, but we do not want to default the `force_logins` to true.

PR #1499, opened 2024-09-30
feat: show application as soon as possible

pedrolamas:
Present Fluidd page as soon as possible instead of waiting for socket connection.


matmen:
I tested the changes and they look fine to me (at least nothing appears to be broken 😛), but just wondering: is this solving any specific issue? I haven't noticed fluidd taking "too long" to load or anything - I believe it's always shown the app even if the socket wasn't connected yet?

pedrolamas:
The current behavior is to try and connect the socket **before** we even show the UI, and that translates in a 3s delay if the address is down.

The new behavior is to show the page with an "indeterminate" progress bar loading while we do the connection.

new

On the above, left is new behavior, right is the old behavior. I pressed F5 on both sides at the same time.

matmen (approved):
I can't really notice any difference on my end (granted I've never had issues with this), but it doesn't look like this broke anything, so LGTM ;)

PR #1500, opened 2024-10-01
fix: disable "Update All" button when an invalid repo is detected

matmen:
The "Update All" action updates all components sequentially and fails if there's an invalid repository up next. When the first updatable repo in the list is an invalid one, this would just refresh fluidd without actually doing anything, resulting in poor UX and confused users (me 🤣). Not sure if this also affects dirty repos, I haven't had time to test that yet.

Not sure this approach is better, because it just disables the update button without showing a reason for it. Maybe the button should have a tooltip showing the reason for it being disabled too?

https://github.com/user-attachments/assets/0a6701db-e7ad-494d-9b70-d2cc9a25f868


pedrolamas (approved):
LGTM!

PR #1527, opened 2024-11-13
feat: Nevermore Controller support

pedrolamas:
Adds full support for Nevermore Controller.

image


src/store/printer/getters.ts
```diff
@@ -760,7 +761,6 @@ export const getters: GetterTree<PrinterState, RootState> = {
       'aht10',
       'bme280',
       'htu21d',
-      'nevermoresensor',
```
SanaaHamel, line 763: This will break VOC plotting on the temp graph.

pedrolamas: This is just extra sensor data (does not include temperature, just pressure, humidity and gas)

SanaaHamel:
Note that there can be both `nevermore<named>` and `nevermore` objects, depending on how they are declared in the cfg.

pedrolamas:
> Note that there can be both `nevermore <named>` and `nevermore` objects, depending on how they are declared in the cfg.

Yes, I made sure to handle both as shown on the screenshot

pedrolamas:
@SanaaHamel This is the test objects state I am using (and how I generated the screenshot bove):

```json
      nevermore: {
        rpm: 10,
        speed: 20,
        intake_temperature: 20,
        exhaust_temperature: 30,
        intake_humidity: 40,
        exhaust_humidity: 50,
        intake_pressure: 60,
        exhaust_pressure: 70,
        intake_gas: 80,
        exhaust_gas: 90
      },
      'nevermore test': {
        rpm: 10,
        speed: 20,
        intake_temperature: 20,
        exhaust_temperature: 30,
        intake_humidity: 40,
        exhaust_humidity: 50,
        intake_pressure: 60,
        exhaust_pressure: 70,
        intake_gas: 80,
        exhaust_gas: 90
      }
```

However, I was looking at the controller code and noticed a `_min` and `_max` for each entity, are those also reported as part of the Klipper status object?

Also, am I correct ti say "pressure" and "gas" are fully int values (have no decimals)?

pedrolamas:
> They are. These were requested by Mainsail for future use. These indicated the upper/lower bound seen this klipper session.

I've added support for them, they will show as tooltips for the gas, temperature, humidity, and pressure.

SanaaHamel (approved):
Looks sane. Passed smoke test w/ my production unit.

pedrolamas:
Thanks for testing and approving, @SanaaHamel! 🙂

PR #1528, opened 2024-11-13
feat: adds config file #region folding support

pedrolamas:
Adds support for regions in Klipper and Moonraker configuration files:

- `#region` will start a new region
- `#endregion` will close the last open region

Regions can contain other regions inside and the editor will collapse them correctly.

## Unfolded

image

## Folded

image


matmen (approved):
Nice, LGTM!

PR #1533, opened 2024-11-18
fix: toolhead position controls enabled on pause

### LINKED ISSUE #1531: Can't type in toolhead position fields whilst paused
### Fluidd Version

v1.30.5-c6996cc

### Browser

Firefox

### Device

Laptop

### Operating System

Linux

### What happened

The toolhead position fields are set to read-only whilst a print is paused.

The toolhead position fields should not be read-only whilst a print is paused, as the toolhead can already be moved with the arrow buttons.

### How to reproduce

Pause a print, note that the position fields are enabled but read-only.

### Additional information

The fields are disabled when the printer is "busy", which includes being paused:

https://github.com/fluidd-core/fluidd/blob/ab16ea2066e5259b8e9f112d108960672d429c83/src/components/widgets/toolhead/ToolheadPosition.vue#L26

https://github.com/fluidd-core/fluidd/blob/ab16ea2066e5259b8e9f112d108960672d429c83/src/mixins/state.ts#L52-L61

pedrolamas:
Only disable absolute position input controls when printer is in "printing" state (so we will keep then available if printer is "paused" or "busy")

Fixes #1531

### DIFF
--- a/src/components/widgets/toolhead/ToolheadPosition.vue
+++ b/src/components/widgets/toolhead/ToolheadPosition.vue
@@ -23,7 +23,7 @@
           small
           type="number"
           :disabled="!klippyReady || (!xHomed && !xForceMove)"
-          :readonly="printerBusy"
+          :readonly="printerPrinting"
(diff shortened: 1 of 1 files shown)


matmen (approved):
LGTM

PR #1539, opened 2024-11-23
feat: column reordering

### LINKED ISSUE #1538: Column Order

Is it possible to add the possibility of placing columns in a custom order? 

In the task panel I would like to change the order of some columns. Could it be possible to add the possibility of clicking on the header of a column and moving it to the left or right as needed?

_No response_

### Additional information

_No response_

pedrolamas:
Adds column drag & drop to reorder on the column picker.

This applies to:

- File Browser
- Job Queue
- Job History
- Spoolman selection dialog

image

Resolves #1538


pedrolamas:
@matmen regarding Spoolman, are those all the columns we can have on that dialog or are there more we can provide so users can pick & choose?

matmen:
There's a bunch more fields that are supported OOTB (like density, diameter, extruder/bed temp), but I'm not sure it makes sense to display them in the spool selection modal, though it would probably make sense to support them and hide them by default? I'm planning to add the same functionality for the dashboard card in the future, as well as support for showing custom fields defined in Spoolman itself.
Not sure why, but when I open the spool selection modal with this patch applied none of the headers except for "Filament" have a name? All the other components seem fine.
image

pedrolamas:
> Not sure why, but when I open the spool selection modal with this patch applied none of the headers except for "Filament" have a name? All the other components seem fine.

That's very odd, should work just fine... are you using German as base language?

src/store/config/getters.ts
```diff
@@ -90,25 +90,27 @@ export const getters: GetterTree<ConfigState, RootState> = {
...
+          visible: true,
+          ...headers.find(p => p.value === header.value),
+          ...header,
+        }))
+      : headers
+        .map(header => ({
+          visible: true,
+          ...header,
+          ...configured.find(p => p.value === header.value)
+        }))
```
matmen, line 113: I added some logs locally and noticed my configured header count was the same as the available header count. I played around with the merging code here and this seems to work reliable for me:
```suggestion
    const configuredHeaderValues = configured.map(header => header.value);
    const allHeaders = [...configured];
    for(const header of headers)
      if(!configuredHeaderValues.includes(header.value))
        allHeaders.push(header);

    return allHeaders.map(header => ({
      visible: true,
      ...headers.find(p => p.value === header.value), // apply original values
      ...header, // apply overrides
    }))
```

pedrolamas: If the header count is the same, how would this change make a difference? Something odd is going on here... can you check what you have on the database?

pedrolamas: My guess is that you had some weird error on the database, possibly from the development phase of the column picker code for Spoolman, and this just brought it up!

I've tried in both my printers and even with a brand new database and seems to work fine!

matmen: Sounds good to me then ;)

matmen (approved):

PR #1551, opened 2024-12-12
fix: Moonraker canbus uuid parameter

kluoyun:
* The can interface sent to the backend should be "can0", "can1", not "0" or "1".

Signed-off-by: XiaoK <xiaok@zxkxz.cn>

### DIFF
--- a/src/components/common/PeripheralsDialog.vue
+++ b/src/components/common/PeripheralsDialog.vue
@@ -368,7 +368,8 @@ export default class ManualProbeDialog extends Mixins(StateMixin) {
         type: 'can',
         count: this.canbusUuids
           ? Object.values(this.canbusUuids)
-            .reduce((a, b) => a + b.length, 0)
+            .filter(value => Array.isArray(value))
+            .reduce((a, b) => a + (b ? b.length : 0), 0)
           : undefined
       })
     }
@@ -414,9 +415,9 @@ export default class ManualProbeDialog extends Mixins(StateMixin) {
           break
 
         case 'can':
-          for (const canbusInterface in this.canbusInterfaces) {
+          this.canbusInterfaces.forEach((canbusInterface) => {
             SocketActions.machinePeripheralsCanbus(canbusInterface)
-          }
+          })
           break
       }
     }
--- a/src/components/common/PeripheralsDialog.vue
+++ b/src/components/common/PeripheralsDialog.vue
@@ -368,8 +368,7 @@ export default class ManualProbeDialog extends Mixins(StateMixin) {
         type: 'can',
         count: this.canbusUuids
           ? Object.values(this.canbusUuids)
-            .filter(value => Array.isArray(value))
-            .reduce((a, b) => a + (b ? b.length : 0), 0)
+            .reduce((a, b) => a + b.length, 0)
           : undefined
       })
     }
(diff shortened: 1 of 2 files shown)


src/components/common/PeripheralsDialog.vue
```diff
@@ -414,9 +415,9 @@ export default class ManualProbeDialog extends Mixins(StateMixin) {
           break
 
         case 'can':
-          for (const canbusInterface in this.canbusInterfaces) {
+          this.canbusInterfaces.forEach((canbusInterface) => {
             SocketActions.machinePeripheralsCanbus(canbusInterface)
-          }
+          })
```
pedrolamas, line 420: Plese revert this, I see no reason to use `.forEach` instead of a regular `for` cycle here.

src/components/common/PeripheralsDialog.vue
```diff
@@ -368,7 +368,8 @@ export default class ManualProbeDialog extends Mixins(StateMixin) {
         type: 'can',
         count: this.canbusUuids
           ? Object.values(this.canbusUuids)
-            .reduce((a, b) => a + b.length, 0)
+            .filter(value => Array.isArray(value))
```
pedrolamas, line 371: Not sure I understand the need for this check, according to Moonraker documentation, this is always an array (empty or not), never null or anything else!

src/components/common/PeripheralsDialog.vue
```diff
@@ -368,7 +368,8 @@ export default class ManualProbeDialog extends Mixins(StateMixin) {
...
+            .reduce((a, b) => a + (b ? b.length : 0), 0)
```
pedrolamas, line 372: If the filter for `Array.isArray()` is done, this should not be required as `b` will definitely be defined and an array.

pedrolamas (changes requested):
Hi @kluoyun thank you for submitting this PR.

You are absolutely correct on the `payload.can_uuids` in the mutations file, I completely missed that one!

However, I do not understand the need for the rest of the changes, as far as I can see in the Moonraker documentation, `can_uuids` is always an array, unless you have actually seen otherwise?

src/components/common/PeripheralsDialog.vue
kluoyun: > Plese revert this, I see no reason to use `.forEach` instead of a regular `for` cycle here.

The const canbusInterface obtained directly by using for in is a serial number, not a real network device name.
You will receive an error from moonraker
image


Or can we do this?
```
for (const canbusInterface in this.canbusInterfaces) {
    SocketActions.machinePeripheralsCanbus(this.canbusInterfaces[canbusInterface])
}
```

src/components/common/PeripheralsDialog.vue
kluoyun: I encountered a problem once in the test that `b` does not have the attribute `length`. However, it was not encountered in subsequent tests. This change may not be required.

src/components/common/PeripheralsDialog.vue
pedrolamas: Ohhh, another mistake of mine!

That should be `of`, not `in`... this is how it should have been:

```typescript
for (const canbusInterface of this.canbusInterfaces)
```

src/components/common/PeripheralsDialog.vue
kluoyun: I just tested it and it is indeed like this, so I will restore these changes.

pedrolamas (approved):

Issue #1562, opened 2024-12-31
Spoolman call to `SAVE_VARIABLE` needs to use lower case variable name


pedrolamas:
### Fluidd Version

1.31.3

### Browser

Edge

### Device

Desktop PC

### Operating System

Windows

### What happened

Spoolman integration in Fluidd is currently using `SAVE_VARIABLE`to persist the selected spool across restarts.

The issue is that Klipper variable names are always stored in lowercase (so variable "A" will be stored as "a"), but the in-memory dictionary is case-sensitive (so variables "A" and "a" will have distinct values).

Even worse, when storing variable "A", the value for variable "a" will be returned on restart, thus the other one is ignored.

### What did you expect to happen

Spoolman integration correctly persists the selected spool across restarts.

### How to reproduce

- Ensure you have `[save_variables]` Klipper module loaded and configured in printer.cfg
- Change spools and observe the results in the saved file

### Additional information

Reported by `fboc` in Fluidd Discord.

Public discussion on this issue is available here: https://klipper.discourse.group/t/save-variable-only-works-with-lower-case-variable-names/21032

PR #1563, opened 2024-12-31
fix: lower case variable name for SAVE_VARIABLE

pedrolamas:
Change the call to `SAVE_VARIABLE` in Spoolman integration to ensure we only use lower-case variable names.

Fixes: #1562


matmen (approved):
Weird, so far I haven't run into any issues with this, though I haven't been using multi tool support for the last couple months. Might be a recently introduced bug in Klipper then.. Either way, the fix LGTM

pedrolamas:
To be honest I was surprised by this too, but I managed to reproduce the problem quite easily, hence why I opened that discussion on Klipper Discourse to see how to proceed in handling this upstream.

Nevertheless, this fix seems to do the job on Fluidd side!

PR #1570, opened 2025-01-08
feat: control printer power via Moonraker

### LINKED ISSUE #1566: Power the printer (mainboard) off and on

If i want to change port of fans and other devices on my printers mainboard, i have to power down my whole printer, including my host and all connected stuff. Rebooting it is a time sink.

I changed my hardware, so that i can powerdown/powerup my mainboard via a shelly seperatly. Now i would like not to use the shelly cloud, but issue the off/on commands via the printerweb ui. I learned, that for this i can use moonraker.

I figuered out (hopefully correct) how to write the config in moonraker. But i found no Button for it in Fluidd. 
I found out, that mainsail already had such an implementation and would like to request such a feature also for fluidd. 

```ini
[power shelly]
type: http
bound_services: klipper
initial_state: on
restart_klipper_when_powered: True
restart_delay: 2
user: 
password: 
on_url:http://192.168.178.104/relay/0?turn=on
off_url:http://192.168.178.104/relay/0?turn=off
status_url:http://192.168.178.104/rpc/Switch.GetStatus?id=0
response_template:
```

https://github.com/mainsail-crew/mainsail/issues/897 --> the request in the mainsail git
and https://docs.mainsail.xyz/overview/quicktips/printer-power-switch as other alternative. I don´t know how mainsail did it finally.
--> add:
https://github.com/fluidd-core/fluidd/issues/1350 -> already a FR for a Powerbutton change (but a little bit different)

### Additional information

_No response_

pedrolamas:
Adds a new optional "Printer Power Device" setting that allows selection of a Moonraker power device controller that should then control the printer power.

## Settings

image

## Status when Klipper is not connected and device is off

image

Resolves #1566

### DIFF
--- a/src/mixins/state.ts
+++ b/src/mixins/state.ts
@@ -2,6 +2,7 @@ import Vue from 'vue'
 import { SocketActions } from '@/api/socketActions'
 import { Component } from 'vue-property-decorator'
 import type { Macro } from '@/store/macros/types'
+import type { Device } from '@/store/power/types'
 
 @Component
 export default class StateMixin extends Vue {
@@ -72,6 +73,18 @@ export default class StateMixin extends Vue {
    */
   get printerPrinting (): boolean {
     return this.printerState.toLowerCase() === 'printing'
+  }
+
+  get printerPoweredOff (): boolean {
+    if (this.klippyConnected) {
+      return false
+    }
+
+    const printerPowerDevice: string = this.$store.state.config.uiSettings.general.printerPowerDevice ?? 'printer'
+
+    const device = this.$store.getters['power/getDeviceByName'](printerPowerDevice) as Device | undefined
+
+    return device?.status === 'off'
   }
 
   /**
--- a/src/components/common/KlippyStatusCard.vue
+++ b/src/components/common/KlippyStatusCard.vue
@@ -18,7 +18,20 @@
         >
           <v-row>
             <v-col
-              v-if="klippyStateMessage !== 'Printer is ready'"
+              v-if="printerPoweredOff"
+              cols="12"
+            >
+              <v-alert
+                text
+                dense
+                type="error"
+                class="ma-0"
+              >
+                <span v-html="$t('app.general.error.printer_powered_off')" />
+              </v-alert>
+            </v-col>
+            <v-col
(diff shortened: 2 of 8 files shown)


src/mixins/state.ts
```diff
@@ -74,6 +75,22 @@ export default class StateMixin extends Vue {
     return this.printerState.toLowerCase() === 'printing'
   }
 
+  get printerPoweredOff (): boolean {
+    if (this.klippyConnected) {
+      return false
+    }
```
matmen, line 81: This doesn't seem to work for me - Moonraker always reports Klippy as connected (but shut down):
image

Changing this to `this.klippyReady` seems to work better, but will require the first two `<template>`s in SystemControl to be swapped (first the `<template v-if="printerPoweredOff">"`, then the `<template v-else-if="klippyConnected">`)

matmen (changes requested):

src/mixins/state.ts
pedrolamas: Maybe I am missing something, but if `klippy_connected: true`, then it means the Moonraker is in effect connected to Klipper, so the printer has power (this method is to check if printer isn't powered on, hence we return `false` as it IS powered on)

matmen: Yep, I would interpret it in the same way, but for whatever reason Moonraker does report Klippy to be connected (though the printer is definitely powered off)..

pedrolamas: I wonder if it is reporting the host klippy connection 🤔

pedrolamas: Did you add `bound_services: klipper` to the power element?

```ini
[power printer]
type: <any type>
...
bound_services: klipper
```

This MUST be set as it will stop the service when the power is off and turn the service back on when power is on (also matches [this](https://docs.mainsail.xyz/overview/quicktips/printer-power-switch))

matmen: ah, this must be why! i didn't have this set, i'll test your changes again when i get home ;)

pedrolamas: I will add a page to the docs on this topic!

matmen: Yep, adding the bound service fixed it!

PR #1574, opened 2025-01-14
feat: Add printer url in search params so that URLs become bookmarkable

### LINKED ISSUE #677: Bookmarks in browser for multiple printers

I'd like to have bookmarks in my browser that lead me directly to the right printer in the Fluidd visualisation. I'm running 2 printers on the same RPi4, each with their own Klipper and Moonraker instance. Currently, regardless of which printer you selected inside Fluidd, the url remains the same (http://fluiddpi.local). You can then switch printers inside Fluidd, but I'd like a direct method to call a specific printer. This is also useful when using Fluidd on mobile devices in the browser, as it's common there to use tiles or similar to link to web pages.

The ability to somehow have unique urls for each printer inside Fluidd. 

_No response_

### Additional information

_No response_

andrewboktor:
Add printer url in search params so that URLs become bookmarkable
Look at search param while loading and use that to find the right config

Resolves #677 

Signed-off-by: Andrew Boktor <andrew.boktor@gmail.com>


pedrolamas:
Hi @andrewboktor, first of all, a big THANK YOU for this PR!

I admit I have been working on this for a few months now and my solution was to use the `VueRouter` to do it, only I was hitting a lot of in-app navigation issues and so I just stopped...

Your approach to altering the main URL (the part before the '#' navigation) is the solution for that problem!

I will be taking this PR as base and merging a few changes (I want to try and ensure that page is not reloaded as it currently does, and am considering replacing the btoa with an MD5 hash instead), but it will definitely be based on the work you have provided here! 😀

andrewboktor:
> Hi @andrewboktor, first of all, a big THANK YOU for this PR!
> 
> I admit I have been working on this for a few months now and my solution was to use the `VueRouter` to do it, only I was hitting a lot of in-app navigation issues and so I just stopped...
> 
> Your approach to altering the main URL (the part before the '#' navigation) is the solution for that problem!
> 
> I will be taking this PR as base and merging a few changes (I want to try and ensure that page is not reloaded as it currently does, and am considering replacing the btoa with an MD5 hash instead), but it will definitely be based on the work you have provided here! 😀

Yea MD5 hash works too, or any other hash for that matter if we want to make the URLs smaller.
Another alternative I considered was adding an ID to ApiConfig when it's stored in localstorage, but apiUrl works just as well as an ID so I ditched that idea.

I do agree that likely the cleanest solution is to add a "printer" as part of the app paths in the router to all the routes. But didn't go that direction because this likely would require touching a lot of code that I'm not familiar with. 

Please let me know if I can yield any assistance, bounce ideas, or prototype some solution or other.
Happy to test if you have a branch as well.

andrewboktor:
I pulled the branch and ran a quick test and it seems to be working well!
May be an improvement here is to remove the printer from search params if it's not found in localStorage.

pedrolamas (approved):

PR #1576, opened 2025-01-18
feat(spoolman): allow selecting fields to show in dashboard card

### LINKED ISSUE #1529: Allow customization of spoolman panel
I don't, and I expect most spoolman users, use the `Location`, `Lot nr` and `Comment` fields provided by Spoolman but they still show up in the panel, always empty.

afbeelding

 I'm more interested in data from other fields like price, used length, density and temperature, or any of the other spoolman metadata.

Spool data:

afbeelding

Manufacturer data:

afbeelding

Would be nice if I can select my own spoolman fields, with a minimum and maximum nr of fields.

fluidd 1.30.6

matmen:
Adds a new setting to select which fields one wants to show in the Spoolman dashboard card:
image
image

Closes #1529


src/components/settings/SpoolmanSettings.vue
```diff
@@ -221,6 +238,38 @@ export default class SpoolmanSettings extends Mixins(StateMixin) {
...
+      'density',
+      'extruder_temp',
+      'bed_temp',
+      'first_used',
+      'last_used',
+      'comment'
+    ].map(field => ({ value: field, text: this.$t(`app.spoolman.label.${field}`) }))
+  }
+
+  get fieldsToShowInSpoolmanCard (): string[] {
```
pedrolamas, line 261: Is it worth using a custom type with all fields instead of `string[]`?

Something like:

```typescript
type SpoolmanKnownFields = 'id' | 'vendor' | 'filament_name' | 'remaining_weight' | 'used_weight' | 'location' | 'material' | 'lot_nr' | 'price' | 'density' | 'extruder_temp' | 'bed_temp' | 'first_used' | 'last_used' | 'comment'
```

matmen: IMO there's no need to add an extra type, the values are never referenced in code anyway (so no compiler warnings).. And I think in the future we could also add custom spoolman fields to the dashboard, which wouldn't be covered by the type

pedrolamas: They are referenced on the if's, but I am fine either way! 🙂

pedrolamas (approved):

PR #1579, opened 2025-01-26
fix(spoolman): QR code detection with MJPEG-adaptive camera sources

matmen:
Fixes a regression introduced by https://github.com/fluidd-core/fluidd/commit/552a4f625a5dc2c48ca82d2a62b2e637249e6611 where MJPEG-adaptive camera sources can't detect QR codes


pedrolamas:
I was under the impression this was actually not required anymore, hence why I removed it... is that not the case? 😕

matmen:
Right now the frame event is only emitted for the first frame, but not subsequent ones. Not sure if other image sources are affected by this too..

pedrolamas (approved):

PR #1638, opened 2025-03-11
feat(Spoolman): adds new metadata

pedrolamas:
## Default columns on the Spoolman dialog

image

## All columns on the Spoolman dialog

image


pedrolamas:
Well, I think I found a Spoolman bug while testing this... https://github.com/Donkie/Spoolman/issues/648

PR #1645, opened 2025-03-17
feat(Spoolman): spool progress indicator

pedrolamas:
Shows a circular progress indicator around the spool icon.

## Screenshot (progress values are random for demo purposes)

image


pedrolamas:
@matmen layout is never my thing, so if you have any thoughts on how better to show this, I'm all ears... 😁

matmen (approved):
I haven't had time to test this yet (though the code LGTM), but layout wise it looks fine to me :)  At first I was unsure if the blue part was showing how much filament has already been used or how much is left on the spool, but I guess it makes sense for it to show how much is left

moggieuk:
I'm not suggesting you change the base icon but users of the Happy Hare panel like the readability of a simple %.  Could you layer a rectangle with % in contrasting color?
Screenshot 2025-03-20 at 9 22 28 PM

The use of progress meter is cute but doesn't immediately jump out at me with purpose.  I do like having a simple indication of available filament though.

pedrolamas:
@moggieuk I did notice that on Happy Hare and do like it!

I ran a couple of experiments on this layout and didn't reach something I actually liked, so for now I will merge this as-is and revisit at a later time for improvements!

PR #1641, opened 2025-03-13
feat: Implementation of MMU panel to support single-extruder-multi-material printing using the Happy Hare for any MMU/AFC

moggieuk:
## Description

The PR adds an additional "MMU" Panel -- very visual and dynamic to both control and monitor state of MMU. This works with Happy Hare for any MMU or AFC (configuration is through klipper extensions). Prerequisite moonraker support has already been incorporated.

Summary:
- There is one new "MMU" panel and 4 dialogs launched from the main panel menu.
- If an MMU is not detected then the panel isn't displayed and the only overlap with existing functionality is in the print start dialog which also tests for existence of MMU.
- The additions are all tested on mobile, tablet and widescreen on dark and light modes.
- The only "connective tissue" is on the print start dialog where I detect if this is an MMU print and use my "tool mapper" instead of spoolman which will still appear for single color prints in MMU bypass is selected.
- This integrates with current spoolman functionality. I made one very small change to the spoolman active spool chooser to allow it to be used for selection purposes without setting active spool
- Finally, the print start functionality relies on some new metadata from moonraker -- that PR has already been merged into the latest moonraker.
- All strings are tokenized and in locales/en.yaml
- I created one additional mixin (mmu.ts) as a base for Mmu components.
- If an MMU is not detected then the panel isn't displayed and the only overlap with existing functionality is in the print start dialog which also tests for existence of MMU.

The attached screenshots give a flavor of the main panel and a couple of the dialogs.

## Mobile & Desktop Screenshots/Recordings

Here is a non-exhaustive selection of screenshots (some dialogs excluded):
Screenshot 2025-03-13 at 11 33 53 AM
Screenshot 2025-03-13 at 11 35 25 AM
Screenshot 2025-03-13 at 11 35 04 AM
Screenshot 2025-03-13 at 11 34 42 AM
Screenshot 2025-03-13 at 11 35 40 AM

Development and user testing has been performed on a public (Happy Hare) discord channel on: https://discord.gg/98TYYUf6f2

## Are there any post-deployment tasks we need to perform?

- The functionality will only be seen if popular Happy Hare klipper extension is installed and configured.  This populates `printer.mmu` and `printer.mmu_machine` printer variables.
- The latest moonraker is required for the start print mapping to function correct since new metadata has been added.

## Signed-off-by
Signed-off-by: Paul Morgan <moggieuk@gmail.com>

PR #1666, opened 2025-04-08
feat: Enhancements to Happy Hare MMU support

moggieuk:
## Description
The PR enhances the recent PR1641 in the following ways:
1. Added config option to display spools that have filament attributes but are unavailable to MMU (user request)
2. Adds logo for new MMX MMU (user request)
3.  Fixes small rendering error in the position of the filament in the extruder in one corner case condition
4. Renames confusing 'unit' variable to 'unit-index' to avoid potential conflict with mmu.ts mixin
5. Shows red "!" when the filament runs out according to spoolman but the MMU has not yet detected runout (user request)

## Are there any post-deployment tasks we need to perform?
- Same as PR1641 : https://github.com/fluidd-core/fluidd/pull/1641

## Signed-off-by
Signed-off-by: Paul Morgan <moggieuk@gmail.com>


moggieuk:
FYI: I know you merge commit log... I had some issues syncing back with latest upstream so more commits appear that the changes warrant.

pedrolamas:
> FYI: I know you merge commit log... I had some issues syncing back with latest upstream so more commits appear that the changes warrant - not sure what happened

Not a problem, we squash the PRs before merging!

src/components/widgets/mmu/MmuFilamentStatus.vue
```diff
@@ -658,6 +665,13 @@ export default class MmuFilamentStatus extends Mixins(StateMixin, MmuMixin) {
     return `${this.encoderPos}`
   }
 
+  get gateSensorName (): string {
+    if (this.unitDetails(this.unit).multiGear) {
+      return 'Hub (Gate)'
```
pedrolamas, line 670: Ideally (and unless I am missing something here), these should be localized strings!

moggieuk: Well, possibly, but those terms are used in config and throughout doc to reference a sensor name. I think it would be confusing to have translated sensor names...   The "Hub" was and add-on to help the Box Turtle and other type-B MMU's who all refer to the "gate" sensor as a "hub" sensor.

PR #1685, opened 2025-06-01
fix: undefined property access in case of deleted macro lookup

matmen:
When a macro is removed from the config, it isn't deleted from the stored config in the database. This can lead to an undefined property access while looking up the stored macro data:

image

### DIFF
--- a/src/store/macros/getters.ts
+++ b/src/store/macros/getters.ts
@@ -25,7 +25,7 @@ export const getters = {
         const lowerCaseKey = key.toLowerCase()
         const lowerCaseName = name.toLowerCase()
         const config = rootState.printer.printer.configfile.settings[lowerCaseKey]
-        const stored = state.stored.find(macro => macro.name.toLowerCase() === lowerCaseName)
+        const stored = state.stored.find(macro => macro.name?.toLowerCase() === lowerCaseName)
         const variables = rootState.printer.printer[key]
 
         const macro: Macro = {


pedrolamas (approved):
Hey @matmen, I haven't been able to reproduce this, but codewise I see no problem with this change!

From what I can see, we have macro settings in the store that have no name, is that it?

That should not be the case, but this will fix the issue, no question about it!

pedrolamas:
We do have similar code on the macro/mutations.ts, I do wonder if the same change should be done there?

matmen:
> Oddly, I can't even find any macro2.name on the code...

yep, me neither, i think that's just what the compilation changes it to tho. If i jump to the source code from the dev tools it marks that line, and I've tested the fix to be working ;)

> From what I can see, we have macro settings in the store that have no name, is that it?

yep, i guess it's not being populated if it's not in the config. Alternatively we could just remove it from the saved macros, but that would delete all the saved settings if a user just decides to comment out the macro from the config for testing 

> We do have similar code on the macro/mutations.ts, I do wonder if the same change should be done there?

I don't think so - the code in there never runs because macros that aren't in the config can't be mutated either (they don't show up in the UI)

pedrolamas:
The thing is that the stored settings are populated from whatever is currently on the existing macros... So I don't get how one could go without a name!

I get that this fixes the problem, I just don't see how deleting/commenting out a macro would cause this as we already have the macro setting stored - we just won't use it as he macro is gone now!

matmen:
> ... as we already have the macro setting stored - we just won't use it as he macro is gone now!

well, that's the thing, we _do_ use the stored macro settings as we iterate over each one in `state.stored.find()` ;) 

I can take a look at what's actually stored in my moonraker db tomorrow - perhaps the stored data was corrupted somehow and is just missing the `name` attribute. Either way, I think it wouldn't hurt to merge this if it's ok with you, just in case someone else runs into the same issue

pedrolamas:
Agreed, merge away, I see no problem with this change!

PR #1693, opened 2025-07-04
feat(WebrtcCamerastreamer): adds ping/pong support

pedrolamas:
Follow up on https://github.com/ayufan/camera-streamer/issues/55#issuecomment-3035240444

### DIFF
--- a/src/components/widgets/camera/services/WebrtcCamerastreamerCamera.vue
+++ b/src/components/widgets/camera/services/WebrtcCamerastreamerCamera.vue
@@ -31,7 +31,7 @@ export default class WebrtcCamerastreamerCamera extends Mixins(CameraMixin) {
   playbackAbortController: AbortController | null = null
   sleepAbortController: AbortController | null = null
 
-  // adapted from https://github.com/ayufan/camera-streamer/blob/4203f89df1596cc349b0260f26bf24a3c446a56b/html/webrtc.html
+  // adapted from https://github.com/ayufan/camera-streamer/blob/2d3a4884378f384346680a55196bf9244b99b6b6/html/webrtc.html
 
   async loadStream () {
     this.pc?.close()
@@ -58,7 +58,8 @@ export default class WebrtcCamerastreamerCamera extends Mixins(CameraMixin) {
                 'stun:stun.l.google.com:19302'
               ]
             }
-          ]
+          ],
+          keepAlive: true
         }),
         headers: {
           'Content-Type': 'application/json'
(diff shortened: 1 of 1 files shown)


src/components/widgets/camera/services/WebrtcCamerastreamerCamera.vue
```diff
@@ -83,6 +84,14 @@ export default class WebrtcCamerastreamerCamera extends Mixins(CameraMixin) {
 
       const pc = this.pc = new RTCPeerConnection(config)
 
+      pc.ondatachannel = (event: RTCDataChannelEvent) => {
+        const dc = event.channel
+
+        dc.onmessage = () => {
```
ayufan: Can you check that this dc is named `keepalive` and send only on this one?

ayufan (changes requested):

src/components/widgets/camera/services/WebrtcCamerastreamerCamera.vue
ayufan: It should be `dc.label == 'keepalive'`.

pedrolamas: Not sure I'm following, doesn't this match the exact changes you have in your reference code?

ayufan: @pedrolamas I will extend my changes to include `dc.label == 'keepalive'` check. If at somepoint there is a proper data channel for control messages it would not interfere with the clients.

src/components/widgets/camera/services/WebrtcCamerastreamerCamera.vue
```diff
@@ -83,6 +84,16 @@ export default class WebrtcCamerastreamerCamera extends Mixins(CameraMixin) {
...
 
+      pc.ondatachannel = (event: RTCDataChannelEvent) => {
+        const dc = event.channel
+
+        if (dc.label === 'keepalive') {
+          dc.onmessage = () => {
+            dc.send('pong')
+          }
+        }
+      }
```
pedrolamas, line 95: @ayufan I believe this is the check you mentioned? Can you confirm that is "keepalive" (all lower case)?

ayufan: Yes, this is correct. I pushed this commit to project: https://github.com/ayufan/camera-streamer/commit/400e036ea239d140a46762bbd852998eee8a92a8.

ayufan (approved):

PR #1710, opened 2025-09-02
chore: Optimized all Happy Hare MMU icons

pedrolamas:
Optimized all Happy Hare MMU svg icons with `npx svgo@latest --folder .`


pedrolamas:
@moggieuk can you pull these and see if they don't break in your Happy Hare please? 😁

PR #1818, opened 2026-04-11
fix(notifications): fix suffix icon display

matmen:
fixes support for suffix icons in the notification system. this has been bugging me for a while :p

before:
image

after:
image


pedrolamas (approved):
Wow, nice one, that is some seriously old bug!! 😁

PR #1852, opened 2026-05-24
feat: add native filaman widget

ManuelW77:
I implemented a Widget for my FilaMan (Filamentmanagement Tool)

You can contact me:
Manuel Weiser
manuelw@filaman.app

https://www.filaman.app


thijstriemstra:
FilaMan? Never heard of it? Maybe a link would be useful?

ManuelW77:
Ah, sorry i forgot :)

https://www.filaman.app

src/views/Settings.vue
```diff
@@ -84,8 +84,11 @@ export default class Settings extends Mixins(StateMixin) {
     return this.$typedGetters['server/componentSupport']('timelapse')
   }
 
-  get supportsSpoolman (): boolean {
-    return this.$typedGetters['server/componentSupport']('spoolman')
+  get supportsSpoolTracking (): boolean {
+    return (
+      this.$typedGetters['server/componentSupport']('spoolman') ||
+      this.$typedGetters['server/componentSupport']('filaman')
```
pedrolamas, line 90: Does this module actually exist? From what I can see in your code and your website, you are using the `[spoolman]` module and then have a Spoolman compatible API in Filaman, is that correct?

pedrolamas: Ah, found it! 😁
https://github.com/Arksine/moonraker/pull/1074

ManuelW77: Hey there, FilaMan is used by a lot of people and they wish a native implementation, because it can have more features than Spoolman. 
Yes, i build a spoolman api plugin to make the switch für the user easier. But, it can do all the things that FilaMan can do and I won't be bind FilaMan to a other system, because of complexity. 

So it would be nice, if I can implement FilaMan natively and get the development forward.

src/api/socketActions.ts
```diff
@@ -1094,6 +1119,25 @@ export const SocketActions = {
     )
   },
 
+  serverFilamanProxyGetAvailableSpools (page = 1, pageSize = 200, options?: NotifyOptions) {
+    const query = new URLSearchParams({
+      page: `${page}`,
+      page_size: `${pageSize}`
+    }).toString()
```
pedrolamas:
```suggestion
    const query = new URLSearchParams({
      page,
      page_size: pageSize
    }).toString()
```

src/store/spoolman/actions.ts
```diff
@@ -49,6 +49,135 @@ const createSpoolmanSocket = (spoolmanUrl: string): WebSocket | undefined => {
...
 
+const DEFAULT_DENSITY_G_CM3 = 1.24
+const DEFAULT_DIAMETER_MM = 1.75
+
+const isFilamanPaginatedResponse = (value: unknown): value is Moonraker.Spoolman.FilamanPaginatedResponse => {
+  return (
+    value != null &&
+    typeof value === 'object' &&
+    'items' in value &&
+    Array.isArray((value as Moonraker.Spoolman.FilamanPaginatedResponse).items)
```
pedrolamas:
```suggestion
    Array.isArray(value.items)
```

pedrolamas:
Hi @ManuelW77 thank you for sending this Pull Request.

I've done a quick review of these changes, and the bit I am not clear is on the changes from "Spoolman" to something generic - if the user configured with `[spoolman]` module in Moonraker and is just pointing to the Spoolman compatible API in Filaman, that should still be identified as Spoolman (just like it happens with Moonraker and the compatible Octoprint API exposed by it).

If the user is using a `[filaman]` module in Moonraker, then it should be identified as Filaman.

ManuelW77:
Hi @pedrolamas,

I’ve also implemented additional features. For example, Spoolman only supports a single toolhead in Fluidd, whereas Filaman supports multiple toolheads. I think this is especially important for newer printers like the Snapmaker U1 or Prusa models with multiple toolheads.

pedrolamas:
> Spoolman only supports a single toolhead in Fluidd, whereas Filaman supports multiple toolheads.

I assume that requires the `[filaman]` Moonraker module, correct?

ManuelW77:
Yes, that's right.

PR #1855, opened 2026-05-30
fix(afc): pull remaining weight from Spoolman when enabled

paxx12:
When Spoolman is configured, `spoolRemainingWeight` now reads
`spool.remaining_weight` from the Spoolman spool data instead of
`lane.weight` from the AFC lane state.

This keeps the displayed weight and spool fill percentage in sync
with Spoolman's tracking (which accounts for filament consumed
across all prints), rather than relying solely on the value AFC
reports for the lane.

Falls back to `lane.weight` when Spoolman is not enabled or the
spool has no `remaining_weight` recorded.

Signed-off-by: paxx12 <paxx12dev@gmail.com>

### DIFF
--- a/src/components/widgets/afc/AfcCardUnitLaneBody.vue
+++ b/src/components/widgets/afc/AfcCardUnitLaneBody.vue
@@ -148,6 +148,9 @@ export default class AfcCardUnitLaneBody extends Mixins(StateMixin, AfcMixin) {
   }
 
   get spoolRemainingWeight (): number {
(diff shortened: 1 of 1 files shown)


paxx12:
@jimmyjon711 Following our conversation about this change. Please review and approve.

jimmyjon711 (approved):
Tested with spoolman and change weight manually that AFC tracks, acts like it should and only pulls in spoolmans weight. Switch to not having spoolman and weight tracks AFC weight correctly.

pedrolamas (approved):

PR #1856, opened 2026-05-30
feat(afc): support multiple tools per lane via `map` array

paxx12:
Broadens `AfcLaneState.map` from `string | null` to `string | string[] | null` so the backend can assign multiple tools to a single lane. This is backwards-compatible — existing firmware that sends a plain string continues to work without changes.

This is to provide support for mapping logical tools (provided by slicer) to physical topology (actual filament loaded): https://github.com/AFCProject/AFC-Klipper-Add-On/issues/605

## Changes

- **`klipper.d.ts`** — type updated to `string | string[] | null`
- **`AfcCardUnitLaneHeader`** — arrays are joined with `, ` so the button reads `Lane1 > T0, T1`; empty/`null` shows `NONE` (consistent with the runout lane convention)
- **`AfcUnitLaneMappingToolDialog`** — tool roster built by flattening all lanes' `map` values; tools already in the current lane's `map` are disabled (grayed out) rather than hidden

## Screen recording

The behavior of mapping, whether the tool is remapped, or moved depends purely on how `AFC` executes the `SET_MAP`. This basically allows to visualise `map: string[]` values, and select from those values.

recording

> Note: The latest commit swapped `Lane > Tool` to `Tool > Lane`. It otherwise behaves exactly the same.

Signed-off-by: paxx12 <paxx12dev@gmail.com>

### DIFF
--- a/src/components/widgets/afc/AfcCardUnitLaneHeader.vue
+++ b/src/components/widgets/afc/AfcCardUnitLaneHeader.vue
@@ -38,7 +38,9 @@ export default class AfcCardUnitLaneHeader extends Mixins(StateMixin, AfcMixin) 
   }
 
   get mappedTool (): string {
-    return this.lane?.map ?? '--'
+    const map = this.lane?.map
+    if (map == null || map.length === 0) return 'NONE'
+    return Array.isArray(map) ? map.join(', ') : map
   }
 }
 </script>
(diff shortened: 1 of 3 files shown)


paxx12:
@jimmyjon711 Please validate this change.

jimmyjon711 (approved):
Tested with current AFC, looks good to me

https://github.com/user-attachments/assets/496beb59-1b2c-40eb-8ea6-9ce702ef3d8a

pedrolamas (approved):

Issue #1343, opened 2024-02-03
Renaming Fans & Outputs / Thermals items


yw4z:
### Is your feature request related to a problem? Please describe

Hi, first thanks for all efforts on fluid. its very handy
Im using a rooted K1 with Guilouz guide. and i tried to rename fan0 and other other pins with config file but i loose many features like controls not in sync with touch screen and also i started to get errors while printing. problem is fan0 etc is hardcoded to firmware and cannot edit them

### Describe the solution you'd like

simply giving custom names to outputs and thermal items will be easier to read
here is my current status
![Screenshot-20240203144600](https://github.com/fluidd-core/fluidd/assets/28517890/72d4b004-6e2d-47de-88d9-a3c4213b0177)

and here is edited version but i have many problems with this changes
![Screenshot-20240131000628](https://github.com/fluidd-core/fluidd/assets/28517890/93274af3-d626-4079-82fa-6a723ae2e51c)

Also "Round numbers in values" option will be nice to have

### Describe alternatives you've considered

Using an input box for item names can be solution

pedrolamas:
Hi @yw4z thank you for posting this Feature Request.

Can I ask that you upload your klipper.log please? It will include the full printer config and allow us to check it and test some of it on our side.

That will also allow us to pinpoint what control is having the rounding issue and fix that.

Regarding the renaming of the entities, we will probably wait for someone to contribute such a feature as I don't believe it is such a priority at the moment and our dev team is already quite small.

yw4z:
Hi
i added 2 logs one with custom names and other one is stock names for fans
if you cant find issue i can send older ones too
[klippy.zip](https://github.com/fluidd-core/fluidd/files/14152623/klippy.zip)

macros could be different on both files still experimenting on macros

pedrolamas:
@yw4z any specific reason why you are using an `[output_pin side_fan]` instead of `[generic_fan side_fan]` ?

yw4z:
actually its hardcoded to firmware as [output_pin fan2] and fan control from touch UI will not work if i use [output_pin side_fan] . so i started to use [output_pin fan2] again. thats why im asking for a renaming feature. I guess creality used fan2 in other parts in code and probably they used as [output_pin fan2]
thanks for your efforts

pedrolamas:
I am indeed seeing what you mean on the convertion script stock config files, however one can't say it is "hard coded" when it is all just easily changeable text files! 🙂

[...]

PR #1902, opened 2026-07-08
feat: display-name aliases for Klipper objects

philippop:
## What

Adds a user-editable, **display-only** alias map (`objectKey → friendly name`) for fans, `output_pin`s, LEDs, heaters and temperature sensors. Users can give an object a friendly name in the UI without renaming the underlying Klipper config section.

Motivation: on some machines (e.g. rooted Creality K1) fans are hardcoded in firmware as `output_pin fan0/1/2`; renaming the Klipper section breaks the printer's touchscreen fan control. This provides friendly names that **never touch Klipper config or emit G-code**.

## How

Mirrors the existing `sensorColors` feature end-to-end:

- **Store**: `aliases: Record<string,string>` on `uiSettings.dashboard`, with `setAlias`/`setRemoveAlias` mutations and `updateAlias`/`removeAlias` actions, persisted to the Moonraker DB `fluidd` namespace via the existing `dbKey` mechanism. Auto-loaded through the existing `mergeWith` in `setInitUiSettings` — no bespoke load wiring.
- **Getters**: `getOutputs`/`getHeaters`/`getSensors` resolve `aliases[key]` first and expose a new `defaultPrettyName` (the true pre-alias name) so the settings panel placeholder shows the real default even while an alias is active.
- **Settings panel**: new `AliasSettings.vue` to set/clear aliases (placeholder = `defaultPrettyName`).
- **Display-only guarantee**: the fan/pin G-code paths are keyed off the raw object `name`, never the alias (extracted to `src/util/output-gcode.ts`).
- **XSS**: user aliases reach two raw-HTML sinks — the ECharts thermal-tooltip formatter and the `v-safe-html` `OutputFan`/`OutputPin` labels — and are HTML-escaped at both via a single `escapeHtml` util.
- **i18n**: `src/locales/en.yaml` only (other locales left to Weblate).

## Testing

- `pnpm run lint`, `pnpm run type-check`, `pnpm run test` (375 pass), `pnpm run circular-check` (no cycles), and a production build all pass clean.
- 51 new unit tests: store mutations/actions incl. guarded delete + prototype-chain safety, getter alias-override + `defaultPrettyName`, escape/resolve-label utils, thermal-tooltip formatter XSS, output-label escape, a display-only G-code regression (asserts the command path uses `name`, not the alias), and `dbKey` dot-in-value parity.
- Manually deployed to a Creality K1 Max: the alias settings panel renders and — critically for the display-only guarantee — the printer's touchscreen and physical fan control remain fully functional with aliases set.

## Note for reviewers (`tsconfig.vitest.json`)

The new store/getter specs import store source that references the ambient global namespaces (`Moonraker`/`Klipper`/`TSHelpers` in `src/typings`) and the `$filters`/`$colorset` Vue augmentations declared in `src/plugins`. Because `tsconfig.vitest.json`'s `include` **overrides** (does not merge with) the extended `tsconfig.app` include, those globals were unresolved under `vue-tsc --build`. I widened the vitest `include` to mirror the app source so the globals resolve.

A narrower include isn't possible without dragging the whole component graph in (`src/plugins/filters.ts` value-imports `@/router`), and a project reference to the app project is blocked by its `noEmit`/non-composite config. Happy to take direction if you'd prefer a different structural fix here.

---

Signed-off-by: Philipp Oppolzer <philipp.oppolzer@gmail.com>

### DIFF
--- a/src/components/layout/AppSettingsNav.vue
+++ b/src/components/layout/AppSettingsNav.vue
@@ -45,6 +45,7 @@ export default class AppSettingsNav extends Vue {
       { name: this.$tc('app.setting.title.camera', 2), hash: '#camera', visible: true },
       { name: this.$t('app.setting.title.tool'), hash: '#toolhead', visible: true },
       { name: this.$t('app.setting.title.thermal_presets'), hash: '#presets', visible: true },
+      { name: this.$t('app.setting.title.aliases'), hash: '#aliases', visible: true },
       { name: this.$t('app.setting.title.gcode_preview'), hash: '#gcodePreview', visible: true },
       { name: this.$t('app.general.title.timelapse'), hash: '#timelapse', visible: this.supportsTimelapse },
       { name: this.$t('app.mmu.title.headline'), hash: '#mmu', visible: this.supportsMmu },
--- a/src/components/settings/AliasSettings.vue
+++ b/src/components/settings/AliasSettings.vue
@@ -1,0 +1,112 @@
+<template>
+  <div>
+    <v-subheader id="aliases">
+      {{ $t('app.setting.title.aliases') }}
+    </v-subheader>
+    <v-card
+      :elevation="5"
+      dense
+      class="mb-4"
+    >
+      <app-setting :title="$t('app.setting.label.aliases')">
+        <template #sub-title>
+          {{ $t('app.setting.tooltip.aliases') }}
+        </template>
+      </app-setting>
+
+      <template v-if="items.length">
+        <v-divider />
+
+        <template v-for="(item, index) in items">
+          <app-setting
+            :key="item.key"
+            :title="item.defaultPrettyName"
+            :sub-title="item.key"
+          >
+            <v-text-field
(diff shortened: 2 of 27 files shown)


pedrolamas:
Hi @philippop thank you for sending this PR.

From what I can see, this is related and would resolve #1343.

However, and as I have previously stated on the above...

> my personal view from this is that Creality has once again taken a different approach as to how Klipper is supposed to be used...
> 
> Personally, I have zero inclination to support Creality work as they have proven multiple times that they have no problems with taking open-source code (eg. Klipper, Moonraker, Fluidd, Mainsail, KlipperScreen), changing it, and then using it on their printers without releasing said changes (thus breaking the open-source licenses).

philippop:
Hi @pedrolamas,
Thanks for the response.
I get the stance. I just got really annoyed having to remember (or re-test), after not using the printer for a while, which fan is which and solved it like this. Since I know that others are annoyed by it too, I thought I share it.

PR #1904, opened 2026-07-22
feat: optional trailing-average smoothing for power/speed chart series

### LINKED ISSUE #1896: feat: optional trailing-average smoothing for power/speed chart series

Fluidd samples the thermal chart at ~1 Hz. Heater-power and fan-speed are PWM duty cycles, so at that rate they render as a full-scale square wave — snapping between 0% and 100% almost every sample. The instantaneous values are accurate, but they read as visual noise: it's hard to see the *trend* the controller is actually holding (e.g. "the hotend is sitting around 35–40% power"). Temperatures don't have this problem — they move slowly and their raw shape is meaningful.

An **opt-in** General setting that applies a trailing moving-average to the **power/speed series only** — temperatures stay raw, so real thermal lag is never hidden. Default is **None**, so there's no behavior change unless a user turns it on. Window presets: None / 1 / 3 / 5 / 10 / 15 / 30 s.

I already have a small, working reference implementation on a fork — a pure, unit-tested util injected between the chart store and the ECharts dataset, plus the dropdown in General settings. It's been running on my own printer (Creality K1) and follows the repo's conventions (conventional commits, sign-off, lint/type-check/tests green). **Happy to open a PR if this is something you'd want in Fluidd** — and I'd rather hear your take on the design first (e.g. global setting vs. per-chart, which presets) than presume.

- *Leave it raw (status quo):* accurate, but the trend is hard to read on PWM series.
- *Smooth everything including temps:* rejected — it would hide meaningful thermal lag and could mask sensor issues.
- *Per-chart toggle instead of a global setting:* viable; I leaned toward one General setting for simplicity, but open to your preference.

### Additional information

I searched existing issues/PRs for smoothing / moving-average and didn't find a prior request — apologies if I missed one.

**Screenshot:** the same power trace with smoothing off (top, *none*) vs. 3 s (bottom) — _added below_.

Image

dvystrcil:
Closes #1896

## Summary

Adds an **opt-in** General setting that applies a trailing moving-average to the thermal chart's power/speed series only. Temperatures are left raw — real thermal lag stays visible.

- Default is `None` — no behavior change unless a user turns it on.
- Window presets: `None / 1 / 3 / 5 / 10 / 15 / 30 s`.
- Smoothing is a pure, unit-tested util (`smoothChartData`) applied between the chart store and the ECharts dataset; the window is stored in `uiSettings.general.chartSmoothingWindow` (seconds, `0` = none) and changes re-render the chart live.

I posted this as a feature request three weeks ago (#1896) and offered to open a PR if there was interest, but haven't heard back — I understand maintainer time is limited, so I'm going ahead and opening this as a concrete, working reference per the offer in that issue. Very happy to adjust the design (global setting vs. per-chart, different presets, or close this out entirely) based on your take — no pressure either way.

## Test plan

- [x] `pnpm run lint` — clean
- [x] `pnpm run type-check` — clean
- [x] `pnpm run test` — 330/330 passing (includes new `chart-smoothing.spec.ts` covering the util at every window size)
- [x] `pnpm run circular-check` — clean
- [x] `pnpm run build` — production build succeeds
- [x] Manually verified on my own printer (Creality K1) — screenshot in #1896 shows the same power trace with smoothing off vs. 3s

Signed-off-by: Daniel Vystrcil <dvystrcil@gmail.com>

### DIFF
--- a/src/components/settings/GeneralSettings.vue
+++ b/src/components/settings/GeneralSettings.vue
@@ -66,19 +66,6 @@
 
       <v-divider />
 
-      <app-setting :title="$t('app.setting.label.chart_smoothing')">
-        <v-select
-          v-model="chartSmoothingWindow"
-          filled
-          dense
-          single-line
-          hide-details="auto"
-          :items="availableChartSmoothingWindows"
-        />
-      </app-setting>
-
-      <v-divider />
-
       <app-setting
         :title="$t('app.setting.label.keyboard_shortcuts')"
         :sub-title="$t('app.setting.tooltip.keyboard_shortcuts')"
@@ -239,6 +226,19 @@
 
       <v-divider />
 
+      <app-setting :title="$t('app.setting.label.thermal_chart_smoothing')">
+        <v-select
+          v-model="chartSmoothingWindow"
+          filled
+          dense
+          single-line
+          hide-details="auto"
+          :items="availableChartSmoothingWindows"
+        />
+      </app-setting>
+
+      <v-divider />
+
       <app-setting
         :title="$t('app.setting.label.enable_diagnostics')"
         :sub-title="$t('app.setting.tooltip.diagnostics_performance')"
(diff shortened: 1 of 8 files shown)


dvystrcil:
@pedrolamas that toolhead resetting is interesting to know. Is there a place I could have understood that better?

pedrolamas:
> that toolhead resetting is interesting to know. Is there a place I could have understood that better?

Not something documented anywhere, I think... though probably should be!

Truth be told, that's a behavior we noticed some years ago, unfortunately there isn't much we can do there without breaking everyone's settings.

pedrolamas (approved):

PR #1907, opened 2026-07-26
feat(AFC): Adding toolchanger support into AFC panel

jimmyjon711:
- Adding toolchanger support to AFC panel with dedicated buttons to select/swap and dock toolheads
- Updated locales so that updated AFC toolchanger code translated correctly in panel
- Cleaned up what panel displays for standalone tools( toolhead without a filament changer attached to it ) 
- Added default `initial_weight` because spool percentage show wrong for users that don't use spoolman, this is a visable change for all users using Spoolman not just toolchangers. (Latest AFC release adds initial weight, but still updating for those that are still using older AFC software) 
- For transparency sake, new extruder panel buttons were added with help from claude, the rest of the changes were done by me.

Video of this in action:

https://github.com/user-attachments/assets/929e0d57-2adb-45e0-ad6b-cff57ba836d8

Signed-off-by: Jim Madill <jcmadill1@gmail.com>


src/components/widgets/afc/AfcCardExtruder.vue
```diff
@@ -170,16 +230,31 @@ export default class AfcCardExtruder extends Mixins(StateMixin, AfcMixin) {
   get bufferOutput (): string {
     const extruder = this.afcCurrentLane?.extruder
 
+    // "Standalone" lanes don't have buffers, so don't show any buffer status
+    if (this.afcExtruder?.is_standalone) {
+      return ' '
+    }
```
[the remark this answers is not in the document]
jimmyjon711: @pedrolamas Either a space or empty string works, just needed to not return `Buffer not Active` text. I made sure the change worked like I expected before committing it.

src/components/widgets/afc/AfcCardExtruder.vue
```diff
@@ -196,10 +271,6 @@ export default class AfcCardExtruder extends Mixins(StateMixin, AfcMixin) {
       return this.afcExtruder.lane_loaded
     }
 
-    if (this.afcCurrentLane) {
-      return this.afcCurrentLane.name
-    }
-
     return this.$t('app.afc.LaneLoadedNone').toString()
```
pedrolamas, line 254: **Removing the `afcCurrentLane` fallback here regresses the single-extruder case.**

The old fallback was genuinely wrong for multi-extruder setups (it could show a *different* extruder's lane), but deleting it outright means that during a load — while `afcExtruder.lane_loaded` is still empty — the card now reads `Loading: None` instead of `Loading: lane1`. That's the common non-toolchanger path.

Suggest scoping it rather than dropping it, using the existing `hasActiveLane` getter which already checks the current lane belongs to *this* extruder:

```ts
get stateLane (): string {
  if (this.afcExtruder?.lane_loaded) {
    return this.afcExtruder.lane_loaded
  }

  if (this.hasActiveLane && this.afcCurrentLane) {
    return this.afcCurrentLane.name
  }

  return this.$t('app.afc.LaneLoadedNone').toString()
}
```

src/components/widgets/afc/AfcCardExtruder.vue
```diff
@@ -230,4 +309,13 @@ export default class AfcCardExtruder extends Mixins(StateMixin, AfcMixin) {
 .v-application .border-error {
   border-color: var(--v-error-base) !important;
 }
+
+.v-application .border-warning {
```
pedrolamas, line 302: **`border-warning` silently wins over `border-error`.**

All three rules have identical specificity (`.v-application .border-*`), so source order decides the winner. Since `border-warning` is appended last, an extruder that is both `next_pickup` and in an error state renders a warning border and the error state becomes invisible.

Two options:

- move the `.border-warning` rule above `.border-error` in the stylesheet, or (preferably)
- make the class map mutually exclusive in `containerClasses` so precedence is explicit in TS rather than implied by CSS ordering:

```ts
get containerClasses () {
  const isError = this.hasActiveLane && this.afcErrorState

  return {
    'border-primary': !isError && (this.hasActiveLane || this.onShuttle),
    'border-warning': !isError && this.afcExtruder?.next_pickup === true,
    'border-error': isError,
    ...
  }
}
```

(Also note line 151 uses `this.afcExtruder?.on_shuttle` directly instead of the `onShuttle` getter you added just above it.)

src/mixins/afc.ts
```diff
@@ -44,6 +47,12 @@ export default class AfcMixin extends Vue {
     return this.afc?.error_state === true
   }
 
+  get afcHasToolchanger (): boolean {
+    const printerState = this.$typedState.printer.printer
+
+    return Object.keys(printerState).some(key => key.startsWith('AFC_Toolchanger'))
```
pedrolamas: **Prefix match needs a trailing space, and the casing needs confirming.**

Every other AFC accessor in this mixin keys off an exact prefix *with a trailing space* — `AFC_stepper `, `AFC_extruder `, `AFC_hub `, `AFC_buffer `. Without it this also matches something like `AFC_ToolchangerFoo`:

```ts
return Object.keys(printerState).some(key => key.startsWith('AFC_Toolchanger '))
```

…assuming AFC registers it as a *named* section. If it's a singleton object, an exact key lookup (`printerState.AFC_Toolchanger != null`) would be cheaper and unambiguous.

Can you confirm the exact object name/casing against the AFC source? `Toolchanger` vs `toolchanger` is the difference between the buttons appearing and silently never appearing, and there's no test covering it. A short comment citing the AFC object name would help future readers.

src/components/widgets/afc/AfcCardExtruder.vue
```diff
@@ -170,16 +230,31 @@ export default class AfcCardExtruder extends Mixins(StateMixin, AfcMixin) {
...
+    const status = this.afcExtruder?.status
+    return !this.klippyReady || this.printerPrinting ||
+      ((status != null && status !== 'Idle') || this.afcCurrentState !== 'Idle')
+  }
+
   get state (): string {
     const extruder = this.afcCurrentLane?.extruder
 
+    if (this.afcExtruder?.status && this.afcExtruder.status !== 'Idle') {
+      return this.$t(`app.afc.${this.afcExtruder.status}`).toString()
```
pedrolamas, line 231: **Missing locale key for `Initialized`.**

This widens the `$t('app.afc.<status>')` path to the whole `AfcStateState` union. `ToolSwap`/`ToolPickup`/`ToolDock` are covered by the new `en.yaml` keys ✅, but `Initialized` has no entry — it will render the raw `app.afc.Initialized` path. (Pre-existing on the `afcCurrentState` branch below, but this change makes it more reachable.) Please add the key.

Separately: typing the new extruder field as `status?: AfcStateState` conflates two state machines — an extruder will never be `Restoring`/`Ejecting`/`Moving`. A narrower `AfcExtruderStatus` union in `klipper.d.ts` would make the required locale key set provable instead of best-effort.

Minor, on `toolSelectionDisabled` (line 245): the parens are redundant and the null handling can collapse to

```ts
return !this.klippyReady || this.printerPrinting ||
  (this.afcExtruder?.status ?? 'Idle') !== 'Idle' ||
  this.afcCurrentState !== 'Idle'
```

Worth double-checking AFC never rests at `Initialized` after boot — if it does, the buttons stay permanently disabled until the first action.

src/mixins/afc.ts
```diff
@@ -164,7 +173,7 @@ export default class AfcMixin extends Vue {
 
     const material = spool?.filament?.material || laneObj?.material || ''
     const remainingWeight = spool?.remaining_weight ?? laneObj?.weight
-    const fullWeight = spool?.initial_weight ?? laneObj?.initial_weight
+    const fullWeight = spool?.initial_weight ?? laneObj?.initial_weight ?? AFC_DEFAULT_SPOOL_WEIGHT
```
pedrolamas, line 177: **This can now push `spoolPercent` above 100%, and the reel doesn't clamp.**

Previously a missing `fullWeight` short-circuited to `spoolPercent = 100`. With the 1000 g default, a lane reporting `weight: 2000` and no `initial_weight` yields `200`, and `AfcFilamentReel` uses `percent` unclamped:

```ts
const scaleY = minScale + (this.percent / 100) * (1 - minScale)
```

…so the fill overflows the SVG. Please clamp here:

```ts
spoolPercent = Math.min(100, Math.round((remainingWeight / fullWeight) * 100))
```

Also worth calling out in the PR description: this is a visible change for *every* non-Spoolman AFC user, not just toolchanger users. Anyone running 500 g or 750 g spools without `initial_weight` will now see a partially-empty reel where it previously read full. Matching AFC's own 1000 g default is defensible, but it shouldn't land unannounced.

(For what it's worth, the blast radius is limited — `fullWeight` is only surfaced via `spoolFullWeight` in `AfcCardUnitLaneBody.vue`, which is currently unused, so no fabricated "x / 1000 g" text leaks into the UI.)

pedrolamas (commented):
Thanks for this — the implementation follows the existing AFC widget patterns nicely (tooltip/icon buttons, `encodeGcodeParamValue` on the `TOOL=` param, `sendGcode` from `StateMixin`), and the toolchanger UX looks great in the video.

A few things I'd like addressed before merge — inline comments below cover:

1. `stateLane` fallback removal regresses single-extruder setups
2. `border-warning` silently outranks `border-error`
3. `AFC_Toolchanger` prefix match needs a trailing space + casing confirmation
4. `Initialized` has no locale key on the widened `$t` path
5. `AFC_DEFAULT_SPOOL_WEIGHT` can push `spoolPercent` over 100%

Smaller nits (non-blocking): `containerClasses` uses `this.afcExtruder?.on_shuttle` instead of the new `onShuttle` getter; the two `v-tooltip`/`v-btn` branches differ only by icon/label/handler and could collapse into one; `ToolSwap`/`ToolPickup`/`ToolDock` break `en.yaml`'s alphabetical ordering (should be `ToolDock`, `ToolPickup`, `ToolSwap`); and the PR title is 52 chars, just over our 50-char commit subject limit if this gets squashed.

src/components/widgets/afc/AfcCardExtruder.vue
jimmyjon711: Fixed and found an error during testing where `this.afcCurrentState !== 'Error'` needed to be added to `get toolSelectionDisabled` when AFC error state was set since the select/unselect radials did not show up when AFC was in an error state.

src/components/widgets/afc/AfcCardExtruder.vue
jimmyjon711: Fixed

src/mixins/afc.ts
jimmyjon711: Fixed

src/components/widgets/afc/AfcCardExtruder.vue
jimmyjon711: Fixed

src/mixins/afc.ts
jimmyjon711: Fixed, PR description edited for spoolman

src/typings/klipper.d.ts
```diff
@@ -770,6 +770,7 @@ declare namespace Klipper {
     }
```
pedrolamas, line 770: Minor naming collision: this new type is `AFCExtruderState` (all-caps acronym), while the pre-existing interface two lines below is `AfcExtruderState` (Pascal-case acronym) — same characters modulo case, describing related-but-different things (extruder *status* vs the extruder *object*). Every other type/interface in this file uses the `Afc` (Pascal-case) prefix consistently (`AfcStateState`, `AfcBufferStatus`, `AfcLaneSettings`, …).

Suggest renaming to something like `AfcExtruderStatus` — matches the file's convention and avoids two identifiers that are one keystroke apart:

```ts
export type AfcExtruderStatus = 'Idle' | 'ToolSwap' | 'ToolDock' | 'ToolPickup' | 'Error'

export interface AfcExtruderState {
  ...
  status?: AfcExtruderStatus;
}
```

pedrolamas (commented):
Re-reviewed after the updates — all 5 blocking items from the previous pass are resolved:

1. ✅ `stateLane` now gated on `hasActiveLane && this.afcCurrentLane` — single-extruder regression fixed.
2. ✅ `containerClasses` now computes `isError` once and makes `border-primary`/`border-warning`/`border-error` mutually exclusive — no more source-order precedence issue.
3. ✅ `afcHasToolchanger` matches `'AFC_Toolchanger '` (trailing space) with a comment explaining the named-section requirement.
4. ✅ `Initialized` locale key added; `status` is now typed with the narrower `AFCExtruderState` union instead of reusing `AfcStateState`.
5. ✅ `spoolPercent` is clamped with `Math.min(100, ...)`, and the PR description now calls out the visible change for non-Spoolman users.

Good catch in your own testing on the `afcCurrentState !== 'Error'` addition to `toolSelectionDisabled` — makes sense that the buttons need to stay usable to recover once AFC is already in a global error state, while the per-extruder `status === 'Error'` check still blocks that specific extruder.

One new (small) thing from the narrowing in #4 — see inline comment on `klipper.d.ts`.

Non-blocking, still open from last round, up to you whether to bother:
- ~`en.yaml`: `ToolSwap`/`ToolPickup`/`ToolDock` are still out of the file's alphabetical order (would be `ToolDock`, `ToolPickup`, `ToolSwap`).~
- The two `v-tooltip`/`v-btn` blocks for select/unselect still differ only by icon/label/handler and could collapse into one parameterized block.

Neither blocks merging as far as I'm concerned.

src/typings/klipper.d.ts
jimmyjon711: Fixed, and also fixed another regression that I found for single-extruder toolheads when a lane was not loaded into toolhead.

pedrolamas (approved):

PR #1910, opened 2026-07-28
feat(AFC): Adding ability to manually enter spool data when spoolman is enabled

jimmyjon711:
Main changes:
- Added "Manual Entry" button to spoolman panel when users have AFC installed. This allows user to manually enter their spool data, since sometimes we just want to use a spool without always entering it into spoolman.
- Added ability to enter extruder and bed temperature(if bed temp variable exists) in manual entry panel
- Added ability to clear spoolman ID and still update values. If user does not check this box, values are still sent and updated locally to AFC.
- To add my normal disclaimer, changes were done with help by Claude but verified my me.

Update in action:

https://github.com/user-attachments/assets/c4d75900-9b84-41f6-8b27-9c83df694c87

Signed-off-by: Jim Madill jcmadill1@gmail.com


src/components/widgets/afc/dialogs/AfcUnitLaneFilamentDialog.vue
```diff
@@ -37,13 +37,54 @@
...
+          hide-details
+        />
+      </app-setting>
+      <v-divider class="my-3" />
+      <app-setting
+        :title="$t('app.afc.BedTemp')"
+        :sub-title="$t('app.afc.BedTempSubtitle')"
+      >
+        <v-text-field
+          v-model="bedTemp"
```
pedrolamas: `v-model` on a `v-text-field` always yields a **string**, and clearing the field yields `''` — `type="number"` doesn't change that. Combined with the missing guard in `disableSetBtn`, an emptied field emits:

```text
AFC_SET_SPOOL_TEMP LANE=lane1 BED_TEMP= EXTRUDER_TEMP=
```

which Klipper will reject.

Please use `v-model.number` on both fields:

```diff
-          v-model="extruderTemp"
+          v-model.number="extruderTemp"
```

Note also that `:min="0"` is not validation in Vuetify — a user can still type `-5` and it will be sent.

src/components/widgets/afc/dialogs/AfcUnitLaneFilamentDialog.vue
```diff
@@ -83,6 +131,14 @@ export default class AfcUnitLaneFilamentDialog extends Mixins(StateMixin, AfcMix
...
 
+  get currentExtruderTemp (): number {
+    return Math.round(this.lane?.extruder_temp ?? 0)
+  }
+
+  get currentBedTemp (): number {
+    return Math.round(this.lane?.bed_temp ?? 0)
+  }
+
   get disableSetBtn (): boolean {
```
pedrolamas, line 161: `material` and `weight` are guarded here but the two new temperature fields aren't, so an empty temp field can still be saved (see the comment on the inputs above):

```diff
   get disableSetBtn (): boolean {
     return (
       !this.material ||
       !this.weight ||
-      !this.color
+      !this.color ||
+      !Number.isFinite(Number(this.extruderTemp)) ||
+      !Number.isFinite(Number(this.bedTemp))
     )
   }
```

src/components/widgets/afc/dialogs/AfcUnitLaneFilamentDialog.vue
```diff
@@ -99,19 +155,31 @@ export default class AfcUnitLaneFilamentDialog extends Mixins(StateMixin, AfcMix
...
+    if (this.clearSpoolId || this.weight !== this.currentWeight) {
       gcode.push(`SET_WEIGHT LANE=${encodeGcodeParamValue(this.name)} WEIGHT=${this.weight}`)
     }
 
+    if (
+      this.clearSpoolId ||
+      this.extruderTemp !== this.currentExtruderTemp ||
+      this.bedTemp !== this.currentBedTemp
+    ) {
+      gcode.push(`AFC_SET_SPOOL_TEMP LANE=${encodeGcodeParamValue(this.name)} BED_TEMP=${this.bedTemp} EXTRUDER_TEMP=${this.extruderTemp}`)
```
pedrolamas: Follow-on from the `v-model` issue: once the user touches either field the model is a string, so this compares `'220' !== 220` and is always true. `AFC_SET_SPOOL_TEMP` then gets sent (and logged to the console) even when nothing changed. `v-model.number` on the inputs fixes this too.

Separately — worth confirming AFC applies this batch in order. The correctness of re-sending color/material/weight/temp when `clearSpoolId` is set depends on the `SET_SPOOL_ID LANE=… SPOOL_ID=` above having completed its lane reset first.

src/typings/klipper.d.ts
```diff
@@ -819,6 +819,7 @@ declare namespace Klipper {
     weight: number;
     filament_name?: string;
     extruder_temp: number | null;
+    bed_temp?: number | null;
```
pedrolamas, line 827: Optional here makes sense — older AFC won't report it. But that has a UI consequence worth handling: if `bed_temp` is missing on older AFC, `AFC_SET_SPOOL_TEMP` almost certainly is too. On those versions the dialog renders "Bed Temp: 0" (`currentBedTemp` falls back to `0`), the user edits it, and the save emits a command Klipper rejects with `Unknown command` — after the rest of the batch has already been applied.

Either gate the two temperature `app-setting` blocks on `lane?.bed_temp !== undefined` so older setups don't see them, or confirm the minimum supported AFC version already ships both and note it in the PR description. Right now the UI advertises a capability the backend may not have.

src/store/spoolman/types.ts
```diff
@@ -37,6 +37,8 @@ export interface SpoolSelectionDialogState {
   targetMacro?: string;
   spoolSelectionOnly?: boolean;
   selectedSpoolId?: number;
+  afcLaneName?: string;
```
pedrolamas: `afcLaneName` is never read for its value — `handleSwitchToManualEntry()` ignores it, and `AfcCardUnitLaneBody` reopens its own dialog using `this.name`. It's only tested for truthiness in a `v-if`.

Either rename it to what it actually does (`allowManualEntry?: boolean`) or genuinely route the lane name through. Carrying a string that's only used as a flag invites someone to later trust its value.

src/components/widgets/spoolman/SpoolSelectionDialog.vue
```diff
@@ -565,6 +575,17 @@ export default class SpoolSelectionDialog extends Mixins(StateMixin, BrowserMixi
...
   }
 
+  get afcLaneName (): string | undefined {
+    return this.$typedState.spoolman.dialog.afcLaneName
+  }
+
+  handleSwitchToManualEntry () {
+    this.$typedCommit('spoolman/setDialogState', {
+      show: false,
+      switchToManualEntry: true
```
pedrolamas: `switchToManualEntry` is never cleared after it's consumed — `state.spoolman.dialog` keeps `{ show: false, switchToManualEntry: true }` indefinitely.

It's safe today only because every other `setDialogState` call replaces the whole object. The `open` setter (~line 379) does **not** — it spreads the previous state. Nothing currently assigns `open = true`, so there's no live bug, but it's one line away from one.

Suggest resetting the flag in `AfcCardUnitLaneBody.onSpoolmanChanged` once it's been acted on.

src/components/widgets/spoolman/SpoolSelectionDialog.vue
```diff
@@ -565,6 +575,17 @@ export default class SpoolSelectionDialog extends Mixins(StateMixin, BrowserMixi
     return this.$typedState.spoolman.dialog.spoolSelectionOnly ?? false
   }
```
pedrolamas, line 577: Nit: `handleSwitchToManualEntry()` is inserted between two getters, breaking the getters-then-methods grouping used through the rest of this file. Please move it down with the other methods.

pedrolamas (commented):
Thanks for this! Nice feature, and it slots into the existing dialog-handoff pattern well. A few things to address before merge — the first one is a functional bug, the rest are hardening/cleanup.

src/components/widgets/afc/dialogs/AfcUnitLaneFilamentDialog.vue
jimmyjon711: Fixed, also added `number` to the weight field since it was also missing.

src/components/widgets/afc/dialogs/AfcUnitLaneFilamentDialog.vue
jimmyjon711: Fixed

src/components/widgets/afc/dialogs/AfcUnitLaneFilamentDialog.vue
jimmyjon711: Fixed

src/typings/klipper.d.ts
jimmyjon711: Fixed and tested to verify by removing `bed_temp` from AFC side. Also added a qualifier to only display extruder/bed temp if `AFC_SET_SPOOL_TEMP` macro is found since this is a new macro with the v1.2.0 version of AFC.

src/store/spoolman/types.ts
jimmyjon711: Fixed, changed to `allowManualEntry`

src/components/widgets/spoolman/SpoolSelectionDialog.vue
jimmyjon711: Fixed, also set `switchToManualEntry` to false once consumed

src/components/widgets/spoolman/SpoolSelectionDialog.vue
jimmyjon711: Moved down to where like code blocks are at.

src/components/widgets/afc/AfcCardUnitLaneBody.vue
```diff
@@ -236,6 +237,13 @@ export default class AfcCardUnitLaneBody extends Mixins(StateMixin, AfcMixin) {
     ) {
       this.spoolmanSelection = false
 
+      if (dialog.switchToManualEntry) {
+        dialog.allowManualEntry = false
+        dialog.switchToManualEntry = false
```
pedrolamas: `dialog` here *is* the Vuex state object, so these two assignments mutate store state outside a mutation handler. Strict mode is enabled in dev (`strict: import.meta.env.DEV` in the store), so this throws `Do not mutate vuex store state outside mutation handlers` on every manual-entry switch in a dev build.

They're also no-ops: `setDialogState` replaces the dialog object wholesale, and `handleSwitchToManualEntry` commits a fresh `{ show: false, switchToManualEntry: true }` — `allowManualEntry` is already gone by the time this runs, and the next `handleSelectSpool()` commits a brand new object anyway.

Suggest just dropping both lines:
```suggestion
      if (dialog.switchToManualEntry) {
```

src/components/widgets/afc/dialogs/AfcUnitLaneFilamentDialog.vue
```diff
@@ -83,11 +144,23 @@ export default class AfcUnitLaneFilamentDialog extends Mixins(StateMixin, AfcMix
...
+  }
+
   get disableSetBtn (): boolean {
     return (
       !this.material ||
       !this.weight ||
-      !this.color
+      !this.color ||
+      !Number.isFinite(Number(this.extruderTemp)) ||
+      (this.bedTemp !== undefined && !Number.isFinite(Number(this.bedTemp)))
```
pedrolamas: The `Number()` wrapper defeats the check. With `v-model.number`, clearing the field leaves `''`, and `Number('') === 0` → finite → the Set button stays enabled → `AFC_SET_SPOOL_TEMP LANE=lane1 EXTRUDER_TEMP=` goes out with an empty value, which Klipper rejects.

`Number.isFinite` doesn't coerce, so dropping the wrapper gives exactly the intended behaviour (`Number.isFinite('')` is `false`):
```suggestion
      !Number.isFinite(this.extruderTemp) ||
      (this.bedTemp !== undefined && !Number.isFinite(this.bedTemp))
```
Worth considering a `< 0` rejection too — `:min="0"` on `type="number"` isn't enforced against typed input.

src/components/widgets/afc/dialogs/AfcUnitLaneFilamentDialog.vue
```diff
@@ -36,6 +36,42 @@
...
+            placeholder="220"
+            dense
+            outlined
+            type="number"
+            :min="0"
+            :step="1"
+            hide-details
+          />
+        </app-setting>
+        <template v-if="lane?.bed_temp != null">
```
pedrolamas: Nit: this uses `lane?.bed_temp != null` while the send logic keys off `bedTemp !== undefined` / `currentBedTemp`. Same result today, but two sources of truth for the same condition — `v-if="currentBedTemp !== undefined"` keeps them from drifting.

pedrolamas (commented):
Nice feature — the handoff through the existing `spoolSelectionOnly` / `selectedSpoolId` dialog state is the right pattern, and gating the temp fields on `'AFC_SET_SPOOL_TEMP' in availableCommands` matches how `AfcCardButtons.vue` does it. Two things need fixing before merge, plus one nit — all inline.

src/components/widgets/afc/dialogs/AfcUnitLaneFilamentDialog.vue
pedrolamas: Personal note: I'm always inclined for the simple `!= null`

src/components/widgets/afc/AfcCardUnitLaneBody.vue
pedrolamas: Personal note: if open the browser console, you will probably see a warning from Vuex from this code!

jimmyjon711: Fixed, changed to a proper dispatch.

src/components/widgets/afc/dialogs/AfcUnitLaneFilamentDialog.vue
jimmyjon711: Fixed, also added a weight check to make sure value was not less that zero as well

src/components/widgets/afc/dialogs/AfcUnitLaneFilamentDialog.vue
jimmyjon711: Fixed

src/components/widgets/afc/dialogs/AfcUnitLaneFilamentDialog.vue
```diff
@@ -36,6 +36,42 @@
...
+            :min="0"
+            :step="1"
+            hide-details
+          />
+        </app-setting>
+        <template v-if="hasBedTemp">
+          <v-divider class="my-3" />
+          <app-setting
+            :title="$t('app.afc.BedTemp')"
+            :sub-title="$t('app.afc.BedTempSubtitle')"
```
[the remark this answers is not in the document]
pedrolamas: Personal note: this is actually expected, right?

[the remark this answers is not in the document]
jimmyjon711: @pedrolamas Correct, since older AFC installs might not have the bed temp value, but will have the extruder temp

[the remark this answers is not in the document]
jimmyjon711: I updated the PR description to call out `if bed temp variable exists`

src/components/widgets/afc/dialogs/AfcUnitLaneFilamentDialog.vue
```diff
@@ -99,19 +178,36 @@ export default class AfcUnitLaneFilamentDialog extends Mixins(StateMixin, AfcMix
...
+      )
+    ) {
+      // Only include BED_TEMP parameter if the lane reports a bed_temp
+      const bedTempParam = this.bedTemp !== null ? ` BED_TEMP=${this.bedTemp}` : ''
+      gcode.push(`AFC_SET_SPOOL_TEMP LANE=${encodeGcodeParamValue(this.name)}${bedTempParam} EXTRUDER_TEMP=${this.extruderTemp}`)
+    }
+
     this.sendGcode(gcode.join('\n'))
 
     this.show = false
```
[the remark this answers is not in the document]
jimmyjon711: fixed

pedrolamas (approved):

PR #1915, opened 2026-08-04
feat: G-code template highlighting in Klipper config

pedrolamas:
Values of Klipper config options named `gcode` or ending in `_gcode` (`activate_gcode`, `pre_unload_gcode`, …) are G-code templates, but the editor rendered them as flat strings. They are the most edited and most error-prone part of a `printer.cfg`.

This extracts the G-code grammar from the standalone `gcode` language into a shared factory and embeds it in `klipper-config` for those keys. The match is a suffix, not a substring: `gcode_id`, `gcode_x_offset` and `gcode_load_sequence` merely start with `gcode` and stay plain values.

It also recognises Klipper's Jinja2 layer, without which the highlighting would be unusable — Klipper builds its environment with non-standard delimiters, so expressions are single-brace `{ }` rather than `{{ }}`:

```ini
[gcode_macro MY_MACRO]
gcode:
  {% set SPEED = params.SPEED | default(3000) | int %}
  G1 X{printer.toolhead.axis_maximum.x} F{SPEED}  ; move
```

Keyword matching is case-sensitive and names after `.` are attributes, so `{% set MACRO = params.MACRO %}` and `foo.call` read as variables rather than as the `macro` and `call` keywords. Values keep the configparser rule that a whitespace-preceded `#`/`;` is stripped from the raw line before Klipper or Jinja ever sees it.

## Notes

- `gcode.monarch.spec.ts` is unchanged and passes as-is — it is the regression gate for the extraction.
- The continuation state machine (`checkValue`) is now shared between plain and G-code values rather than duplicated.
- Theme rules are added for the new Jinja scopes in both light and dark; the G-code scopes inherit the existing rules by prefix.

## Testing

`pnpm run test`, `type-check`, `lint`, `circular-check` and `build` all pass. The tokenizer spec covers key matching, the negative `gcode_*` cases, multi-line continuations, comment stripping, and the Jinja regions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/monaco/language/__tests__/klipper-config.monarch.spec.ts
+++ b/src/monaco/language/__tests__/klipper-config.monarch.spec.ts
@@ -213,9 +213,11 @@ describe('klipper-config Monarch tokenizer', () => {
   })
 
   describe('multi-line continuations', () => {
+    // `description` is deliberately not a G-code key, so this block pins the
+    // generic continuation machinery rather than the G-code path below.
     it('continues into more-indented lines (zero-indent key)', () => {
-      expect(tokenizeInSection('gcode:\n  G28\n  G1 X0')).toEqual([
-        t(['keyword', 'gcode'], ['separator', ':']),
+      expect(tokenizeInSection('description:\n  G28\n  G1 X0')).toEqual([
+        t(['keyword', 'description'], ['separator', ':']),
         t(['white', '  '], ['string', 'G28']),
         t(['white', '  '], ['string', 'G1 X0'])
       ])
@@ -229,8 +231,8 @@ describe('klipper-config Monarch tokenizer', () => {
     })
 
     it('returns to root when next line drops below key indent', () => {
-      expect(tokenizeInSection('gcode:\n  G28\nother_key = 1')).toEqual([
-        t(['keyword', 'gcode'], ['separator', ':']),
+      expect(tokenizeInSection('description:\n  G28\nother_key = 1')).toEqual([
+        t(['keyword', 'description'], ['separator', ':']),
         t(['white', '  '], ['string', 'G28']),
         t(['keyword', 'other_key'], ['white', ' '], ['separator', '='], ['white', ' '], ['string', '1'])
       ])
@@ -244,40 +246,40 @@ describe('klipper-config Monarch tokenizer', () => {
     })
 
     it('skips blank lines while waiting for the continuation', () => {
-      expect(tokenizeInSection('gcode:\n\n  G28')).toEqual([
-        t(['keyword', 'gcode'], ['separator', ':']),
+      expect(tokenizeInSection('description:\n\n  G28')).toEqual([
+        t(['keyword', 'description'], ['separator', ':']),
         t(),
         t(['white', '  '], ['string', 'G28'])
       ])
     })
 
     it('skips full-line comments while waiting for the continuation', () => {
-      expect(tokenizeInSection('gcode:\n# blank-ish\n  G28')).toEqual([
-        t(['keyword', 'gcode'], ['separator', ':']),
+      expect(tokenizeInSection('description:\n# blank-ish\n  G28')).toEqual([
+        t(['keyword', 'description'], ['separator', ':']),
         t(['comment', '# blank-ish']),
         t(['white', '  '], ['string', 'G28'])
       ])
     })
 
     it('also skips ; comment lines while waiting', () => {
-      expect(tokenizeInSection('gcode:\n; mid\n  G28')).toEqual([
-        t(['keyword', 'gcode'], ['separator', ':']),
+      expect(tokenizeInSection('description:\n; mid\n  G28')).toEqual([
+        t(['keyword', 'description'], ['separator', ':']),
         t(['comment', '; mid']),
         t(['white', '  '], ['string', 'G28'])
       ])
     })
 
     it('skips indented comment lines while waiting', () => {
-      expect(tokenizeInSection('gcode:\n  # indented\n  G28')).toEqual([
-        t(['keyword', 'gcode'], ['separator', ':']),
+      expect(tokenizeInSection('description:\n  # indented\n  G28')).toEqual([
+        t(['keyword', 'description'], ['separator', ':']),
         t(['white', '  '], ['comment', '# indented']),
         t(['white', '  '], ['string', 'G28'])
       ])
     })
 
     it('skips whitespace-only lines while waiting', () => {
-      expect(tokenizeInSection('gcode:\n   \n  G28')).toEqual([
-        t(['keyword', 'gcode'], ['separator', ':']),
+      expect(tokenizeInSection('description:\n   \n  G28')).toEqual([
+        t(['keyword', 'description'], ['separator', ':']),
         t(['white', '   ']),
         t(['white', '  '], ['string', 'G28'])
       ])
@@ -287,8 +289,8 @@ describe('klipper-config Monarch tokenizer', () => {
     // `#`) preceded by whitespace inside a continuation body is stripped
     // the same way as on the original key=value line.
     it('strips inline comments inside a continuation body', () => {
-      expect(tokenizeInSection('gcode:\n  G28 ; mid-comment')).toEqual([
-        t(['keyword', 'gcode'], ['separator', ':']),
+      expect(tokenizeInSection('description:\n  G28 ; mid-comment')).toEqual([
+        t(['keyword', 'description'], ['separator', ':']),
         t(['white', '  '], ['string', 'G28'], ['white', ' '], ['comment', '; mid-comment'])
       ])
     })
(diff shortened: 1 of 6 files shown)


pedrolamas (commented):
🤖 AI Code Review — 3 finding(s) across 2 file(s)

Clean extraction of the G-code grammar into a shared factory, with careful handling of Klipper’s non-standard Jinja delimiters, case-sensitive keyword matching via `$0==`, and a shared `continuationCheck` for plain vs G-code values. Test coverage for the happy paths is strong.

Two gaps in the embedded path are worth fixing before merge: configparser comment semantics are undermined by the inherited standalone `;` rule, and Jinja inside macro named-arg values is not entered.

src/monaco/language/gcode-rules.ts
```diff
@@ -0,0 +1,384 @@
...
+
+    [
+      /^(\s*)(n\d+)/,
+      ['white', eosSwitch('keyword.command.n')]
+    ],
+
+    [
+      /;.*$/,
+      eosSwitch('comment')
+    ],
```
pedrolamas, line 163: 🤖 **Major**: Embedded mode still treats any `;` as a G-code comment.

The prologue correctly implements configparser’s whitespace-before-`#`/`;` rule, and the PR text says that rule wins over G-code/Jinja tokenizing. But this standalone rule remains in `entryRules` for both modes:

```ts
[/;.*$/, eosSwitch('comment')]
```

So in a `gcode:` value, `G28;note` and `M117 Hello;World` highlight `;…` as a comment even though configparser would keep it in the value (letter before `;`). Plain values already pin the opposite (`value;not-a-comment`).

**Suggested fix:** omit `/;.*$/` when `embedded` (rely on `prologue` only), and add a regression case next to the existing whitespace-comment tests.

src/monaco/language/gcode-rules.ts
```diff
@@ -0,0 +1,384 @@
...
+        cases: {
+          '@eos': { token: 'invalid', next: abandon },
+          '@default': { token: 'string.quote', bracket: '@open', next: push(stringState) }
+        }
+      }
+    ],
+
+    [
+      /\S+/,
+      { token: 'string', next: '@pop' }
```
pedrolamas: 🤖 **Major**: Jinja in macro named-arg values is never entered.

`argValue` consumes the whole value with `/\S+/` as an opaque `string`. Common Klipper patterns like `SET_GCODE_OFFSET Z={params.Z|float}` or `SET_VELOCITY_LIMIT VELOCITY={SPEED}` therefore get no Jinja highlighting, while the same `{…}` works after a G-code param letter (`G1 X{x}`) via the dedicated entry rule.

**Suggested fix:** in embedded mode, before `/\S+/`, add the same Jinja openers used in `entryRules` (`{%`, `{#`, `{`), and/or teach `argValue` to stop at `{` and rematch. Worth a tokenizer test for `MACRO ARG={expr}`.

src/monaco/language/__tests__/klipper-config.monarch.spec.ts
```diff
@@ -364,14 +363,220 @@ describe('klipper-config Monarch tokenizer', () => {
...
+            ['keyword.control.jinja', 'set'], ['white', ' '],
+            ['variable.jinja', 'x'], ['white', ' '],
+            ['operator', '='], ['white', ' '],
+            ['number.jinja', '1'], ['white', ' '],
+            ['delimiter.jinja', '%}'], ['white', ' '],
+            ['comment', '; note']
+          )
+        ]
+      ]
+    ])('strips a whitespace-preceded comment inside a G-code value: %j', (input, expected) => {
```
pedrolamas, line 469: 🤖 **Minor**: Comment-stripping cases only cover the whitespace-preceded path.

These tests pin `G28 ; note` and `{% … %} ; note`, which is good. Adding the negative twin (`G28;note` / `M117 Hello;World` must **not** become `comment` once the embedded `/;.*$/` rule is removed) would lock the configparser distinction the rest of this language is built around.

src/monaco/language/gcode-rules.ts
pedrolamas: Not applying this one — I think the premise is off.

There are two parsers in sequence, not one. configparser hands the raw template string to Klipper, and Klipper's G-code dispatcher then strips from the first `;` **unconditionally** (`klippy/gcode.py`):

```python
line = origline = line.strip()
cpos = line.find(';')
if cpos >= 0:
    line = line[:cpos]
```

There's no whitespace condition there. So in a `gcode:` value `G28;note` really does execute as `G28`, and highlighting `;note` as a comment is correct. The divergence from `value;not-a-comment` in a plain value is deliberate: a plain value is only ever seen by configparser.

You did surface a real bug next to it, though. M117/M118 are the exception — they read the raw command line via `get_raw_command_parameters()` (`klippy/extras/display_status.py`, `klippy/extras/respond.py`), which strips a line number and a trailing checksum but never a `;`. The old `([^;]*)` payload cut the message at `;`, so `M117 Hello;World` displayed wrongly. Fixed in db79d970, along with the two `gcode.monarch.spec.ts` assertions that pinned the old behaviour.

Embedded, the payload still stops where configparser would have cut a whitespace-preceded `#`/`;` — confirmed against CPython with Klipper's exact parser settings:

```python
>>> cfg = configparser.RawConfigParser(strict=False, inline_comment_prefixes=(';', '#'))
# "M117 test ;test"  ->  "M117 test"
# "M117 a;b"         ->  "M117 a;b"
```

Both behaviours are now pinned in the specs.

src/monaco/language/gcode-rules.ts
pedrolamas: Confirmed and fixed in db79d970.

Adding the openers alone wasn't enough: `jinjaCloseEos` returned to the *entry* state, so `Z={z} MOVE=1` would have closed into `gcodeValue` and read `MOVE` as `keyword.macro` instead of `keyword.param`. Jinja regions now carry a return tag and close back into the params state.

One trap worth recording — my first attempt stored the return *state name* in the state argument, which blew up at runtime:

```
klipper-config: trying to switch to a state 'gcodevalue.' that is undefined
```

`$Sn` substitution runs through Monarch's `fixCase`, so `ignoreCase: true` lower-cases the substituted name into a state that doesn't exist. It now stores a lower-case `value`/`params` tag and branches on `$S4==params`.

Known limit: only a value *starting* with `{` enters Jinja, so `Z=abc{x}` stays an opaque string. Handling that needs `argValue` to stop at `{` and rematch — happy to do it if you think it's worth it, but it looks rare next to `ARG={expr}`.

src/monaco/language/__tests__/klipper-config.monarch.spec.ts
pedrolamas: Added in db79d970, though inverted from the suggestion — since the embedded `/;.*$/` rule is correct (see the thread above), the useful test pins that behaviour rather than the opposite:

```ts
'gcode:\n  G28;note'  →  keyword.command.g + comment
```

The `it.each` also covers the M117/M118 exception in both directions: `M117 Hello;World ;tail` keeps the inline `;` in the payload but loses the configparser-stripped ` ;tail`, plus the `#` variant and M118. On the standalone side `gcode.monarch.spec.ts` gained `M117 test ;test` / `M118 test ;test` / `M117 a;b`, which pin the payload running to end of line — the case that actually distinguishes the two contexts.

src/monaco/language/gcode-rules.ts
```diff
@@ -0,0 +1,410 @@
...
+    [argValueState]: argValueRules,
+    [stringState]: stringRules
+  }
+
+  if (embedded) {
+    // A matched closer resumes G-code on the rest of the line, if any.
+    const jinjaCloseEos = (token: string): Action => ({
+      cases: {
+        '@eos': { token, switchTo: `@${checkState}.$S2.none` },
+        '$S4==params': { token, switchTo: `@${paramsState}.$S2` },
```
pedrolamas: 🤖 **Minor**: Closing Jinja always returns to `params`, so compound arg values break.

For the common case `Z={params.Z|float} MOVE=1` this is correct — the expression *is* the whole arg, then `MOVE=` must be seen from `params`.

But after `{% … %}` (or any Jinja region that is only a *prefix* of the arg), switching to `params` means the remainder is not scanned as arg text. Example:

```ini
SET_X V={% if 1 %}1{% endif %} MOVE=1
```

tokenizes the first `{% if 1 %}`, returns to `params`, then falls through to entry’s `/.*$/` and paints `1{% endif %} MOVE=1` as one opaque `string` — sibling params are swallowed. `Z={x}y MOVE=1` fails the same way.

Not blocking if statement-style arg values are considered out of scope; if you want them, close back into `argValue` and teach that state to pop to `params` on whitespace (and to stop `/\S+/` at `{`) so both `{expr} NEXT=` and `{% if %}…{% endif %}` work.

pedrolamas (commented):
🤖 AI Code Review — 1 finding(s) across 1 file(s)

Prior feedback is addressed: M117/M118 now follow Klipper’s raw-line semantics (with configparser cutting whitespace-`#`/`;` in the embedded path), and Jinja in macro named-args returns to `params` so sibling `NAME=value` pairs keep working. The `$S4` return-tag approach is a solid fix for the `fixCase`/`ignoreCase` trap.

One residual edge case below around statement-style / compound Jinja arg values.

src/monaco/language/gcode-rules.ts
pedrolamas: Good catch, and it turned out to be the visible corner of something worse — fixed in 0f657d9f.

Chasing the compound-arg case showed that *any* line ending inside a macro call left the tokenizer parked in the params/argValue/string sub-machine, so the next line never reached the continuation check:

```ini
gcode:
  MY_MACRO
[next_section]        # → string="[next_section]" — header eaten, rest of file mis-parsed
```
```ini
gcode:
  SET_PIN PIN=x
other = 1             # → keyword.macro="other" — sibling key eaten
```

`G28` was fine, because the entry state implements the hand-back and the three pushed sub-states never did — standalone `.gcode` has no continuation check to satisfy and legitimately resumes mid-macro on the next line. A macro whose last line is a macro call (`SET_PIN …`, `RESTORE_GCODE_STATE NAME=…`) is common enough that this was easy to hit.

Embedded now hands back from every state, and the sub-machine moves with `switchTo` rather than push/pop — otherwise handing back from a pushed state strands a frame per line, giving O(n) stack depth over a long macro. Standalone keeps push/pop and is untouched.

Your case falls out of the same fix, closing back into the value rather than into `params`:

```
SET_X V={% if 1 %}1{% endif %} MOVE=1
      → {%·if·1·%} + string "1" + {%·endif·%} + keyword.param MOVE ✓
SET_X Z={x}y MOVE=1
      → {·x·} + string "y" + keyword.param MOVE ✓
```

Regression tests cover both shapes plus the three end-of-line cases, including unterminated-quote recovery. 355 tests green.

Issue #1921, opened 2026-08-09
History page paginates over the loaded 50 jobs, so the paginator looks complete when it is not


arlophoenix:
# Body

`JobHistory.vue` renders `<v-data-table :items="history" :items-per-page="15">`, where `history` is the `history/getHistory` getter — i.e. whatever `history/init` loaded, which is `Globals.JOB_HISTORY_LOAD` (50) jobs.

On a printer with 184 jobs, that presents a paginator reading "1-15 of 50" across four pages. Reaching the last page looks like reaching the end of the history. There is nothing on the page to indicate that 134 jobs were never fetched; the Load All button is the only way to see them, and nothing prompts you to press it.

Same root cause as the file browser rendering a blank Status for a file whose last print is outside the loaded page (#1922), but arguably a worse symptom: the file browser shows an empty cell, which at least looks like missing information, whereas the paginator actively asserts a total that is wrong.

Reproduced on a stock Creality K1C with 184 history jobs, no local patches.

Issue #1920, opened 2026-08-09
history store's state.count is rows-returned, not the total


arlophoenix:
# Body

`src/store/history/mutations.ts` sets `state.count` from `payload.count` on every `server.history.list` response. Moonraker's `_handle_jobs_list` ends with `return {"count": len(jobs), "jobs": jobs}` — `count` is the length of the page it just returned, not the number of rows in the table. After `history/init` runs with `limit: Globals.JOB_HISTORY_LOAD`, the store holds `count === 50` and `jobs.length === 50` on any printer with 50 or more jobs.

That makes `count` unusable as an "is everything loaded" signal, which is the obvious thing to reach for. Any guard of the shape `if (jobs.length >= count) { /* everything is loaded */ }` is permanently true and would silently disable whatever it guards. This was very nearly written into a fix before the semantics were checked.

`setDeleteJob` also splices `state.jobs` without touching `count`, so it drifts further as jobs are deleted.

Nothing reads `state.count` today, so there is no user-visible bug — this is about the trap it sets. Worth either fixing the semantics, or removing the field, or documenting what it actually holds.

For the record, there is no reliable total available either: `job_totals.total_jobs` is a lifetime counter that survives deletes and is zeroed by `server.history.reset_totals`.

Issue #998, opened 2022-12-31
Allow the number of history items to be configurable


simpat1zq:
### Is your feature request related to a problem? Please describe

History is currently hardcoded at 50 entries. For me, and I'm sure many other people, that only goes back about 2 weeks. I read an older issue that said that it may cause performance issues with higher numbers, but I'm not sure if that problem is still relevant. Either way, I would be nice if we could set that limit ourselves. 

### Describe the solution you'd like

Allow the history limit to be user configurable. 

### Describe alternatives you've considered

Mainsail is able to do it, so I've installed that on my printers as well, and I can go there for history, but I would prefer to just stay in Fluidd.

PR #1922, opened 2026-08-09
fix(history): resolve jobs outside the loaded page

arlophoenix:
Refs #998

The G-code file browser shows a `Status` and `Total Duration` per file, both resolved from the history job the file's metadata names. Only the newest `JOB_HISTORY_LOAD` (50) jobs are ever loaded, so a file whose last print is older than that page renders blank — indistinguishable from a file that has never been printed. Opening History → Load All fixes it until the next reconnect, which re-runs `history/init` and drops back to 50.

Raising the constant only moves the cliff. The browser needs the job behind each file's *last* print, and those are spread across the whole history rather than clustered at the recent end, which is the one thing a "newest N" page guarantees. On a printer with 127 jobs and 75 linked files, a page of 50 renders 33 and blanks 42; a page of 100 renders 73 and blanks 2; only "all" renders every one — and "all" is exactly what a default must not be.

So instead of widening the page, this resolves the misses individually. A directory load collects the job ids its files name, drops the ones already loaded, already resolved, already known missing, or already in flight, and fetches the rest through `server.history.get_job` — two requests in flight at a time, because Moonraker serializes every history endpoint behind one lock, so more concurrency buys no wall-clock and a wide pool would hold that lock against the History page. Results are committed once per drain and kept beside the loaded page rather than merged into it, so History keeps showing a contiguous page. The same trigger runs for a single file's metadata, because `files/getFile` is decorated the same way and has eight consumers.

Deleting a job leaves its `job_id` on the file — Moonraker's `delete_job` only touches `job_history` — so orphaned ids are easy to produce, and Fluidd's own Remove All produces them for every file at once. Those come back 404. Each one is recorded so it is asked for exactly once, and the 404 is suppressed rather than toasted, via a new per-request `suppressError` on `NotifyOptions`. It is opt-in per request: every other 4xx in the app keeps its toast, which is what the first test below guards. The negative cache is dropped on reconnect, since a 404 recorded against a previous connection is not evidence about this one, and cleared for an id when a job is added under it — Moonraker's job ids are a SQLite `INTEGER PRIMARY KEY` with no `AUTOINCREMENT`, so they are reused.

Two measurements, one printer each, both read from Moonraker's own API rather than off a rendered page. A stock Creality K1C with no local patches: 184 history jobs, 112 G-code files, 105 carrying a `job_id`, of which 64 fall outside the 50-row page. All 64 were probed with `server.history.get_job` and all 64 resolve — zero orphans — so every one is a completed print still in the database that the browser renders as blank. That is 61% of the files that have ever printed, or 57% of all G-code files. The 127-job figures above come from a second printer.

An alternative was considered and rejected on measurement: one `server.history.list` scoped by `since: min(print_start_time)` over the misses, which would have been a single round trip and no 404 path at all. Against the stock printer that query returns 183 of 184 rows — 99% of the history — because the oldest miss sits at the very start of it. It collapses into `limit: 0`.

While in `getHistoryById`, it now reads from an index rather than a linear scan. It is a method-style getter called once per file by `files/getDirectory`, which made decorating a directory O(files × jobs).

## Test plan

- [x] `pnpm run lint` clean
- [x] `pnpm run type-check` clean
- [x] `pnpm run test:unit` — 447 tests pass, including 47 new ones covering the cache, its invalidation, the bounded drain, the reconnect repair, and `files/getDirectory` returning a populated `history` for a file outside the loaded page
- [x] `pnpm run circular-check` clean
- [ ] Manual test pass (see checklist below) — **not performed**; no printer was available to the session that opened this PR

## Manual test pass

- [ ] On a printer with more than 50 history jobs, open Jobs. A file whose last print is older than the 50th-newest job shows a real Status (e.g. `Completed`) and a real Total Duration (e.g. `1h 12m`) without visiting History → Load All.
- [ ] History → Remove All, then open Jobs. Zero red toasts; every Status and Total Duration is blank.
- [ ] Delete one job in History whose file is visible in Jobs. That file's Status and Total Duration go blank immediately.
- [ ] Trigger any other Moonraker 4xx (e.g. delete a file that no longer exists). Its usual red toast still appears.
- [ ] Restart Moonraker while on Jobs. After the socket reconnects, the columns are still populated.

## Not addressed

- `getHistoryByFilename` has no consumers and carries the same window defect. Left alone to keep this single-concern.
- The History page's own paginator shows 15 per page over the loaded 50, so a 184-job history presents a four-page paginator that looks complete. Same root cause, worse symptom — filed separately as #1921.
- `state.count` is the number of rows returned, not the number of rows that exist, and is never decremented on delete. Nothing reads it today, and no "is everything loaded" signal can be built from it. Filed separately as #1920.
- Porting the history store to Pinia to match #1918. Not a contributor's call to make.

### DIFF
--- a/src/api/socketActions.ts
+++ b/src/api/socketActions.ts
@@ -587,6 +587,17 @@ export const SocketActions = {
     )
   },
 
+  serverHistoryGetJob (uid: string, options?: NotifyOptions) {
+    return baseEmit<Moonraker.History.JobResponse>(
+      'server.history.get_job', {
+        ...options,
+        params: {
+          uid
+        }
+      }
+    )
+  },
+
   serverHistoryList (params?: { start?: number; limit?: number; before?: number; since?: number; order?: string }, options?: NotifyOptions) {
     return baseEmit<Moonraker.History.ListResponse>(
       'server.history.list', {
--- a/src/plugins/__tests__/socketClient.spec.ts
+++ b/src/plugins/__tests__/socketClient.spec.ts
@@ -1,0 +1,127 @@
+import { TypedStore } from '@/store'
+import { WebSocketClient } from '@/plugins/socketClient'
+
+const sockets: FakeWebSocket[] = []
+
+class FakeWebSocket {
+  static readonly OPEN = 1
+
+  readyState = FakeWebSocket.OPEN
+  sent: string[] = []
+
+  onopen: (() => void) | null = null
+  onclose: ((event: unknown) => void) | null = null
+  onerror: ((event: unknown) => void) | null = null
+  onmessage: ((message: { data: string }) => void) | null = null
+
+  constructor () {
+    sockets.push(this)
+  }
+
+  send (data: string) {
+    this.sent.push(data)
+  }
+
+  close () {}
+}
+
+const socket = () => sockets[sockets.length - 1]
+
+const onSocketError = vi.fn()
+const onSetStatus = vi.fn()
+
+const createSocketStore = () => new TypedStore({
+  modules: {
+    socket: {
+      namespaced: true,
+      state: () => ({ status: 'ready' }),
+      actions: {
+        onSetStatus: (...args: unknown[]) => onSetStatus(...args),
+        onSocketError: (...args: unknown[]) => onSocketError(...args)
+      }
+    },
+    config: {
+      namespaced: true,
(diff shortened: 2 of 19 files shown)


pedrolamas:
Replaced by https://github.com/fluidd-core/fluidd/pull/1942

PR #1939, opened 2026-08-19
fix: use debian base image for dev container

### LINKED ISSUE #1938: Devcontainer compatibility issues with alpine & typescript native bridge
### Fluidd Version

dev container as of 94ea4ea

### Browser

Other (Please tell us below)

### Device

Laptop

### Operating System

macOS

### What happened

I work in the devcontainer on macOS (Apple Silicon), the devcontainer builds the Alpine base image for linux/arm64.

Since the introduction of TypeScript Native Bridge (TNB), running `pnpm run dev` on the dev container fails with a seg fault on this platform.

The project's own documentation confirms this is a known, unsupported combination rather than a bug specific to this repo:

>Linux packages target glibc 2.31 and are rejected by the release gate if they acquire a newer symbol requirement. Alpine/musl is not supported: Go's -buildmode=c-shared runtime crashes at load on musl libc — even a trivial hello-world c-shared library segfaults, on Go 1.22 through 1.26 alike (golang/go#13492, a 10-year-open upstream issue with an active fix in golang/go#75048; tsgo's own CLI works on Alpine only because it ships CGO-free static binaries, and a NAPI bridge cannot be CGO-free). Workaround: run the typecheck/lint step in a glibc-based image (node:24 or node:24-bookworm-slim) and deploy into the Alpine stage of your multi-stage build; apk add gcompat does not help.

I've confirmed the issue is not present with commit ad13d2e, the commit just prior to the introduction of TNB.

### Current local workaround:
I've been manually overriding .devcontainer/Dockerfile to use node:24-slim (Debian/glibc) instead of node:24-alpine, which resolves the failure.

### Suggested Fix
Switch the default devcontainer base image from node:24-alpine to a glibc-based image (node:24-slim or node:24-bookworm-slim) on arm64 hosts. This is the option seems to be working locally, and matches the workaround the TNB project itself documents for glibc-dependent steps.

`pnpm run dev` should work in the default devcontainer on Apple Silicon (arm64) without requiring a manual override of .devcontainer/Dockerfile.

### How to reproduce

Steps to Reproduce

- On a macOS host with Apple Silicon (arm64), open the project in a devcontainer using the default .devcontainer/Dockerfile (node:24-alpine base image).
- Let the container build — it targets linux/arm64 by default on this host.
- Accept corepack installation.
- Run `pnpm run dev.`
- The process fails immediately with:
   `Error: Error loading shared library libresolv.so.2: No such file or directory (needed by .../@typescript-native-bridge/linux-arm64/native/bridge.node)`
-  Inside the container, install the suggested compatibility shim:
-    `apk add gcompat`
- Run `pnpm run dev` again.
- The libresolv.so.2 error is gone, but the process now segfaults instead.

### Additional information

_No response_

pedrolamas:
Changing devcontainer base image from Alpine to Debian as typeScript-native-bridge does not work with Alpine/musl.

Fixes #1938

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,17 +1,20 @@
-FROM node:24-alpine
+# NOTE: must stay on a glibc base. typescript-native-bridge (the `typescript`
+# catalog entry) ships a Go c-shared NAPI bridge with no musl build — on Alpine
+# it fails to load, and segfaults even with `gcompat`. See issue #1938.
+FROM node:24-trixie-slim
 
 RUN \
-  apk update \
-  && apk add \
+  apt-get update \
+  && apt-get install -y --no-install-recommends \
+    ca-certificates \
     git \
-    git-zsh-completion \
+    wget \
     zsh \
-    zsh-completions \
-  && sed -i '/^node:/s|/bin/ash|/bin/zsh|' /etc/passwd \
+  && usermod --shell /bin/zsh node \
   && corepack enable \
   && mkdir -p /workspaces/fluidd/node_modules /workspaces/fluidd/.pnpm-store \
   && chown -R node:node /workspaces/fluidd \
-  && rm -f /etc/apk/cache/* /root/.cache
+  && rm -rf /var/lib/apt/lists/*
 
 USER node
 WORKDIR /workspaces/fluidd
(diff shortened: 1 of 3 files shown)


pedrolamas:
Hi @tworkman08, can you check if this works for you please?

tworkman08:
> Hi @tworkman08, can you check if this works for you please?

Looks good, Thank you!

Screenshot 2026-08-19 at 11 59 45 AM

PR #1958, opened 2026-09-19
fix: keep the axis pointer on the mouse

KapJI:
Each data update calls `setOption`, which clears the highlight the axis pointer put on each series. The symbols at the pointer vanish at the next update, within a second, and stay gone until the pointer moves to another sample. The axis pointer still records those samples as highlighted, so it never restores them itself. Its line also stays on the sample it last snapped to, so with the mouse still it scrolls away with the data.

After each data update the chart now resets the axis pointer (`updateAxisPointer` with `currTrigger: 'leave'`) and shows it again at the mouse position (`showTip`), so the symbols come back and the line snaps back to the sample nearest the mouse.

## Before (symbols disappear, axis drifts):

https://github.com/user-attachments/assets/0003b91d-84fe-4baa-a84c-0f7b5f810520

## After:

https://github.com/user-attachments/assets/7563b83e-fd26-40a5-b4fa-dddcd95b85b9


KapJI:
Hi @pedrolamas, rebased on the latest `develop`. Does this approach look ok to you? #1959 sits on top of it.

KapJI:
Btw, I found this reported upstream as apache/echarts#20588, open since Dec 2024. The cause is the FIXME in `dispatchHighDownActually` (`axisTrigger.js`): it keeps its own record of highlighted samples that `setOption` never clears, and `updateAxisPointer` with `currTrigger: 'leave'` is the cleanest public way to reset it.

PR #1959, opened 2026-09-19
fix: highlight only the line under the pointer

KapJI:
Depends on #1958, which is the first commit here; the second is this change.

Since #1858, hovering the temperature chart makes it flash. Whenever the pointer lands on a line, a symbol or an area, `emphasis.focus: 'series'` dims every other series, and on the next move the axis pointer highlights all of them again, so the chart keeps switching between one line and all of them. Between lines nothing is focused, so each line needs careful aiming.

The series are now `silent`, which alone stops the flashing. Hovering a sensor in the Thermals card still highlights its series as before.

On top of that, I propose widening the visible line nearest to the pointer (within 16 px, interpolated between samples, keeping its dash pattern) as a nicer highlight. It is a `setOption`, so it re-syncs the axis pointer as #1958 does after data updates. If you would rather not have it, I can remove it, leaving only the tooltip on chart hover.

Rows hover works as before.

## Before:

https://github.com/user-attachments/assets/33824353-0f6e-4f61-89c2-9d6c53a48015

## After:

https://github.com/user-attachments/assets/368eac7d-e58f-4de9-b0e8-479112dcc30f


KapJI:
Rebased and decreased `HOVER_WIDTH_INCREASE` from 1.5 to 1, I think it looks nicer this way.

